PR #10, opened 2015-05-12
Resolution data added to frame output

gdhungana:
Resolution data on same sampling as spectra is added. For I/O, frame file is created using desispec io routine. For sky and calibration, parameters need to be defined to be able to write using desispec io. For example, for calibration, argument for example is standard star model flux. Not sure if we want to do so for quickGen.


weaverba137:
This pull request contains a spurious file called "1".  It appears to be a leftover from a bad git log message.

gdhungana:
It seems so. That needs to be deleted. Should I delete that file and send another pull request?

weaverba137:
Delete it on the branch you are trying to merge, & this PR should track it.

PR #13, opened 2015-09-02
defines imports for general utilities: config file + positioner shape

forero:
Here I am adding a submodule "util". I use the functions there to describe positioner shape (to be used for fiber collision checks) and define a simple configuration file parser (to setup the parameters for a simulated survey).

The new script in bin/ mini-desi.py tests that "from desisim import util" actually works.

### DIFF
--- a/bin/mini-desi.py
+++ b/bin/mini-desi.py
@@ -1,0 +1,12 @@
+#!/usr/bin/env python
+"""
+vanilla survey strategy
+"""
+
+import sys
+import os
+import numpy as np
+#sys.path.insert(0, '/home/forero/desihub/desisim/py/')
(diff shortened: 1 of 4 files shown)


sbailey:
TargetTile, TargetSurvey, and the config parser are appropriate for desisim.

The other pieces seem to describe real positioner and focal plane hardware geometry.  Perhaps they would fit better in a python-based fiberassign code (pyfiber?).

It is fine for desisim to depend upon other products but not vice-a-versa: real data/hardware products like desispec and fiberassign should *not* depend upon desisim.  Since real fiber assignment or other code might want to access these positioner collision, geometry, etc. routines it seems they should not live in desisim.

forero:
OK. I will move the hardware geometry to a different repo. Pyfiber sounds good!

PR #14, opened 2015-09-02
Cleanup

sbailey:
This pull request implements several cleanup items:

* separate the creation of a new exposure from the pixel-level simulation of putting those photons onto a CCD.  Previously pixsim-desi did both, creating confusion since you had to run "pixsim-desi" as the starting point for later non-pixsim operations.
* fix header keyword bugs in propagating FLAVOR and EXPTIME
* fix wavelength bug in simspec files for arc exposures with a non-uniform wavelength grid.  Now the wavelength is kept as dedicated arrays in separate HDUs instead of just CRVAL1 and CDELT1 (which can't handle non-uniform grids)

There is more cleanup that could be done (e.g. desisim doesn't follow the guidelines of clearly separating I/O from algorithms), but these 3 changes are a reasonable set to merge in now before doing other cleanup.

### DIFF
--- a/bin/newexp-desi
+++ b/bin/newexp-desi
@@ -1,0 +1,90 @@
+#!/usr/bin/env python
+
+"""
+Generate a new DESI exposure
+
+newexp-desi arc/flat/science [options]
+
+Stephen Bailey, LBL
+Fall 2014
+"""
+
+import sys
+import os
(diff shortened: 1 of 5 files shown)


profxj:
Inspected and approved.  Merging.

PR #17, opened 2015-09-07
Format flow

forero:
Updated targets.py with the classes used for bookkeeping during fiber assignment.
Added script quicksurvey to test new classes and functions. Also to learn how to reuse code already written in desisim/desimodel/desispec.

### DIFF
--- a/bin/quicksurvey
+++ b/bin/quicksurvey
@@ -1,0 +1,16 @@
+#!/usr/bin/env python
+"""
+vanilla survey strategy
+"""
+
+import sys
+import os
+import numpy as np
+
+sys.path.insert(0, '/home/forero/desihub/desisim/py/')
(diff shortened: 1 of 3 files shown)


bin/quicksurvey
```diff
@@ -0,0 +1,27 @@
...
+"""
+
+import sys
+import os
+import numpy as np
+
+sys.path.insert(0, '/home/forero/desihub/desisim/py/')
+sys.path.insert(0, '/home/forero/desihub/desimodel/py/')
+sys.path.insert(0, '/home/forero/desihub/desispec/py/')
+sys.path.insert(0, '/home/forero/desihub/specter/py/')
```
sbailey: These should be configured by the user prior to running the scripts; personal directories specific to a single machine shouldn't be in the scripts themselves.

forero:
Good. I removed that setup from the script.

PR #20, opened 2015-09-12
Format flow

forero:
Poor man's routines selecting observable fields close to zenith and defining bright/dark time. Makes heavy use of pyephem.


forero:
Just got Stephen's email about https://github.com/desihub/desisurvey/blob/master/py/desisurvey/nextfield.py

PR #22, opened 2015-09-19
updated to rawdata_root

profxj:
Small edit to handle desispec update.


sbailey:
D'oh.  Sorry I changed this an didn't catch the cross-repository impact.  Near term plans include adding QA code to the integration tests to catch stuff like this.

PR #24, opened 2015-09-29
Cosmics

sbailey:
This pull request has 3 primary pieces:

  * Adds ability to add cosmics to pixsim images, using a library of input cosmics.  This is the primary algorithmic update in this PR.
  * Partially refactors I/O to be more like desispec, separating algorithms from I/O from the file/directory locations.  More could be done in this direction.
  * Adds unittests to confirm that the above changes still work.  These currently require large input template files and lots of DESI* environment variables to be set, so they aren't ready for Travis integration.  But having them is better than nothing.  No coverage yet for
    * TargetTile and TargetSurvey objects in desisim.target
    * desisim.templates

This also renames version() to gitversion() since it was returning the non-PEP 440 `git --describe ...` version string.  Unit tests confirm that this is consistent with the `__version__` string (i.e. they both start from the same base tag).  This will likely evolve in the future.


moustakas:
Looks good.

PR #28, opened 2015-10-12
Fix the spectrograph higher than 0 and nspec beyond 500

gdhungana:
This PR focusses the suggested changes and addresses the issues of not going beyond 500 spectra or spectrograph 1-9. Also did some recorganization for writing the files etc. Can well test a ~5000 spectra simspec file. I only had ~ O(2000), where all the suggested tests worked fine.

### DIFF
--- a/bin/quickgen
+++ b/bin/quickgen
@@ -9,8 +9,9 @@
 #	3. Sky Model File (x3)
 #	4. cframe file (x3)
 #
-#Example:
-#
+# Example:
+# 
+# G. Dhungana+, Spring 2015 
 #********************************
 
 import argparse
@@ -38,7 +39,7 @@ parser=argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpForm
 
 parser.add_argument("--simspec",type=str, help="input simspec file")
 parser.add_argument("--fibermap",type=str, help='input fibermap file')
-parser.add_argument("--nspec",type=int,default=500,help='no. of spectra to be simulated, starting from first')
+parser.add_argument("--nspec",type=int,default=5000,help='no. of spectra to be simulated, starting from first')
 parser.add_argument("--nstart", type=int, default=0,help='starting spectra # for simulation 0-4999')
 parser.add_argument("--spectrograph",type=int, default=None,help='Spectrograph no. 0-9')
 args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


bin/quickgen
```diff
@@ -185,15 +186,15 @@ maxbin=max(waveMaxbin['b'],waveMaxbin['r'],waveMaxbin['z'])
...
-nobj=np.zeros((500,3,maxbin))     # Object Photons
-nsky=np.zeros((500,3,maxbin))      # sky photons
-nivar=np.zeros((500,3,maxbin))     # inverse variance  (object+sky)
-sky_ivar=np.zeros((500,3,maxbin)) # inverse variance of sky
-cframe_observedflux=np.zeros((500,3,maxbin))    # calibrated object flux
-cframe_ivar=np.zeros((500,3,maxbin))    # inverse variance of calibrated object flux
-frame_rand_noise=np.zeros((500,3,maxbin))     # random Gaussian noise to nobj+nsky
-sky_rand_noise=np.zeros((500,3,maxbin))  # random Gaussian noise to sky only
-cframe_rand_noise=np.zeros((500,3,maxbin))  # random Gaussian noise to calibrated flux
+nobj=np.zeros((5000,3,maxbin))     # Object Photons
```
rainwoodman: Why not use the spec value passed in from the arguments? This one seems will die with more than 5000 specs.

gdhungana: I am not sure there will be a single sim file with more than 5000 specs. But if yes then we might want to change that. Earlier I was using 500 for a single spectrograph but now I looped outside of this, so there is 5000 corresponding to 10 spectrographs.

sbailey:
@rainwoodman is right that it would be better to base the allocated array sizes on the input options, or at least catch errors if the input dimensions or options don't make sense (e.g. requesting to simulate an exposure with 6000 spectra).  Or even moving the hardcoded 500 and 5000 values into named variables would be good.

At first look, it appears that this over-allocates memory for 5000 objects, even if it is only using 500 of them, e.g. to fill 3000:3500.  This is a convenience that makes the code indexing easier; we could fix that if the memory footprint becomes problematic.

None-the-less, I will merge this now to get something working into the repo, and we can cleanup these issues in a second pass.

gdhungana:
Agreed. The original thought was that the sim file will be limited to 5000 spectra(?). We definitely don't want to do this for something like resolution_data, which will increase the matrix size tremendously. That will also allocate the memory for 5000 objects when have that many spectra to simulate, which can likely be the case later on. Let's work/test on this for now and I can work on the second round of limiting objects by nspec. Thanks @sbailey @rainwoodman !

PR #31, opened 2015-10-23
add support for generating white dwarf and QSO templates

moustakas:
Dedicated templates for white dwarfs can now be generated.  For example, to generate 100 templates with g-band magnitudes in the range [16,19] and no radial velocity jitter do:
```
simulate-templates -o WD -n 100 --gmagrange-wd 16 19 --vrad-wd 0 0
```
Note that you first need to set the DESI_WD_TEMPLATES environment variable and to grab the corresponding "basis" templates from NERSC, e.g.,
```
export DESI_WD_TEMPLATES  ${DESI_ROOT}/spectro/templates/wd_templates/v1.0/wd_templates_v1.0.fits 
```

These templates are based on the theoretical DA white dwarf models built by Carlos Allende Prieto spanning a range of Teff [20,000-65,000 K] and log-g [7.0-9.5 cm/s] at high spectral resolution from 0.2-2.5 micron.

The I/O routines in desisim.templates were also moved to desisim.io.

Proper documentation for these templates plus optional QA plots on output are still forthcoming...

### DIFF
--- a/py/desisim/io.py
+++ b/py/desisim/io.py
@@ -500,6 +500,98 @@ def read_templates(wave, objtype, nspec=None, randseed=1, infile=None):
     return outflux, outmeta
     
 
+def read_base_templates(objtype='ELG', observed=False, emlines=False):
+    """Return the base, rest-frame, spectral continuum templates for each objtype.
+
+    The appropriate environment variable must be set depending on OBJTYPE.  For example,
+    DESI_ELG_TEMPLATES, DESI_LRG_TEMPLATES, etc., otherwise an exception will be raised.
+
+    Args:
+      objtype (str, optional): object type to read (ELG, LRG, QSO, BGS, STD, or STAR;
+        defaults to 'ELG').
+      observed (bool): Read the observed-frame templates (defaults to False).
+      emlines (bool): Read the spectral templates which include emission lines (defaults
+        to False; only applies to object types ELG and BGS).
+
+    Returns:
+      flux (numpy.ndarray): Array [ntemplate,npix] of flux values [erg/s/cm2/A].
+      wave (numpy.ndarray): Array [npix] of wavelengths for FLUX [Angstrom].
+      meta (astropy.Table): Meta-data table for each object.  The contents of this
+        table varies depending on what OBJTYPE has been read.
+
+    Raises:
+      EnvironmentError: If the appropriate environment variable is not set.
+      IOError: If the base templates are not found.
+    
+    """
+    from astropy.io import fits
+    from astropy.table import Table
+    from desispec.io.util import header2wave
+
+    otype = objtype.upper()
+    if otype=='FSTD':
+        otype = 'STAR'
+
+    key = 'DESI_'+otype+'_TEMPLATES'
+    if key not in os.environ:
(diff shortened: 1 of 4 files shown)


py/desisim/io.py
```diff
@@ -500,6 +500,98 @@ def read_templates(wave, objtype, nspec=None, randseed=1, infile=None):
...
+
+    flux, hdr = fits.getdata(objfile, 0, header=True)
+    meta = Table(fits.getdata(objfile, 1))
+    wave = header2wave(hdr)
+
+    return flux, wave, meta
+
+def write_templates(flux, wave, meta, objtype, outfile=None, comments=None,
+                    units=None, header_comments=None):
+    """Write out simulated galaxy templates.  (Incomplete documentation...)
```
sbailey: In `desispec.io`, the standard is `write_blat(outfile, data...)` rather than `write_blat(data, ..., outfile)`.  The use of `outfile=None` here appears to be preparing for the possibility that someone would call this without giving the output filename, but that just causes the current code to crash.  Is there a planned case for when not specifying `outfile` would be useful?  If not, please reorder to make outfile the first required parameter for consistency with other desihub code.

moustakas: The optional outfile was put in to specifically to accommodate a capability you had asked for, namely, the ability to generate templates on-the-fly without having to write them out or read them in.  However, you're right that the code will crash as written.  I'll fix this but try to retain this capability.

sbailey: Generating templates on-the-fly is still desired, but in that case the code should just never call write_templates() at all.  i.e. the algorithm code should be able to generate templates and return them, and wrapper script calls that algorithm and then calls write_templates with the results.  If someone else wants to use the algorithm to generate templates but not write them out (or write them out differently), they can just call the algorithm directly and move on.

sbailey:
Is this PR stuck on my questions about the optional outfile?

moustakas:
desisim.io.write_templates now requires outfile and the I/O was refactored so that the templates are written out by the top-level (bin/simulate-templates) script rather than the individual spectral classes.  This also allows templates to be generated in memory without having to write them out, which perhaps gets us closer to how we want QuickGen to interact with the templates code (which is still TBD).

gdhungana:
This may be little off topic for this, but just clearing a bit. QuickGen ( and also pixsim-desi) will interact to the new templates only through the simspec file and not directly with the templates code. So relevant template generation should go inside the newexp-desi (if that is the plan) which would read the base templates, simulates based on input and outputs to simspec file. Equivalently newexp-desi can also read the simulated templates from the output file of bin/simulate-templates, and this is similar to the case we are adopting for data_challenge 2 templates, albeit the templates are fixed there. This second option will always keep simulate-templates separate from newexp-desi. Isn't that right?

moustakas:
Monte Carlo-based QSO templates can now also be generated:
```
simulate-templates -o QSO -n 100 --gmagrange-qso 21 23 --zrange-qso 0.5 4.0
```

@gdhungana -- newexp-desi will indeed be modified to build templates on-the-fly by calling the relevant code in desisim.templates, thereby bypassing the command-line script 'simulate-templates'.  I'm working on this now.

PR #46, opened 2015-11-17
Put a zeroth order redshift efficiency as a place holder

gdhungana:


sbailey:
Apologies for the delay in reviewing this.  Three requests:
1. apply_redshift_eff: this is applying noise to the redshift *precision* (how precise is the z measurement), not changing the *efficiency* (how many targets do we successfully get a redshift?).  This would be better named "apply_redshift_precision" or leave it as apply_redshift_eff and have it return both new z (precision) and zwarn (for downgrading efficiency; zwarn != 0 means uncertain redshift).
2. The actual RMS error isn't constant with z; deltaz/(1+z) is more likely the uniform RMS quantity.  Re-parameterize the error in terms of deltaz/(1+z) and then add the z noise as RMS*(1+z).  i.e. for the samples you have, calculate RMS( (z-ztrue) / (1+ztrue) ) and use that RMS*(1+z) for the error you add.
3. Add a command line option --perfectz to maintain the "v0" feature of redshift in = redshift out with perfect precision and efficiency.

Thanks!

weaverba137:
Is there still a plan to merge this?  This PR is approaching 4 months old.

sbailey:
Closing this out, to be replaced with PR #85 .

Issue #52, opened 2015-12-16
ELG.make_templates seed problem (desisim/py/desisim/templates.py)


cballand:
When running quickbrick to simulate ELG spectra, it appears that for some seed values, make_templates is getting trapped in an infinite loop:

quickbrick -b elg-100 --objtype=elg -n 100 --wavestep=0.2 --downsampling=5 --seed=36278

INFO:io.py:466:read_basis_templates: Reading /Users/balland/software/DESI/templates/elg_templates_v1.5.fits
(nobj gflux rflux zflux)
0 0.389182203653 0.889768319563 2.82101159881
0 0.34244854018 0.520201495332 1.83512049751
0 3.2059061274 3.65286151342 8.91696554622
1 0.596939871811 0.690797688129 1.07839839904
2 0.730462842891 1.22903017783 2.84984074649
2 0.44975232066 0.488796456556 0.945432319503
3 1.25288209136 1.54069514849 2.29774914482
4 1.20394097408 1.7500975788 3.30527697546
5 1.9869155923 2.19026235619 5.25214423389
6 3.46967750653 3.53729113706 5.63974807984
7 0.500691078679 0.938085739886 4.22415601743
7 0.434556737135 0.548816630125 1.26927907769
8 0.734986473212 1.05794442062 3.1308010843
8 0.736668451755 1.37582535989 3.78068489899
8 0.719109088305 0.695289532948 1.29016218049
9 2.66904210178 3.20630371357 9.18626840421
9 1.80427167496 2.53900820452 7.9647101176
9 1.68421749473 3.10725360778 12.0505634408
9 0.753069301385 1.72120610228 4.16594234824
9 0.735655517961 0.737320745965 2.26880454757
9 0.549928377833 0.935599046027 2.83142791768
9 1.88236666985 2.45955548667 10.6525302436
9 1.61184959441 2.16935519679 12.4670519826
9 1.03290594521 1.34613820197 8.10347802548
9 1.23061306005 2.92369667412 19.2532084199
9 1.21513103363 1.58595420827 30.2256374531
9 1.49554906193 2.45919176657 78.9215522665
...
9 0.448388329099 0.661248172651 4.73202638011e+28
9 9.77195130461e+23 1.93529094339 2.7037030858e+29
9 0.566638308873 0.655562775011 9.51792409584e+28
9 0.783621573571 1.59963579161 3.65527790031e+29
9 2.96939642325e+24 1.71422942333 8.28233438182e+29
[...]

PR #53, opened 2015-12-17
Fix issue/bug #52 and add fiducial velocity broadening to the LRG templates

moustakas:
This PR fixes issue #52 and also adds (intrinsic) velocity broadening to the LRG templates based on some reasonable priors.  The smoothing is done using a module (pixelsplines) written by Bolton which caused some problems when broadening spectra with emission lines (i.e., the ELG spectra), so that was left off for now.  (The emission lines themselves, however, still have intrinsic velocity broadening.)


weaverba137:
Once this is merged, we'll need to create a new tag of desisim, so we can proceed to generating the test data for the redshift data challenge.

PR #56, opened 2015-12-18
implement specsim into quickgen

rstaten:
I have made a few changes to quickgen in comparison with changes to quickbrick in order to implement specsim. After these changes, the output is identical to the quickgen output when using quicksim.


bin/quickgen
```diff
@@ -23,6 +23,7 @@ import sys
 import desimodel.simulate as sim
```
dkirkby: Is this import still needed for anything?

dkirkby:
It looks like `quicksim` is still mentioned in some comments (`__init__.py` and `quickgen`).

dkirkby:
@rstaten for future reference, you don't need to clone a desihub repo in order to make a PR.  Instead, you should normally put your changes directly on a new branch of the desihub repo, then issue a PR to merge your new branch into master.

What you did here, instead, is to request a merge of your cloned master into a new desihub branch.  I am going to merge this now, but then a second merge will be required to bring your changes into the desihub master branch.

PR #60, opened 2016-01-04
Modification of quickbrick output in channel b brick

cballand:
Following a comment by @sbailey, I modified quickbrick to write the pre-downsampling high res source fluxes of simulated spectra along with the corresponding wavelength grid into the b-channel brick. @sbailey : Once merged into the master branch, could you tag a new version of desisim and have it installed on cori, so that I can re-generate the training sample for the challenge (using the same seeds as before so that the simulated spectra are the same as the current version).

### DIFF
--- a/bin/quickbrick
+++ b/bin/quickbrick
@@ -266,7 +266,7 @@ for channel in ['b', 'r', 'z']:
         header = desispec.io.fitsheader(header)
         fx.append(fits.ImageHDU(trueflux[channel][:,iiobs].astype(np.float32), name='_TRUEFLUX', header=header))
         if (channel == 'b'):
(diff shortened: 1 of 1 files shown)


sbailey:
I made a minor change of _FULLWAVE -> _SOURCEWAVE to make it more obvious that that is the wavelength HDU that corresponds to the _SOURCEFLUX HDU.  Otherwise this looks good.

I haven't been able to reproduce the zdc1 training spectra (see desi-data email thread) even using the original tags, so I'm not merging this yet just so that there aren't too many moving pieces while you are debugging that.  You can merge this PR when you are ready.

sbailey:
We haven't solved the reproducibility issue, but I am going to go ahead and merge this change into master, since it is useful and seems unrelated to how the random number generators are used.

Issue #26, opened 2015-10-10
quickgen can't simulate spectra above 500


sbailey:
Now that quickgen has been announced to the collaboration, bugs in it become high priority.

The options for simulating spectra above 499 appear to be broken.  All of these should work:
```
quickgen --simspec {simspecfile...} --fibermap {fibermapfile} --nspec 500 --nstart 500 --spectrograph 1
quickgen --simspec {simspecfile...} --fibermap {fibermapfile} --nspec 500 --nstart 500
quickgen --simspec {simspecfile...} --fibermap {fibermapfile} --nspec 1000
quickgen --simspec {simspecfile...} --fibermap {fibermapfile} --nspec 5000
```
But they all fail with some version of:
```
Simulating spectrum 500,  object type=ELG
Traceback (most recent call last):
  File "/project/projectdirs/desi/software/edison/desisim/master/bin/quickgen", line 257, in <module>
    nobj[j,i,:waveMaxbin[channel]]=results.nobj[waveRange[channel],i]
IndexError: index 500 is out of bounds for axis 0 with size 500
```
As a result, quickgen can only simulate spectrograph 0 and not 1-9.  Please fix.

gdhungana:
@sbailey I just made a PR. The change is verified for all suggested options. Can even give any arbitrary values for nspec or nstart. like, 

quickgen --simspec {simspecfile...} --fibermap {fibermapfile} --nspec 240 --nstart 950

PR #61, opened 2016-01-08
fixes (more!) random seed issues for ELGs identified in #59

### LINKED ISSUE #59: Minor random number issues for ELG templates
I noticed two potential problems with how random numbers are generated in the ELG template code, both in `templates.ELG.make_templates()` [line numbers refer to 497e2e9]:

1. The [OII] equivalent width calculation adds some Gaussian dispersion about the central value calculated from D4000, but uses the same dispersion value for all spectra in a "chunk" [line 255]. I think the solution is simply to replace `rand.normal(0.0,0.3)` with `rand.normal(0.0, 0.3, nchunk)`.
2. The calls to `EMSpectrum.spectrum()` used to add emission lines to each continuum template always pass the same seed [line 274], so it looks like all calls to `np.random.*` will return the same values for each simulated emission spectrum.

Unfortunately, a test of reproducibility with the same seed does not catch either of these issues since they both lead to spectra that are too reproducible!

moustakas:
This PR fixes #59 and moves the ball forward a bit toward using desitarget.cuts for the color-cuts. The filter convolution stuff has not yet be factored out (will be moved to desiutil) but that's next.  All tests pass (although there was a divide-by-zero in the synthesized photometry I'd like to track down).


sbailey:
I made a minor change to not import desitarget.cuts since it isn't used (yet).

We should do the desitarget cuts per-objtype refactor ([desitarget issue #26](https://github.com/desihub/desitarget/issues/26)) first and then use that rather than shoe-horning this into the current apply_cuts.

Waiting for the Travis tests and then I'll merge this.

PR #63, opened 2016-01-13
In quickbrick: replace actual srcflux in b brick by pre-downsampling …

cballand:
…high res source flux and corresponding wavelength grid. Fixes a bug with option --outdir_truth. Updates __version__ string to 0.8.2


cballand:
Merging PR #63

PR #79, opened 2016-02-04
Testlite

sbailey:
I'm not ready to merge this yet, but turning it into a pull request so that I can see coverage statistics — it appears that Travis runs the coverage tests on branches but that doesn't get propagated to the coveralls badge or the coveralls gui.

Basic idea of this branch: I created a *separate* desispec-testdata repo with trimmed down basis templates (and the script that trims them down).  This can be used as a replacement for $DESI_BASIS_TEMPLATES for testing.  This can be fetched from https://github.com/desihub/desisim-testdata/archive/master.zip — it's pretty small and doesn't contain the history either.  At the same time, if it does get unwieldy we can blow that repo away and start over without losing our code history in this repo.

A lot of tests are still skipped due to needing cosmics templates under $DESI_ROOT and needing $DESIMODEL itself.  Trimmed down cosmics could be moved into desispec-testdata, and @weaverba137 is working on a trimmed down version of desimodel.

On my laptop, nosetests says there is 41% coverage with only $DESI_BASIS_TEMPLATES set (without $DESIMODEL or $DESI_ROOT).  Let's see if coveralls agrees...

@weaverba137 — please don't merge this yet, but please do take a look to see if you think this is a reasonable idea.

Credit to Kyle Barbary too for a brainstorming session that came up with this separate-repo-for-test-data idea.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ env:
         - SCIPY_VERSION=0.14
         - ASTROPY_VERSION=1.0.4
         - SPHINX_VERSION=1.3
-        - DESIUTIL_VERSION=1.3.2
+        - DESIUTIL_VERSION=1.3.3
         - SPECTER_VERSION=0.3
         - DESISPEC_VERSION=0.3.1
         - DESIMODEL_VERSION=trunk
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,13 @@
 desisim
 =======
 
+.. image:: https://img.shields.io/travis/desihub/desisim.svg
+    :target: https://travis-ci.org/desihub/desisim
+    :alt: Travis Build Status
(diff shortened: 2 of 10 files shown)


sbailey:
@weaverba137 this is ready for review.  Please review these together:
  * desisim branch testlite (this PR)
  * desisim-testdata (used by this PR)
  * desimodel branch trimdata and its [PR #3](https://github.com/desihub/desimodel/pull/3), used to create desimodel branch test-0.4, which is also used by this PR

weaverba137:
This looks good to me, & tests are passing, so merge when ready.

Due diligence:

- This branch looks good.
- desisim-testdata looks good.
- I like that desimodel/test-0.4 has no files that are even larger than a megabyte, as far as I can tell.

Issue #78, opened 2016-02-03
basis templates aren't FITS standard compliant, breaking some FITS readers


sbailey:
The elg, lrg, star, and wd basis templates in /project/projectdirs/desi/spectro/templates/basis_templates/v1.1/ aren't FITS-standard compliant (the QSO templates are).  In particular, HDU2 has an extraneous SIMPLE=T second keyword, when BITPIX is supposed to be the second keyword.  These files work with astropy and mrdfits, but cause the more picky cfitsio to fail (and thus fitsio too).

Current code works because astropy.io.fits is our standard fits reader, but the next iteration on templates should fix this as well.

moustakas:
OK, I think this was a pretty easy fix -- just needed to remove the SIMPLE=T keyword.  Will close this when I release the new set of basis templates (<days).
```
import fitsio
wave, hdr = fitsio.read('bgs_templates_v2.0.fits',ext=2,header=True)
print(hdr)
XTENSION= 'IMAGE   '           /Image Extension created by MWRFITS v1.12b
BITPIX  =                  -32 / Number of bits per data pixel
NAXIS   =                    1 / Number of data axes
NAXIS1  =                33281 /
WAVEMIN =              950.000 / minimum wavelength
WAVEMAX =              40673.7 / maximum wavelength
AIRORVAC= 'vac     '           / wavelengths in vacuum (vac) or air
CUNIT1  = 'Angstrom'           / units of wavelength array
PCOUNT  =                    0 /
GCOUNT  =                    1 /
```

sbailey:
For the record, these templates still fail fitsverify due to:
```
=================== HDU 3: Image Exten. ====================
 
*** Error:   PCOUNT is not in record 5 of the header.
*** Error:   GCOUNT is not in record 6 of the header.
```
However, I think that is actually a bug in fitsverify, not in these files themselves.  The [fits standard](https://archive.stsci.edu/fits/fits_standard/node39.html#SECTION00941200000000000000) specifies that PCOUNT and GCOUNT must exist, but is pretty clear that they don't have to be keywords 5 and 6 as fitsverify seems to think.

Closing this ticket as fixed.

Issue #70, opened 2016-01-29
Add BGS-like templates


sbailey:
Add a template class appropriate for simulating Bright Galaxy Survey (BGS) targets.  A short term fallback hack could be to use the existing LRG class with a custom zrange, zmagrange, and nocolorcuts=True.

sbailey:
Fixed via PR #81 .

Issue #64, opened 2016-01-14
objtype STAR vs. different flavors of stellar targets


crockosi:
The template generator for STARs need to be a little more flexible to accept more flavors of STAR targets. For example, right now the input gmag distribution is used as the input WD mag distribution and the input rmag for the all other star templates.  Can all the target mag distributions be taken from targets.dat in desimodel?  Also, can we make the input flavor of star (WD, FSTD, MWS) and argument rather than individual true/false flags?  Right now the only way to specify sub-types is WD=True or FSTD=True. For example, I can imagine wanting to have multiple flavors of STAR_MWS so I can get different target cuts and even now the MWS wants a different range than the FSTDs, and I expect the BAD_QSO stars want yet a different mag distribution. 

Related, the same objtype used for magnitude limits and TS cuts is passed to specter/throughput.py and used to compute the fiber losses. It would be nice to be able to specify just "star" to specter, since it doesn't need more than that. Right now specter uses the "default input loss" and just generates a warning, but here as well it would be good to distinguish target type from object type. 

I'm happy to make these changes, but need input on where some of the default data (like color cuts) should go and  I don't want to give grief to all uses of FSTD=True code without warning.

moustakas:
> The template generator for STARs need to be a little more flexible to accept more flavors of STAR targets. For example, right now the input gmag distribution is used as the input WD mag distribution and the input rmag for the all other star templates. Can all the target mag distributions be taken from targets.dat in desimodel? Also, can we make the input flavor of star (WD, FSTD, MWS) and argument rather than individual true/false flags? Right now the only way to specify sub-types is WD=True or FSTD=True. For example, I can imagine wanting to have multiple flavors of STAR_MWS so I can get different target cuts and even now the MWS wants a different range than the FSTDs, and I expect the BAD_QSO stars want yet a different mag distribution.

Thank you for taking a look @crockosi !  We began with very simple (i.e., flat) priors to get something in place but will certainly support more options, e.g., various input magnitude / redshift distributions, selection bands, color cuts, etc.  But let's start with supporting other types of stars.  How do you want to specify the spectral types?  I'm using Carlo's latest models, documented here:
  https://desi.lbl.gov/DocDB/cgi-bin/private/RetrieveFile?docid=870;filename=stellar-templates.pdf;version=1

The FSTD class is based on a simple selection in (DECaLS) grz color-color space:
[...]

Issue #80, opened 2016-02-14
Update to new specsim v0.3 API


dkirkby:
I am in the middle of some substantial changes to specsim that move all of the hardcoded desimodel dependencies into a yaml configuration file, with some unavoidable API changes.  These changes lay the groundwork for simulating other instruments (eBOSS, for example) and also running Travis tests without desimodel.

This issue is to update desisim's usage of specsim when v0.3 is released with the new API.  I believe that the only affected code is `bin/quickgen` and `bin/quickbrick`.   A version requirement `specsim >= 0.3` will also need to be imposed, but I'm not sure how to do that (it looks like `specsim` is not mentioned at all in the current `requirements.txt`).

dkirkby:
I am using the output generated by the following command, with specsim v0.2 installed, as my baseline for working on this issue:
```
quickbrick -b original --objtype qso -n 10 -o . -s 123
```
This generates some warnings which I will ignore:
```
INFO:io.py:457:read_basis_templates: Reading /Data/DESI/basis_templates/v1.1/qso_templates_v1.1.fits
/Users/david/Cosmo/DESI/code/desisim/py/desisim/filterfunc.py:68: RuntimeWarning: invalid value encountered in double_scalars
  flux /= np.trapz(self.interp(wave1)*wave1,wave1)
```

moustakas:
This is the old filter code complaining (since the desisim branch I've been developing which uses speclite.filters) has not yet been merged.  However, this code is *also* producing non-catastrophic warnings which I've decided to not yet suppress while we're still vetting & testing these ~major changes.

dkirkby:
Using branch `speclite_api` and the `specsim` release candidate (with the new API) I now get identical contents for HDUs 0-4 in all three brick files using:
```
quickbrick -b updated --objtype qso -n 10 -o . -s 123
```
There are now some new warnings from astropy (via specsim) but these are a separate issue (deishub/desimodel#4):
```
WARNING: UnitsWarning: The unit 'Angstrom' has been deprecated in the FITS standard. Suggested: 10**-1 nm. [astropy.units.format.utils]
WARNING: UnitsWarning: 'Angstrom/row' did not parse as fits unit: At col 9, Unit 'row' not supported by the FITS standard.  [astropy.units.core]
INFO:io.py:457:read_basis_templates: Reading /Data/DESI/basis_templates/v1.1/qso_templates_v1.1.fits
[...]

PR #81, opened 2016-02-15
add BGS templates and speclite & desitarget dependencies

moustakas:
I'm creating this PR now for two reasons even though the code and templates are still in a state of flux/testing.  The first reason is that BGS templates were promised to our Overlord by Valentine's day (i.e., today).  And second, I want @crockosi to be able to move forward with #64 and not have to wait around for me any more.  

In addition to adding a basic set of BGS templates (see #70) I have also updated the ELG, STAR, WD, and LRG templates in order to address #78 .  In addition, the ELG templates (v1.5-->v2.0) now utilize Conroy's higher-resolution continuum models and the STAR models (v1.2-->v2.0) extend redward to 6 microns (thanks to Carlos!).

Consequently, once this PR is merged $DESI_BASIS_TEMPLATES should point to $DESI_ROOT/spectro/templates/basis_templates/v2.0

However, I still have to clean up some issues with the ELG and BGS templates, so user beware until everything is stable.

This PR also brings in an explicit dependence on the filter convolution stuff in (the master branch of) speclite (with many thanks to @dkirkby ).  This dependence is still a bit fragile (e.g., the desisim.templates.QSO class isn't quite working until I can pad the spectra) and on the target selection cuts in desitarget.cuts.

I am also in the process of updating the emission-line spectrum priors in desisim.templates (which will also be part of this PR), so stand by for a bit more code.  Updated TechNotes for all these spectral classes (including a new one for the BGS templates) are also forthcoming.


sbailey:
Tests were failing due to a bad scipy install; it is unclear why that particular combination of astropy/scipy/numpy stopped working (welcome to dependency hell...), but it is the same symptom that was fixed by updating versions in desispec.  I bumped the versions here and we'll see...

sbailey:
Thanks for the BGS and other template updates, @moustakas.  If the code and template updates will be stable by the end of this week, let's wait on merging this PR.  In the meantime @crockosi can work off of a branch of this branch for the purposes of adding the BGS and bright flavors to newexp (#71) and we could merge them all in at the end of the week.

@dkirkby what is the timescale for speclite v0.3 with the updated config API (see #80)?  Since this PR moves speclite to a core dependency instead of an isolated optional dependency, it would be preferable to do this after the API change so that we don't have to change again very soon.

For the record: I originally opposed making speclite a core dependency, just to avoid further dependency proliferation.  However, I admit that the new filter code in speclite is a thing of beauty, and much better than any of the multiple versions of similar functionality that we've written in house, so it makes sense to depend on that.  I withdraw my objection (as long as galsim doesn't become a core dependency of speclite...see comment thread in #77).

sbailey:
Tests are failing due to:

    File "/home/travis/build/desihub/desisim/py/desisim/templates.py", line 887, in make_templates
        decam_flux = self.decam.get_ab_maggies(flux, zwave, mask_invalid=True)
    TypeError: get_ab_maggies() got an unexpected keyword argument 'mask_invalid'

speclite is installed in etc/travis_env_common.sh with "pip install speclite" without specifying a version, which presumably gets the latest registered tag.  If "mask_invalid" is a feature in master that hasn't made it into a tag yet, can you cut a new tag @dkirkby?  And/or @moustakas is there a reasonably simply way to work around using this keyword?

General guideline about dependencies: it is ok for us to temporarily depend upon master within our own desihub packages while working toward the next set of consistent tags (e.g. a feature added to desiutil to be used by desispec; once both are vetted together, a new tag for both).  But for external products we should only depend upon tagged released versions, not master (speclite is admittedly a special case in the definition of "external", but we should still aim for only using tagged versions)

dkirkby:
@sbailey The API change mentioned in #80 is for specsim, not speclite, so does not need to hold up this PR (confusing package names, I know!)

I totally agree that dependencies should be versioned and I will tag speclite v0.4 as soon as John and I converge on a few more details (dkirkby/speclite#12).

dkirkby:
I am ready to tag speclite v0.4 once I get the green light from John.

@weaverba137 What is the correct way to specify a minimum version requirement for pip installation and travis testing?

moustakas:
Apologies, yes, please feel free to tag.

dkirkby:
I just tagged speclite v0.4 with improvements suggested by John (thanks!).

The travis tests are still failing, but not because of speclite version problems anymore (but I still think we want to specify the >= v0.4 requirement somewhere).

The failing tests are now `test_random_seed` and `test_stars `.

weaverba137:
Pip: http://pip.readthedocs.org/en/stable/user_guide/#requirements-files
Travis: you'll have to edit the .travis.yml file and/or the travis_env_common.sh file.

sbailey:
@moustakas it used to be the case that WD stars didn't have an FEH column in their metadata (and there was a test to confirm that).  Now they do have an FEH column, but the entries are left as all zeros.  Is that the intended behavior?

sbailey:
@moustakas the ELG metadata WISE_FLUX has NaN values for w3 and w4 (columns 2 and 3).  Is this intended?  is NaN distinctly better than filling them in with 0 or -1, or perhaps not including those columns at all?

The random seed consistency test is failing on this since NaN != NaN.  We can catch that as a special case in the tests if you really really want NaNs in there, but I thought I'd check first...

dkirkby:
@moustakas Do you need W3, W4 magnitudes for any targets?  If not, you could just remove these filters from your list, e.g.
```
my_filters = speclite.filters.load_filters('decam2014-*', 'wise2010-W1', 'wise2010-W2')
```

moustakas:
I'll fix these tests (and apologies for not fixing them in the first place).  A couple quick comments on this thread:

I should probably take (back) out the [Fe/H] metadata column (of zero value) for the white dwarfs, unless @crockosi weighs in and tells me/us otherwise.

Regarding the NaN's in WISE_FLUX, I don't know what we should put here.  Maybe zero would be better (and 99 for the corresponding magnitudes, e.g., W1MAG, W2MAG, which are computed and stored simply for convenience). I added these columns so the metadata table would look more like a Tractor catalog.  The idea would be to simulate a broad set of spectra *without* color cuts and then apply color cuts (via desitarget) after the fact (e.g., in order to explore redshift efficiency in other parts of color space).  

But looking a bit more carefully at the target-selection code it looks like WISE_FLUX doesn't need to be a 4-element array, so I could just drop the W3/W4 photometry wholesale (as @dkirkby suggested, which I was aware of).  Then the only undefined photometry (for now) would be the W1/W2 photometry of the QSOs and of the white dwarfs (all the other spectral models go red enough).

crockosi:
No preference for having the [Fe/H] metadata for the WDs.

I could throw that question about useful metadata on the WDs to the MWS
working group, where there are a some experts, but for now it is probably
quickest to remove it.

--Connie

moustakas:
OK, I think this PR is ready to merge.  In addition to ~finalizing the ELG (v2.0) and BGS (v2.0) templates (currently transferring to NERSC...), this PR should also close #78, #70, and #8.  It also fixes the issues identified above by @sbailey.

There have been some significant changes/updates to the templates but I've done as much spot-checking as I can do for now.  All tests pass on my laptop (except the ones that depend on files I don't have...)

I still owe the team complete documentation (still in prep) but some high-level stuff is here:
  https://github.com/desihub/desisim/blob/desisim-templates/doc/nb/calib-emline.ipynb
Thanks to @dkirkby for helping with the Gaussian mixture model stuff.

sbailey:
Travis tests are failing, but I think that is due to out-of-date desisim-testdata templates.  I can track this down tomorrow when I have a faster internet connection for moving templates around.

I'm also getting an unexpected success on flagging that QSO and stellar templates don't have WISE mags.  Check and fix test if they do have WISE mags now.

+1 for using ecsv format for the data files.

Code looks good, but I want to get the Travis tests fully working with desisim-testdata before merging.

moustakas:
The "expected failure" test on the wise photometry of the QSO and WD templates (note: the stars now go red enough) is now fixed.

PR #82, opened 2016-02-17
Quickbrick zrange

julienguy:
Includes z range option in quickbrick


julienguy:
I had some travis failed tests but this has nothing to do with the minor modification in this branch.

sbailey:
In general, we should always fix tests even if they are broken for reasons not directly related to the PR.  However, in this case it appears to be a temporary communication failure between travis and github, with log messages like
```
$ cd desihub/desisim
$ git fetch origin +refs/pull/83/merge:
fatal: Couldn't find remote ref refs/pull/83/merge
Unexpected end of command stream
The command "eval git fetch origin +refs/pull/83/merge:" failed. Retrying, 2 of 3.
```
Tests are running again and failing on master now, though perhaps unrelated to this merge.  Debugging that separately.

PR #86, opened 2016-02-18
Update quickgen and quickbrick to new specsim API

dkirkby:
This is not ready to merge until I tag and release specsim v0.3, but now passes my tests on #80 using desihub/specsim#29.  Please review if you are interested.  I am finding identical results after the updates, but your mileage may vary.

Note that there are some trivial diffs in this PR where my editor (atom) automatically removed the invisible white space from the end of lines.  Sorry about that.  Could we standardize our whitespace conventions using [editorconfig](http://editorconfig.org)?   For example, the conventions I use in specsim are specified [here](https://github.com/desihub/specsim/blob/master/.editorconfig).

### DIFF
--- a/bin/quickbrick
+++ b/bin/quickbrick
@@ -14,6 +14,7 @@ import scipy.sparse
 import string
 
 from astropy.table import Table, Column, vstack
+import astropy.units as u
 
 import desimodel.io
 import specsim
@@ -199,7 +200,7 @@ meta.add_column(Column(targetids, name='TARGETID'))
 meta.rename_column('REDSHIFT', 'TRUEZ')
 
 #- Actually do the simulations for each target
-fluxunits = specsim.spectrum.SpectralFluxDensity.fiducialFluxUnit * 1e17
+fluxunits = u.erg / (u.s * u.cm ** 2 * u.Angstrom)
 source = specsim.source.initialize(config)
 
 #- output arrays to fill; these cover wavelengths across all cameras
--- a/bin/quickbrick
+++ b/bin/quickbrick
@@ -17,7 +17,8 @@ from astropy.table import Table, Column, vstack
 import astropy.units as u
 
 import desimodel.io
-import specsim
+import specsim.config
+import specsim.simulator
 import desispec
 from desispec.log import get_logger
 import desisim.templates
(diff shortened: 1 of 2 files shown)


sbailey:
I was surprised by this snippet:
```
config = specsim.config.load_config('desi')
...
config.wavelength_grid.min = desimodel.io.load_throughput('b').wavemin
config.wavelength_grid.max = desimodel.io.load_throughput('z').wavemax
```

It seems that wavemin/wavemax from desimodel should be part of the default desi config.  I haven't looked under the hood for how or where specsim is getting that desi config, however.  Should that config be encoded as a file in desimodel itself?  Perhaps desimodel should have a load_specsim_config() function that returns a properly configured specsim Configuration object for the parameters in that version of desimodel?

i.e. can we find a way to load a fully correct desi configuration directly, without having to load it and then modify it to get it into the actually correct state to use?

dkirkby:
In this PR, I am aiming for the minimal changes required by the new specsim API, while preserving identical output results.  In that spirit, I did not change how quickbrick sets its limits (which are different from those used by quickgen) even though it doesn't make any practical difference as long as the simulation limits include the throughput limits.  However, it does make a tiny numerical difference since changing the grid min adds a sub-pixel offset to the grid, so I wanted to avoid that complication for my testing.

Looking ahead, it would be good to pick a canonical simulation range used by both quickgen and quickbrick, and update the specsim [desi.yaml](https://github.com/desihub/specsim/blob/speclite/specsim/data/config/desi.yaml) if necessary.  Currently, it contains:
```
wavelength_grid:
    unit: Angstrom
    min: 3500.3
    max: 9999.7
    step: 0.1
```
While the throughput limits (which specsim also knows about from the throughput files) are:
```
>>> desimodel.io.load_throughput('b').wavemin, desimodel.io.load_throughput('z').wavemax
(3533.0, 9913.0000000092878)
```
Note that `quickbrick` originally used:
```
wavemin = desimodel.io.load_throughput('b').wavemin
wavemax = desimodel.io.load_throughput('z').wavemax
wave = np.arange(wavemin-1, wavemax+1, opts.wavestep)
```
but I eliminated the +/-1 padding in this PR since it makes no difference to the results.

dkirkby:
I just tagged and released [specsim v0.3](http://specsim.readthedocs.org/en/latest/) so this PR is ready to merge as far as I am concerned.

sbailey:
Looks fine to me.  Refactoring the wavelength grid provenance can be a future PR.  Merge away!

dkirkby:
desihub/desimodel#7 is relevant to the question of what the canonical grid should actually be.

PR #88, opened 2016-02-26
fix mag range of simulated stdstars

sbailey:
The default magnitude range for desisim.templates.STAR is not good for simulating standard stars, even when FSTD=True (that gets the colors right, but not the magnitudes).  This PR adds 16 < rmag < 19 constraints in desisim.targets.get_targets() when it generates stars to use as standard stars.

I think #64 will further clean this up by creating a separate FSTD class that could have appropriate magnitude ranges for FSTD.  This is an interim fix using the current multi-purpose STAR class.

### DIFF
--- a/py/desisim/targets.py
+++ b/py/desisim/targets.py
@@ -200,7 +200,10 @@ def get_targets(nspec, flavor, tileid=None):
         elif objtype == 'STD':
             from desisim.templates import STAR
             star = STAR(wave=wave,FSTD=True)
-            simflux, wave1, meta = star.make_templates(nmodel=nobj)
+            rr = (16.0, 19.0)
+            gg = (16.0, 19.5)
+            simflux, wave1, meta = \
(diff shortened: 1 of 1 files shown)


moustakas:
Can you make the same change to desisim/bin/simulate-templates too please?

sbailey:
It appears taht bin/simulate-templates is already correct, since it has separate options for fstd vs. other objtype mag ranges:
```
    fstd_parser.add_argument('--rmagrange-fstd', type=float, default=(16.0,19.0), nargs=2, metavar='',
                            help='Minimum and maximum r-band (AB) magnitude range')
...
            if FSTD: # simulate an F-standard
                flux, wave, meta = star.make_templates(nmodel=args.nmodel,
                                                       vrad_meansig=args.vrad_fstd,
                                                       rmagrange=args.rmagrange_fstd,
                                                       seed=args.seed)
```

moustakas:
Right, OK, that makes sense. As you say, having a new FSTD template class will resolve this.  Thanks.

Issue #71, opened 2016-01-30
newexp-desi --flavor bgs and bright


sbailey:
Once #70 (add BGS-like templates) is done, add `newexp-desi --flavor bgs` and `bright` options.

`--flavor bgs` would simulate just Bright Galaxy Survey targets, while `bright` would do a mix of BGS and MWS targets based upon the target densities in $DESIMODEL/data/targets/targets.dat .

Use `exptime_bright` from desimodel.io.load_params() for exposure time for flavors `mws`, `bgs`, and `bright`.

moustakas:
I note that that file (targets.dat) doesn't contain any information about MWS, and the BGS numbers are woefully out of date / incorrect.  

A related question: will MWS be applying color cuts to select particular subpopulations?

sbailey:
Placeholder numbers for MWS, BGS have been added to desimodel and PR #92 fixed a bug for newexp-desi --flavor bright.  Closing ticket.

PR #92, opened 2016-03-04
fix BRIGHT vs bright flavor bug

sbailey:
Issue #71 to add newexp-desi --flavor "bright", "bgs", and "mws" was already implemented, albeit with a bug in "BRIGHT" vs. "bright".  This PR fixes that so that flavors are case-insensitive.


moustakas:
Looks fine.

Issue #89, opened 2016-03-02
Streamline specsim-quickgen interaction


dkirkby:
Quickgen's use of specsim is quite convoluted and exposes a lot of specsim implementation details because specsim does not provide convenient access to the information it needs (see for example desihub/specsim#17 and desihub/specsim#18).  This issue is to review this interaction and make changes to specsim and quickgen to reduce the coupling and generally make quickgen more maintainable.

dkirkby:
There are four types of information flowing from specsim to quickgen:
1. Wavelength grids for the simulation and each camera.
2. Mean number of detected electrons from the source and sky and the corresponding ivars in downsampled bins for each camera.
3. Perfectly calibrated mean flux in downsampled bins, with resolution applied for each camera.
4. Sparse resolution matrix for each camera.

Quickgen also adds Gaussian noise to the specim-calculated means, but I would rather do this in specsim (e.g., for consistency between quickgen and quickbrick).  See desihub/specsim#3.

dkirkby:
On the issue of wavelength grids, we had some discussion in #86 and desihub/desimodel#7 is very relevant.

The coverage of each camera is implicitly specified by two arrays, throughput and resolution, and explicitly by header keywords `WMIN_ALL` and `WMAX_ALL` in `desimodel/data/specpsf/psf-quicksim.fits`.

Specsim works with two different ranges for each camera: the physical extent of active pixels (`camera.ccd_slice`) and the slightly larger range of source wavelengths that can disperse into the CCD (`camera.response_slice`), where resolution tails are truncated at n-sigma (nominally 5-sigma, set by the `instrument.cameras.*.constants.num_sigmas_clip` config parameters).

Here are the numbers, for reference:
```
  Range     b(min)   b(max)   r(min)   r(max)   z(min)   z(max) 
           Angstrom Angstrom Angstrom Angstrom Angstrom Angstrom
---------- -------- -------- -------- -------- -------- --------
Throughput 3533.000 5998.000 5564.000 7805.000 7360.000 9913.000
    Header 3569.000 5949.000 5625.000 7741.000 7435.000 9834.000
       CCD 3569.000 5949.000 5625.000 7741.000 7435.000 9834.000
  Response 3565.400 5952.600 5621.700 7744.300 7431.000 9838.000
```
[...]

PR #93, opened 2016-03-06
Streamline use of specsim by quickgen and quickbrick

dkirkby:
The changes here depend on specsim changes in desihub/specsim#35, which will be merged and tagged (as v0.4) before this PR is merged.

Both quickgen and quickbrick now have much cleaner and consistent interactions with specsim.

Externally visible changes to quickgen:
* Add `--seed` option for reproducible results with different seeds (was always seed=0 before).
* Same noise realization is used for frame and cframe files.
* Noise and ivar are based on individual camera statistics, not the coadd, which is most noticeable at the camera edges.
* Resolution is convolved with the output pixel size.
* Sky model output now has non-zero ivar and noise (after fixing desihub/specsim/#1).

Externally visible changes to quickbrick:
* Random numbers are now reproducible.
* Remove `--wavestep` option (was 0.2A by default, now 0.1A from `desi.yaml`)
* Remove `--downsampling` option (output pixels were 3 x 0.2A by default, now 0.5A from `desi.yaml`)
* Resolution matrix now written correctly to brick files (was using non-standard number of off-diagonals).

Externally visible changes to specsim that affect both scripts:
* The sky now has extinction applied.
* Source flux can now disperse onto the CCDs from just outside.
* Initializing a simulator no longer warns about non-standard FITS units.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# Binary output files.
+*.fits
--- a/bin/quickbrick
+++ b/bin/quickbrick
@@ -36,7 +36,6 @@ parser.add_option("-o", "--outdir", type=str,  help="output data directory; defa
 parser.add_option("-a", "--airmass", type=float,  help="airmass [%default]", default=1.25)
 parser.add_option("-t", "--testwavelength", type=float,  help="test quickbrick with a single emission line", default=None)
 parser.add_option("-s", "--seed", type=int,  help="random seed", default=None)
-parser.add_option("--wavestep", type=float,  help="initial sampling of templates", default=0.2)
 parser.add_option("--outdir_truth", type=str,  help="optional alternative outdir for truth files")
 parser.add_option("-z","--zrange", type=str, help="redshift range of the form zmin:zmax (applies only to ELG,LRG,QSO)",default=None)
 
--- a/bin/quickbrick
+++ b/bin/quickbrick
@@ -32,6 +32,7 @@ parser.add_option("-b", "--brickname", type=str,  help="brickname")
 parser.add_option(      "--objtype", type=str,  help="elg,lrg,qso,star or mix")
 parser.add_option("-n", "--nspec", type=int,  help="number of spectra to simulate")
 parser.add_option("-o", "--outdir", type=str,  help="output data directory; default .", default='.')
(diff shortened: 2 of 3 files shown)


dkirkby:
Fixes #89

sbailey:
These changes all look good to me in both desisim and specsim, but it appears that our desisim travis tests pass without even installing specsim, so none of the quickbrick/quickgen changes are checked at all by the tests unless I'm missing something (i.e. my reading coverage of our travis config may be incomplete).

One question: are quickbrick/quickgen dependent upon desi.yaml in specsim, or is there a way to override which config file they use?  In the long term we should re-consolidate on desimodel as being the only canonical description of the desi configuration (i.e. desimodel should include whatever config file format specsim needs); in the short-term it would be handy to provide an option to override which config file to use.

dkirkby:
@sbailey This PR only updates the use of specsim by the quickgen and quickbrick scripts.  I didn't touch the testing setup, but I will create a new desisim issue for this.

Yes, these scripts are dependent on `desi.yaml` in specsim but that hasn't changed with this PR.  They can read this file from anywhere, so it would be straightforward to have them use a version under `desimodel`.  I will create a desimodel issue for this.

>  in the short-term it would be handy to provide an option to override which config file to use.

Good idea.  I will add that now.

dkirkby:
I added an option `--config` to quickgen and quickbrick that defaults to `desi` (i.e., read `desi.yaml` from the data/config directory of the currently installed version of specsim), but also allows arbitrary yaml files to be used, e.g.
```
quickgen --config mydesi.yaml ...
```
A related feature that could be useful is to write a yaml output file that can be used to exactly reproduce a simulation job, i.e., including changes made at the command line (airmass, exposure time, ...) and using checksums to capture the versions of all external files (throughput, ...).  I don't think we need robust provenance tracking and reproducibility right now, but its good to keep this functionality in mind so we don't make it unnecessarily hard to implement down the road.

sbailey:
OK to merge this whenever the specsim merge and tag is ready.  Would be nice to include that in spectro pipeline + sims tags this week (maybe even today? tomorrow?).

dkirkby:
Ok, I will try to finish this up today.

dkirkby:
I just released specsim v0.4 and will now merge this branch.

PR #94, opened 2016-03-07
refactor stellar templates, issue #64

crockosi:
Refactor STAR template class to make it easy to create multiple kinds of DESI stellar targets without proliferating arguments and if statements. There are now FSTD and MWS_STAR classes that inherit from the STAR class.  This is done "new class" style, I can change to the older style if preferred. Corresponding changes to target.py and io.py to use the new stellar template classes in newexp-desi for the mws, bgs and bright options.  The only remaining problem for bright sims is that spectre understands none of objtype=BGS, MWS_STAR or FSTD and uses the default for the throughput for all three.  That can only be right for the extended source or the stars, not both.

### DIFF
--- a/bin/newexp-desi
+++ b/bin/newexp-desi
@@ -7,7 +7,7 @@ newexp-desi arc/flat/dark/bright/bgs/mws/elg/lrg/qso [options]
 
 dark: a mix of elg, qso and lrg targets
 bright: a mix of BGS and MWS targetls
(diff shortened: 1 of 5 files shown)


sbailey:
The specter default is for a star, so MWS_STAR and FSTD are in luck, but BGS is not.  The short term impact is that specter will use an overly optimistic geometric fiber input loss for BGS, thus giving them too many photons.  We should clean that up, but it isn't a blocking factor for this PR itself.

moustakas:
This looks good to me.

PR #98, opened 2016-03-08
Cleanup

sbailey:
Some cleanup features and fixes:
* adds gray/grey flavor to simulate ELGs with gray sky conditions
* fixes pixsim wavemin/wavemax options
* converts DESI-specific objtypes into specter objtypes before calculating throughput
  * e.g. STAR, STD_FSTAR, and MWS_STAR all map onto "STAR" for specter
* adds get_targets_parallel() for faster template generation
* converts some print statements into logger calls

### DIFF
--- a/bin/newexp-desi
+++ b/bin/newexp-desi
@@ -19,7 +19,6 @@ import numpy as np
 import optparse
 import random
 import time
-import multiprocessing
 import string
 
 from desisim import obs, io
--- a/bin/newexp-desi
+++ b/bin/newexp-desi
@@ -32,7 +32,7 @@ parser = optparse.OptionParser(
     epilog = "See $SPECTER_DIR/doc/datamodel.md for input format details"
     )
         
-parser.add_option("--flavor",   type=str,   help="arc/flat/dark/bright/bgs/mws/elg/lrg/qso", default='dark')
+parser.add_option("--flavor",   type=str,   help="arc/flat/dark/gray/bright/bgs/mws/elg/lrg/qso", default='dark')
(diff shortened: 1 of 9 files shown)


dkirkby:
It looks like `obs.new_exposure()` is still reading the static dark/grey/bright sky spectra from `desimodel`.  Are you planning to use the new grey/bright sky model calculations provided by specsim?  How is the sky flux read by `obs.new_exposure()` actually used?

sbailey:
The plan is to eventually use the specsim sky calculation for the pixel-level simulations, but realistically I'm not going to be able to do that myself this week prior to tagging.  Using the pre-calculated dark/grey/bright sky vectors from desimodel is the triage fallback plan.  If someone can get me a pull request on Wednesday it could get included (it should come with what default parameters to use if flavor = dark/grey/bright, even if it offers more flexible choices).

obs.new_exposure() produces the simspec files which include both the object flux and the photons as observed on the CCD.  These are calculated together so that they can be written to the same file as the truth that should be reproduced (within noise) by the downstream pipeline steps.  Calculating the photons on the CCD requires knowing the sky flux.  The object flux and the sky flux have different throughputs applied to them since they have different fiber input losses, but both are needed to be able to write out the full simspec file (see datamodel at http://desidatamodel.readthedocs.org/en/latest/DESI_SPECTRO_SIM/PIXPROD/NIGHT/simspec-EXPID.html, admittedly out of date, but the key point is that it contains the object FLUX, the sky SKYFLUX, and the PHOT and SKYPHOT for each of the individual channels).

PR #99, opened 2016-03-09
Batch pixsim

sbailey:
This PR adds two utility functions for generating batch scripts for performing pixel-level simulations.  Eventually these will likely be replaced with functions based upon the desisim.pipeline automation, but that is still under active development and this is a placeholder to enable faster sims now.

Smaller items that also came in here:
* OBJTYPE = ARC / FLAT set in the fibermap for arc and flat exposures
* Better propagation of header keywords to get EXPID and SPECTRO in the pix file header.

This branch also merged changes in py/desisim/templates.py from master; these appear in github list of changes but those are already in master.

### DIFF
--- a/bin/newexp-desi
+++ b/bin/newexp-desi
@@ -19,7 +19,6 @@ import numpy as np
 import optparse
 import random
 import time
-import multiprocessing
 import string
 
 from desisim import obs, io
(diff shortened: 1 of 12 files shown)


sbailey:
Coverage tests still "pending", but coveralls.io has been down recently.  Proceeding with merging.

PR #108, opened 2016-04-08
fix random seed usage with parallelism

sbailey:
This fixes a critical bug in the random number seed propagation in get_targets_parallel() that was causing different spectra to receive the same TARGETID.  Tests have been added to verify that different seeds result in different TARGETIDs and different spectra, and to verify that the same seed results in the same TARGETIDs and identical spectra.

Assigning to @tskisner, though if @moustakas is in front of a computer right now I'd appreciate a review from him as well since he faced similar issues getting the random number seed propagation correct in the templates code.

### DIFF
--- a/bin/newexp-desi
+++ b/bin/newexp-desi
@@ -38,8 +38,7 @@ parser.add_option("--exptime",  type=int,   help="exposure time in seconds, defa
 parser.add_option("--night",    type=str,   help="YEARMMDD")
 parser.add_option("--nspec",    type=int,   help="Number of spectra to simulate [%default]", default=5000)
 parser.add_option("--airmass",  type=float, help="Airmass [%default]", default=1.0)
-parser.add_option("--randseed", type=int,   help="random number seed"
-                  )
+parser.add_option("--randseed", type=int,   help="random number seed")
 
 opts, args = parser.parse_args()
 
@@ -68,7 +67,7 @@ if opts.flavor not in ['arc','flat','dark','gray','grey','bright','bgs','mws','l
     log.fatal('ERROR: unknown flavor {}'.format(opts.flavor))
     sys.exit(1)
 
-    
+
 #- TODO: don't hardcode these default exposure times
 if opts.exptime is None:
     if opts.flavor == 'arc':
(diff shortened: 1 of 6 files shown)


py/desisim/targets.py
```diff
@@ -128,29 +128,39 @@ def sample_objtype(nobj, flavor):
...
     import multiprocessing as mp
     if nproc is None:
         nproc = mp.cpu_count() // 2
-    
+
     #- don't bother with parallelism if there aren't many targets
     if nspec < 20:
         log.debug('Not Parallelizing get_targets for only {} targets'.format(nspec))
+        np.random.seed(randseed)
         return get_targets(nspec, flavor, tileid)
```
moustakas: I think you still need to pass randseed to get_targets() here.  Then the naked call to np.random.seed(randseed) (previous line) can probably go away.  There should also be a test covering the case of nspec<20.

py/desisim/targets.py
```diff
@@ -128,29 +128,39 @@ def sample_objtype(nobj, flavor):
...
     else:
         nproc = min(nproc, nspec//10)        
         log.debug('Parallelizing get_targets using {} cores'.format(nproc))
         args = list()
         n = nspec // nproc
+        if randseed is not None:
+            #- use large multiplier since we don't want thread 2 of randseed=1
+            #- to produce the same results as thread 1 of randseed=2
+            irandseed = randseed*100000
```
moustakas: This seems a little fragile.  This is probably equally fragile, but in desisim.ELG.make_templates() [here](https://github.com/desihub/desisim/blob/master/py/desisim/templates.py#L501) I use the random seed to create a distribution of other random seeds. Could you do something similar here and create another set of seeds, one for each core?

bin/newexp-desi
```diff
@@ -38,8 +38,7 @@ parser.add_option("--exptime",  type=int,   help="exposure time in seconds, defa
 parser.add_option("--night",    type=str,   help="YEARMMDD")
 parser.add_option("--nspec",    type=int,   help="Number of spectra to simulate [%default]", default=5000)
 parser.add_option("--airmass",  type=float, help="Airmass [%default]", default=1.0)
-parser.add_option("--randseed", type=int,   help="random number seed"
-                  )
+parser.add_option("--randseed", type=int,   help="random number seed")
```
moustakas: Very minor: Do we want to standardize this variable across scripts to "seed" (as used by the templates) rather than "randseed."

sbailey:
John made good suggestions while I was chatting on the phone with Ted about this PR.  I will make updates along those lines momentarily.

PR #111, opened 2016-04-13
Quicksurvey

forero:
First implementation of quicksurvey class and scripts. Runs a DESI simulation starting from a truth file, target file and epoch files. 

It creates a Merged Target List, runs fiber assignment and creates a redshift catalog for each epoch.

The class also includes a function to get rough efficiency statistics.


forero:
@sbailey This version finally includes all the changes we discussed a week ago.

sbailey:
Apologies for the delay in reviewing; merging now.

PR #112, opened 2016-04-26
Redshift truth

profxj:
Adds a desi script that runs on a given night to perform QA on
redshifts using the 'truth' in simpix files.

Includes a summary figure based on the ones used in z1dc.

Tested on the dogwood productiong, night 20150211

No unit tests.


sbailey:
Thanks for moving this over to desisim.  I added a few small bricks of zbest info on NERSC at /project/projectdirs/desi/spectro/redux/dc3c/bricks for testing and committed a few changes to work with the latest reductions:
  * HDUs should be referenced by name, not number (their order is arbitrary)
  * Catches case of all zbest files are missing
  * Increased logging (some debug, some warn, some fatal)
  * Skip calibration exposures that don't have METADATA table
  * O2FLUX has been renamed OIIFLUX (it is possible that some logic is now extraneous; I didn't try to understand the ptype vs mtag logic that was dealing with O2FLUX vs. OIIFLUX)

Requests before merging:
  * Please get the !!python/unicode stuff out of the yaml
  * Should we standardize on QA scripts having "qa" in their name?  Maybe `desi_qa_zfind`?
  * --qafig option generates a crash; see below.  This may be due to many bricks missing their zbest files, but we should still handle missing redshifts gracefully.

Other cleanup requests for now or later, but don't let this hold up fixing the others prior to a merge:
  * desisim.spec_qa.redshifts has bare print statements that should be log messages; ideally the desi_night_zfind --verbose option would also propagate into those logs for what level is printed
  * if there are bare print statements, they should be motivated by human-readable summary tables, e.g.
efficiency, purity, catastrophic failure rate per target class instead of just "OBJ=QSO_T failed tests CAT_RATE-EFF"

The crash:
```
desi_night_ztruth --specprod /project/projectdirs/desi/spectro/redux/dc3c --night 20160408 --verbose --qafile blat.yaml --qafig blat.pdf
...
  File "/global/homes/s/sjbailey/desi/git/desisim/bin/desi_night_ztruth", line 101, in <module>
    main()
  File "/global/homes/s/sjbailey/desi/git/desisim/bin/desi_night_ztruth", line 92, in main
    dsqa_z.dz_summ(simz_tab, pp=pp)
  File "/global/homes/s/sjbailey/desi/git/desisim/py/desisim/spec_qa/redshifts.py", line 825, in dz_summ
    num_slices=nslice, y_cut=max_dv, axis=axis, min_count=min_count)
  File "/global/homes/s/sjbailey/desi/git/desisim/py/desisim/spec_qa/redshifts.py", line 733, in plot_slices
    weights=weights, color='k', cumulative=True)
  File "/global/common/edison/contrib/desi/edison/hpcports_gnu-8.2/matplotlib-1.4.3_50fa2ed0-8.2/lib/python2.7/site-packages/matplotlib-1.4.3-py2.7-linux-x86_64.egg/matplotlib/axes/_axes.py", line 5678, in hist
    m, bins = np.histogram(x[i], bins, weights=w[i], **hist_kwargs)
  File "/global/common/edison/contrib/desi/edison/hpcports_gnu-8.2/numpy-1.10.4_908fcf2d-8.2/lib/python2.7/site-packages/numpy/lib/function_base.py", line 176, in histogram
    'max must be larger than min in range parameter.')
AttributeError: max must be larger than min in range parameter.
```

profxj:
Corrected small bug making the test with dc3c fail.
Removed print statements (or changed to log)
Removed unicode from yaml files :: Requires desiutil PR#31
Have renamed the script  desi_qa_zfind

sbailey:
desiutil 1.4.0 includes desiutil.io.yamlify used here.  I updated the travis tests to use that version even though the current tests don't cover this.

The yaml output happens in two steps, but it was writing invalid output:
```
{PRODNAME: dc3c, SIMSPECV: '9.999'}
ELG:
  CAT_RATE: 0.45565749235474007
  EFF: 0.41590214067278286
...
```
I included `default_flow_style=False` in the first yaml dump to get it to output valid yaml when combined with the next dump:
```
PRODNAME: dc3c
SIMSPECV: '9.999'
ELG:
  CAT_RATE: 0.45565749235474007
  EFF: 0.41590214067278286
...
```

We'll want to add tests to catch this kind of stuff, but that is a bit tedious to setup all the inputs.  Let's merge for now and do additional cleanup, tests, features in a later PR.

Issue #101, opened 2016-03-17
quickbrick needs to know about bright vs dark survey


moustakas:
```quickbrick``` is barfing because desimodel/data/desi.yaml specifies ```exptime_dark``` and ```exptime_bright``` whereas ```quickbrick``` only expects ```exptime```:
  https://github.com/desihub/desisim/blob/master/bin/quickbrick#L59

I can fix this by adding a bright vs dark optional input to ```quickbrick```, which would also be an opportunity to include default bright-sky conditions to the simulated output spectra.

**Comment by sbailey** [comment]
`desimodel.io.load_desiparams` has this temporary hack while transitioning codes to use `exptime_dark` and `exptime_bright`:
```
    #- for temporary backwards compability after 'exptime' -> 'exptime_dark'
    if ('exptime' not in _params) and ('exptime_dark' in _params):
        _params['exptime'] = _params['exptime_dark']
```
i.e. quickbrick should work if using the latest desimodel code in addition to the latest desimodel data.

None-the-less, +1 for updating quickbrick to know about bright vs. dark and grabbing the appropriate exposure times, templates, and sky conditions.

**Comment by dkirkby** [comment]
Defining the canonical grey and bright observing sky conditions sounds like a great idea but should be a separate issue (labeled 'enhancement' rather than 'bug').

The root cause of this issue is that we do not have a robust way to synchronize version updates between packages.  In this case, incompatible versions resulted in a crash, which is much better than the alternative of an undetected incorrect result.

The first thing I would try is to define `install_requires` in `setup.py`:
```
install_requires=[
    'desimodel >= 0.4.2' 
]
```
See [here](http://peak.telecommunity.com/DevCenter/setuptools#declaring-dependencies) for details.  There are other package dependencies that should also be listed here if this approach works.

**Comment by sbailey** [comment]
Should this be tracked in setup.py, requirements.txt, both, or have setup.py parse requirements.txt ?  I'm not sure what the recommended practices are here.  If we can chose, I suspect that we'll have better luck with developers keeping a separate requirements.txt file up-to-date rather than setup.py itself.  Let's try to avoid having to keep both in sync.

I see that our current setup.py has:
```
setup_keywords['requires'] = ['Python (>2.7.0)']
# setup_keywords['install_requires'] = ['Python (>2.7.0)']
```
Does Ben or someone else know why the 'install_requires' line is commented out, vs. the 'requires' line?

dkirkby:
[...]

Issue #100, opened 2016-03-14
add the option of generating type Ia supernovae spectra


moustakas:
Work with Or Graur and the Time-Domain WG to add supernovae spectra (starting with Type Ia's) to galaxy (BGS, LRG, ELG) and---eventually---QSO spectra.

moustakas:
Added in #102.

Issue #72, opened 2016-01-30
generate canonical brighttime sims


sbailey:
Once #71 is done to enable `newexp-desi --flavor bright`, generate a canonical set of bright time simulations and advertise those for use in future pipeline development (e.g. data challenges).

While we're at it, do the same for an updated set of dark time exposures too.

dkirkby:
Will these bright time simulations use the new scattered moon model in specsim (desihub/specsim#9) ?

moustakas:
Yes, using the lunar model is critical for the BGS and MWS simulations.

Issue #96, opened 2016-03-07
Add travis tests of quickgen and quickbrick


dkirkby:
These scripts are currently not tested by travis or included in coverage reports.

I believe the main obstacle is that any test would need to read the following files that currently live in the `desimodel` svn package:
```
Loaded 65000 rows from .../desimodel/data/spectra/spec-sky.dat with args {'format': 'ascii'}
Loaded 65000 rows from .../desimodel/data/spectra/spec-sky-bright.dat with args {'format': 'ascii'}
Loaded 65000 rows from .../desimodel/data/spectra/spec-sky-grey.dat with args {'format': 'ascii'}
Loaded 65000 rows from .../desimodel/data/spectra/ZenithExtinction-KPNO.dat with args {'format': 'ascii'}
Loaded 667 rows from .../desimodel/data/sky/solarspec.txt with args {'format': 'ascii.basic'}
Loaded 4233 rows from .../desimodel/data/specpsf/psf-quicksim.fits with args {'hdu': 'QUICKSIM-R'}
Loaded 22411 rows from .../desimodel/data/throughput/thru-r.fits with args {'hdu': 'THROUGHPUT'}
Loaded 4761 rows from .../desimodel/data/specpsf/psf-quicksim.fits with args {'hdu': 'QUICKSIM-B'}
Loaded 24651 rows from .../desimodel/data/throughput/thru-b.fits with args {'hdu': 'THROUGHPUT'}
Loaded 4799 rows from .../desimodel/data/specpsf/psf-quicksim.fits with args {'hdu': 'QUICKSIM-Z'}
Loaded 25531 rows from .../desimodel/data/throughput/thru-z.fits with args {'hdu': 'THROUGHPUT'}
Loaded 27 rows from .../desimodel/data/throughput/fiberloss-perfect.dat with args {'format': 'ascii'}
Loaded 27 rows from .../desimodel/data/throughput/fiberloss-star.dat with args {'format': 'ascii'}
Loaded 27 rows from .../desimodel/data/throughput/fiberloss-lrg.dat with args {'format': 'ascii'}
Loaded 27 rows from .../desimodel/data/throughput/fiberloss-qso.dat with args {'format': 'ascii'}
Loaded 27 rows from .../desimodel/data/throughput/fiberloss-sky.dat with args {'format': 'ascii'}
Loaded 27 rows from .../desimodel/data/throughput/fiberloss-elg.dat with args {'format': 'ascii'}
[...]

PR #114, opened 2016-05-02
clean up of quickbrick to support bright-time survey

moustakas:
I'm opening this PR now so I can get  can get eyes on the code sooner rather than later. Algorithmically I haven't changed anything significant in ```quickbrick```, but I have done a thorough "cleaning" of the code (PEP8 style mostly throughout, streamlining of variables and repeat code, some error handling, refactoring into parse() and main() functions, additional optional inputs such as the lunar phase parameters, etc.).  The biggest change is that ```quickbrick``` can now be used to simulate bright-time objects via the BGS, MWS, or BRIGHT_MIX objtype optional input.

This PR should close the bug identified in #101 and should move us significantly closer toward generating the bright-time simulations we need (#72).  

Some outstanding issues I would still like to fix/add:
* Figure out the verbose logging (help on this would be great -- log.info prints output even when the debugging level isn't set to INFO)
* Add additional error checking
* Optionally output spectra with supernova spectra superposed (see #100)


moustakas:
What should we do about testing high-level scripts like ```quickbrick```?  Although at its heart ```quickbrick``` is a wrapper script, there's plenty of stuff that could "go wrong" as the code is developed and changed (as in this PR).  We almost certainly want to know that the code is writing out consistent output, given identical input.  I'd like the answer to this question to not be part of this PR necessarily (to not bog it down), but I think that we do need to construct some set of tests, or at least refactor the code so its components can be tested.

moustakas:
I believe this is in good enough shape that it's ready to merge.

moustakas:
I see that my previous comment was discussed in #96.  I'll work on resolving this issue as part of this PR.

sbailey:
Merging now; let's add tests in a separate PR.

PR #116, opened 2016-05-03
Wave in vacuum

julienguy:
This PR for using an arc lamp emission line spectrum with wavelength in vacuum.
The template is at NERSC in /project/projectdirs/desi/spectro/templates/calib/v0.3
The hardcoded path in py/desisim/obs.py has been modified, we will need to remove this hardcoded path  in the future (the change is made in branch teststand).

This change from air to vacuum wavelength is needed to be consistent with the master version of desispec and specex, and the definitions of wavelength in the DESI target templates.

### DIFF
--- a/etc/travis_env_common.sh
+++ b/etc/travis_env_common.sh
@@ -54,7 +54,7 @@ $PIP_INSTALL speclite
 
 # DESI_ROOT and DESI_BASIS_TEMPLATES with test data
 export DESISIM=$PWD
-testdata_version=0.2
+testdata_version=0.3
 cd ./data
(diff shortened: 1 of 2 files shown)


sbailey:
Travis tests are failing because the lightweight test data in desisim-testdata also needs to be updated for new version.  Keeping the lightweight test data in sync with the full templates is a bit of a pain.

julienguy:
I am merging this PR now that the travis tests succeeded after I updated desisim-testdata to v0.3 and changed accordingly   etc/travis_env_common.sh

Issue #121, opened 2016-05-08
bug in pixel level simulations : unmasked dead column


julienguy:
A dead column with negative pixel values but positive inverse variance found in one simulation of z-band image. This is probably introduced when pasting cosmic rays in images. Need to check cosmics image library. This causes failures of desi_bootcalib.

![capture du 2016-05-08 14 35 29](https://cloud.githubusercontent.com/assets/5192160/15097845/4c962526-152a-11e6-8771-10da27f1a5dc.png)

julienguy:
Fixed in branch fix_cosmicsims (will submit a PR once test at NERSC ended).

PR #122, opened 2016-05-09
Fix cosmicsims

julienguy:
This PR fixes issue #121 (checked with a simulation run at NERSC)


julienguy:
errors on travis test do not seem related to this PR. I let the travis experts investigate, eventhough we need this fix for the pixsim runs.

sbailey:
Our svn server was down when the first set of tests ran, resulting in not being able to install desimodel and thus tests going up in flames.  I triggered Travis to rerun them and then ran without failure.

Longer term, it will be better for us to propagate the dark image bad pixel mask rather than setting the bad pixel masked values to 0, but this is a pragmatic step in the meantime.  Note: we should propagate the bad pixel mask but *not* the cosmics mask from the dark images, since those cosmics were discovered too easily by not having any signal to confuse the cosmics identification algorithm.  We want to honestly re-discover the cosmics after combining them with signal.

Issue #66, opened 2016-01-26
path/environment variables issues with pixel simulations on laptop


julienguy:
It's difficult to run pixel level simulations on a laptop, because there are too many environment variables to set, some hard-coded path, and some inconsistencies in path , and  too many python packages to have this running. 

* packages (and I might be missing some )
```
desiutil,desispec,specter,specsim,desisim,desimodel(git),desimodel(svn)
```

*  One needs 5 environment variables to get 1 sim running , we should deflate this 
```
export PIXPROD=0
export PRODNAME=0
export DESI_SPECTRO_SIM=/home/guy/Projets/DESI/analysis/pixelsims
export DESI_ROOT=/home/guy/Projets/DESI/analysis/pixelsims
export DESI_SPECTRO_DATA=/home/guy/Projets/DESI/analysis/pixelsims
```
There is a semi hard-coded path in desisim/obs.py
```
infile = os.getenv('DESI_ROOT')+'/spectro/templates/calib/v0.2/arc-lines-average.fits
```

After successfully running 
```
newexp-desi --flavor=arc --expid=0 --nspec=25
newexp-desi --flavor=flat --expid=1 --nspec=25
```
I got another path problem when running
```
pixsim-desi --night=20160126 --expid=0 --cameras=b0 --nspec=25 --ncpu=12
pixsim-desi --night=20160126 --expid=1 --cameras=b0 --nspec=25 --ncpu=12
```
in desispec/io/fibermap.py 
```
IOError: cannot open/home/guy/Projets/DESI/analysis/pixelsims/20160126/fibermap-00000000.fits
```
The only fix I found is with a symbolic link 
```
ln -fs 0/20160126 20160126
```

weaverba137:
Leaving aside the number of environment variables for now, you *definitely* don't want to set `PIXPROD` or `PRODNAME` to values that might evaluate as `False`.  Those should be strings.

sbailey:
Even though I'm the original author of pixsim-desi, I agree this is a pain.

I suggest we keep the BOSS-like system of environment variables to set default path locations, while augmenting that with command line options to override output locations if you want to run a simulation outside of the data model directory structure ecosystem.

The need for a symbolic link smells like an outright bug, but it is surprising to me that it hasn't popped up before.  I agree with Ben's point about avoiding environment variable names that might eval as False rather than a string, but I'm not sure that is the underlying problem here.

weaverba137:
[...]

PR #131, opened 2016-05-26
Rawpixsim

sbailey:
This PR refactors the pixel-level simulations:
  * simulate raw desi data including amplifier offsets and overscan regions, then call desispec.preproc code to preprocess that into pix*.fits preprocessed files
  * fixes a bug in the target RA,dec values (parallel processes were all generating the same RA,dec)
  * Somewhat improves the usability issues raised in #66: you can now give direct inputs/outputs without relying upon environment variables
  * more tests (and thus slower testing too)
  * refactored code flow to separate algorithms from I/O
  * simpix\*.fits files with the noise-free simulated truth now mirror the raw desi\*.fits files with one file per exposure with multiple HDUs; the preprocessed pix*.fits files remain one per camera per exposure.

This is getting to be a rather large PR that could be hard to review, but I don't want it to diverge too much more before merging.

Example outputs are on edison scratch in /scratch1/scratchdirs/sjbailey/desi/spectro/sim/rawpixsim/20160525/ .

This is ready for review, but not necessarily for merging yet.  Remaining issues:
  * I may have broken Julien's teststand examples with only a few instrumented fibers; awaiting feedback
  * Travis tests run and report success and then hang when sending results to coveralls.  I'm not sure if this is a configuration problem or bad luck with coveralls.io uptime.
  * runtime is right on the edge for finishing within the debug queue 30 minute time limit.  Fluctuations in python startup time can mean that the simulations don't finish before the job is killed.


sbailey:
Apologies for all the github spam today as I've tried to debug the travis tests.  The problem is isolated to `desisim.test.test_pixsim.test_main()`, which calls `desisim.pixsim.main()` twice with different parameters.  Either of them individually works, but collectively they cause the coverage report to hang while writing (i.e. after the tests have successfully finished, it hangs while writing the coverage results, but prior to trying to send them to coveralls.io).

sbailey:
The coverage output problem persists even if the calls to `pixsim.main()` are isolated into separate tests that cleanup after themselves before the next one runs.  I don't have any problems running `python setup.py test --coverage` on my laptop, but this is smelling like a coverage tool bug with whatever version is running on Travis.

julienguy:
This is great (except for one issue for teststand simulation).

- Raw data and preprocessed data in /scratch1/scratchdirs/sjbailey/desi/spectro/sim/ look nice

- I was able to test the code on edison (after having to do 'module unload desiutil' because python was bizarrely using the installed version instead of my local master copy despite setting correctly the PYTHONPATH).

- I checked 'pixsim-desi --preproc ...' and 'desi_preproc ...' return the same preprocessed image
- I checked rapidely the preproc.py algo, and the returned value of readnoise ... great.

Issues :
1) pixsim-desi doesn't simulate the correct fibers.
I ran as a test "newexp-desi --flavor=arc --expid=3 --night=20160330 --testslit"
which writes a fibermap file with :
h["FIBERMAP"].data["FIBER"] = array([   0,   25,   50,   75, ...]),
ie 200 fibers for the 10 spectrographs, 20 fibers per spectro, and one fiber per bundle per spectro.
But this version of pixsim-desi ignores the fibermap and writes 200 contiguous fibers on the spectrograph 0.

2) pixsim-desi duplicates HDUs in raw data image (eg desi-00000003.fits.fz) if file already exists. This results (luckily) in an error output from preproc.

Improvements :
- simulating raw images of gain of 1 is asking for trouble because we cannot test the gain correction, but there is arguably little room for bugs there.

sbailey:
Thanks for reviewing this.

> Issues :
> 1) pixsim-desi doesn't simulate the correct fibers.
> I ran as a test "newexp-desi --flavor=arc --expid=3 --night=20160330 --testslit"
> which writes a fibermap file with :
> h["FIBERMAP"].data["FIBER"] = array([ 0, 25, 50, 75, ...]),
> ie 200 fibers for the 10 spectrographs, 20 fibers per spectro, and one fiber per bundle per spectro.
> But this version of pixsim-desi ignores the fibermap and writes 200 contiguous fibers on the spectrograph 0.

If the simspec file isn't just fibers 0..n, then you now have to supply the `--fibermap` option to point it to the fibermap to use to know which spectra are there.  It doesn't try to be clever and lookup the fibermap location or assume that it is in the same directory as the simspec file.  For normal desi pixsim the fibermap isn't needed at all.  For teststand sims, let me know if you think is ok to have to explicitly list the fibermap, or if it should automatically find that for you, perhaps with the option just for overriding the default location.

> 2) pixsim-desi duplicates HDUs in raw data image (eg desi-00000003.fits.fz) if file already exists. This results (luckily) in an error output from preproc.

Good point.  I added a check in desispec.io.write_raw [desispec PR #201](https://github.com/desihub/desispec/pull/201) so that it won't output a duplicate camera, and then also added to this desisim branch to do a pre-flight check for duplicate cameras so that it won't waste time simulating a frame and then find out at the end that it can't write it.  The `--overwrite` option will overwrite the existing file regardless of what is in it.

> simulating raw images of gain of 1 is asking for trouble because we cannot test the gain correction, but there is arguably little room for bugs there.

If we want to simulate gain != 1 by default, we should update that in `desimodel/data/desi.yaml`.  It could also be added as an option here for testing, but I haven't implemented that yet.

weaverba137:
As @sbailey reported, Travis tests were hanging.  The culprit is the multiprocessing module used by `desisim.pixsim.parallel_project()`.  When I ran the tests of `desisim.scripts.pixsim.main()` with `--ncpu -1`, the tests did not hang.  Both standard tests and coverage tests are working when multiprocessing is disabled.

I know that we have encountered the problem of multiprocessing and unittest not getting along in the past, but I haven't been able to find the specific reference so far.

julienguy:
I checked my comments 1 and 2 are fixed. Ok for merging (but I let Stephen do it for possible interference with pipeline dev).

sbailey:
Unittests with multiprocessing on Travis are finally working again, by adding `pool.close()` and `pool.join()` after the multiprocessing `pool.map()`.  This doesn't appear to be necessary in normal operations, but something about Travis with coverage checking was causing it to finish the tests and report success but then hang before actually exiting.  I'll let final set of tests finish running and then merge.

Issue #104, opened 2016-04-08
include the option of simulating spectra at a set of given input redshifts


moustakas:
Requested feature from @sbailey:
```
Ability to pass in a redshift array and get back a set of templates at those redshifts.  This is needed for generating simulated data to match an input mock catalog.
```

moustakas:
I have this basically implemented, although there's a potential infinite-loop trap I'm not sure how to deal with.  It's possible -- and indeed likely -- that the user will pass a redshift which will *never* pass the color cuts for a given spectral class.  For example, *none* of the ~8000 ELG basis templates will ever pass the nominal grz color-cuts if they are told to be placed at z=0.1 (and indeed anything below z~0.5 is highly unlikely, by design).  

@sbailey, @ladosamushia How should we deal with this in the context of #110?

sbailey:
At minimum there should be a counter that stops after N tries to avoid an infinite loop.  Options after hitting that limit:
1. raise an exception
2. print a warning and then don't include that redshift in the output list (i.e. fewer outputs than inputs)
3. print a warning, grab another random redshift from the list and keep going, even if that means that it replicates that redshift.

I think I prefer (3) as default, with an option to do (1) instead.  (2) smells like trouble.

dkirkby:
Once the templates and cuts are fixed, you can precompute exactly which templates pass the cuts on a grid of redshifts and save the results in a boolean array `pass[ntemplates,nz]`.
This array is actually quick to compute using `speclite.filters` since you can do the magnitude calculations in parallel for all templates.  Once you have precomputed the `pass` array, you can simulate any given redshift `z ~ z[iz]` without any iteration by randomly selecting an `it` where `pass[it, iz]` is `True`.   You can also immediately detect an impossible redshift when `!np.any(pass[:,iz])` and raise an appropriate exception.

moustakas:
Thanks for the suggestion, @dkirkby.  In detail, what you describe is what I had originally implemented in the IDL (v1.0) version of the ELG templates.  However, those templates had static emission lines on them, whereas in the current template code each continuum template gets a unique emission-line spectrum (according to a physical model).  And this is why the templates are built in a ```for``` loop, which is not terribly slow but slightly annoying.

[...]

Issue #106, opened 2016-04-08
draw template redshifts and magnitudes from more realistic distributions


moustakas:
When generating synthetic spectral templates, the redshifts are drawn from a uniform distribution between minimum and maximum values defined by the user (with sensible default ranges set by the spectral class).  #104 is one alternative.  

Another option is to draw redshifts from the expected redshift distribution n(z) for each target class. (Note that desisim.targets.sample_nz already does this---it's just not yet implemented.)

Similarly, the spectra are normalized to an apparent magnitude distribution (in a band which depends on the target class) which is also (unrealistically) uniformly distributed over a user-defined range.  We should have the flexibility of drawing from an observationally constrained luminosity function, and then asking whether the object passes the faint-magnitude limit by computing the distance modulus and K-correction.

Note that code to draw from the QSO luminosity function has already been implemented [here](https://github.com/dkirkby/ArgonneLymanAlpha/blob/cd84062c112effbd0d036776fa782e991a6abd47/notebooks/SampleProperties.ipynb).

sbailey:
Drawing from n(z) would be handy, but we also need the option of making templates to match an exact input list of redshifts so that template making can be matched to an input mock catalog with known true redshifts.

moustakas:
Defining the input redshifts from mocks is #104.  But what else from the mocks should the templates inherit?  Do the mocks include colors, luminosities, line-strengths, etc.?  [We can take this back on-list if necessary.]

sbailey:
Our current mocks only have (ra, dec, redshift, objtype).  Future mocks may also have colors, etc. — Peder Norberg is particularly keen about that.  Let's get the 0th order n(z) and user-defined redshifts implemented, and spin it to mailing list discussions with the cosmo sims group about what else may arrive and when.

Apologies I hadn't seen #104 when I commented on this.  Perhaps this is obvious, but making sure: the implementations should be very closely related, i.e. this issue should sample n(z) to get an array of redshifts, and then call the exact same code as for a user-provided array of redshifts.

sbailey:
[...]

Issue #105, opened 2016-04-08
optionally include "contaminant" objects/template spectra


moustakas:
Feature request from @sbailey:
```
Contaminant templates: non-XYZ objects that pass the cuts for XYZ.  QSOs are probably the most important one to do first.
```

sbailey:
Closing this issue as superseded by #135 (make contaminant stars that pass cuts of other options)

PR #132, opened 2016-06-04
generate templates at given/input redshifts

### LINKED ISSUE #129: emission lines are affecting the normalization of the BGS, ELG templates
The ELG and BGS template classes -- which have emission-line spectra added to their underlying continuum spectra on-the-fly -- are not being normalized correctly.  For example, if the user requests a set of BGS spectra all normalized to r=19.5, the output r-band magnitudes of the templates varies between r~19.44 and ==19.5, with the residuals correlating with the equivalent width of H-beta.  This happens because I'm normalizing the spectrum *before* adding in the emission-line spectrum (for non-trivial reasons).  Need to fix this.

moustakas:
This isn't ready to merge, but I'd appreciate eyes on some of the core code before I go and break/change everything else.  The main point of this PR is to address #104, i.e., enabling templates to be generated at a given set of input redshifts, which we need for our full-survey data challenge.  This turned out to be a harder problem than I had first anticipated, and required a complete rewrite of the Class for each spectral class; so far I've only rewritten the ```desisim.templates.ELG()``` Class.  For example, there's some chunking and vectorized calls to the ```speclite``` filter functions as kindly suggested by @dkirkby to speed things up.

If the user passes a redshift for which *none* of the templates pass the nominal color-cuts then a warning is issued and the output spectra and metadata table rows for those objects are simply returned empty; maybe we want this to be a more dramatic exception.  I'm not sure, but @sbailey probably has an opinion.  

However, in this situation it takes quite a long time (many 10s of seconds) to consider all ~8000 templates at the desired redshift, which is too long.  For example, if you ask for 1000 templates at z=(0.1-0.4) which *also* pass the DESI ELG color-cuts then you better go for a long, long run instead of waiting around to be told that none of the templates were generated (i.e., not possible).  I note that there's no clean, fast way of knowing ahead of time which templates will work out; you have to do the math (or pre-assign templates to redshifts, e.g., in the mocks, which is something we should pursue).

I plan on addressing this slowness issue by creating a more streamlined set of ELG basis templates.  In the current set there's considerable overlap of the templates in color-color and physical parameter space, so we probably don't need more than a few hundred templates.

I still need to add a handful of unit tests, so stand by.  And this PR also fixes #129 (for ELGs at least).


moustakas:
I need to clean up the tests (Travis tests are failing right now) and add some new ones to test the newest functionality, but it would be great to get eyes sooner rather than later on the code itself (especially @sbailey, @crockosi, and @profxj).   This is a major rewrite of ```desisim.templates``` which includes a new high-level class for stars.

sbailey:
+1 for the great SUPERSTAR class name.

I tested ELGs and it seems to work with input redshifts, though even with the default zrange=[0.6,1.6] it said that ~3% of the spectra couldn't be simulated.

Timing is concerning.  100 ELGs used to take 8 sec, now it takes 1-2 minutes.  LRGs, on the other hand, seem to be much faster now (10sec instead of 2min).

moustakas:
This is useful feedback, @sbailey, thank you.  The LRGs are faster because I inadvertently left off convolving for the velocity dispersion, which I've now put back.  The ELGs are slow for the reasons I outlined at the beginning of this thread. In essence, the ability to input redshifts directly leads to a slow-down in the template-generating step since for certain redshifts all ~8000 basis templates need to be looped over (in vectorized chunks of 50).

One way to speed this up would be to reduce the size of the ELG basis set, since there is significant overlap in color-color space.  This is something I had been planning to do anyway using the archetype method I've shown elsewhere.  If there are any objections to going this route speak now!

I also need to figure out why ~3% of the redshifts in the range z=0.6-1.6 don't work.  What redshift prior did you use?  Flat?

sbailey:
I used defaults for everything, which I think is a flat redshift prior from [0.6,1.6]:
```python
In [1]: from desisim.templates import ELG

In [2]: elgmaker = ELG()
INFO:io.py:611:read_basis_templates: Reading /data/desi/spectro/templates/basis_templates/v2.2/elg_templates_v2.0.fits

In [3]: flux, wave, meta = elgmaker.make_templates(100)
WARNING:templates.py:705:make_templates: 3 spectra could not be computed given the input redshifts (or redshift priors)!
```

moustakas:
I would like to remove the columns ```GMAG, RMAG, ZMAG, W1MAG, W2MAG``` from the metadata table returned by ```desisim.templates``` since I'm already storing the 6-column ```DECAM_FLUX``` and 2-column ```WISE_FLUX``` fluxes (in nanomaggies, a la the Tractor catalogs).  [Note that this photometry is noiseless; a future PR will allow for noisy output photometry.]

I put these in originally as a convenience to the user (me) but I don't know offhand if they're used in any other downstream code.  @sbailey?

sbailey:
Outside of templates.py, template metdata quantities like GMAG, RMAG, ... are also used in desisim/spec_qa/redshifts.py (for plotting redshift success vs. magnitude) and in target.py (for filling the magnitudes in the fibermap table).  Unit tests are catching that you broke the second case but not the first.  I'm not aware of them being used outside of desisim itself.

moustakas:
I should write a few more unit tests to exercise the added functionality, but I don't want to hold up any of the other ongoing work (like the Python 3 transition), so this is ready for a more detailed review by @sbailey, @crockosi, @profxj, and whoever else is inspired.  

The Travis tests are failing for reasons that don't seem to be related to my changes; on my laptop the ```desisim.test.test_templates``` unit tests pass.

Some of the many high-level changes to ```templates.py``` include:
* The code was refactored and streamlined from top to bottom and is now about 200 lines shorter. The bulk of the real work is done by the ```GALAXY``` and ```SUPERSTAR``` classes for the various flavors of galaxies (ELG, BGS, LRG) and stars (STAR, FSTD, MWS_STAR, WD), respectively; the QSOs are treated separately.
* In addition to using flat priors as before, the user can now input any combination of redshift, apparent magnitude, and other quantities (depending on the spectral type). Alternatively, an input table can be passed with a minimum set of required columns.  The required columns are documented in the code (see the documentation for the ```input_meta``` argument), but once I write a unit test for this functionality the required inputs will be more clear. 
* Previously only the LRGs were being properly convolved for the intrinsic velocity dispersion; now the ELGs and BGS templates are handled correctly.
* The output metadata table is now type-agnostic; in other words, the metadata table for QSOs will contain quantities (like effective temperature) pertaining to stars, etc.  We can debate whether we really want this, but the change was necessary to address the next point...
* ```desisim.targets.get_targets``` was updated so the unit tests would pass, but this code really needs a lot of work (or perhaps it is becoming obsolete).
* The changes would close #129 and #104.  #135 would need a tiny bit more work to implement, but I'll probably do this as a separate PR.
* #106 is probably obsolete since we're moving to more realistic mocks, and #105 I think belongs elsewhere.
* Also note that I've broken ```desisim.spec_qa.redshifts``` but I'll open a new issue for this.

sbailey:
Tests are failing because they have become very memory hungry.  test_obs.py is using over 11GB of memory on my laptop; Travis has a ceiling of 3 GB.  desisim tests were never very fast, but now they seem to be much much slower (>5 min just for test_obs.py; I've gotten tired of waiting for it to finish and will commit this comment before it is done...).

Agreed that `desisim.targets.get_targets` is due for a major rewrite/refactor/deprecation.

moustakas:
After many struggles with ```get_targets```, tests are *finally* passing.

sbailey:
Odd.  The last test run failed with:
```
FAIL: test_newexp (desisim.test.test_obs.TestObs)
  File "/home/travis/build/desihub/desisim/py/desisim/test/test_obs.py", line 82, in test_newexp
    self.assertTrue(maxflux > 0.1, 'suspiciously low {} flux ({}); wrong units?'.format(objtype, maxflux))
AssertionError: suspiciously low LRG flux (0.0); wrong units?
```
even though the last commit was just for documentation and adding two additional test assertions in a different test, and the previous test run had passed.  I've run `test_obs` 7 times in a row on my laptop without failure, but restarting it on Travis failed again in the same way.  The test is intended to catch cases where the input templates changed units (which has happened) but perhaps this is clumsy.  I'll submit some extra debugging printouts to see why this is failing on Travis but not my laptop...

moustakas:
I spent about an hour debugging this last night but it ended up being a problem in ```targets.get_targets``` (which ```obs.new_exposure``` calls) whereby template spectra (in my case ELGs) were being assigned to (empty) sky spectra.

I don't know why the failure isn't reproducible but it's probably related to the fact that we're not fixing the random seed in ```test_obs``` (and we're only generating a few test spectra), so it's only failing only some of the time.  This is bad, of course.

I'm making 500 LRG spectra (using ```bin/simulate-templates```) to see if I can find any that are empty/zeros.

sbailey:
Comparing typical maxflux values, this test was too close to threshold.  I loosened it.  Merge conflict is due to doc/changes.rst .  I will resolve that my hand and merge.

sbailey:
Ah, I didn't see @moustakas comment prior to merge.  If you can reproduce this, please open new PR fix based on a branch off of master.

moustakas:
How's the timing with the new code?  I haven't profiled it but it'd be great to know/check whether the bottleneck is still in the velocity dispersion convolution.  I think @dkirkby had checked this previously.

weaverba137:
Note that the branch hasn't been deleted from GitHub.

sbailey:
It takes about a second per spectrum, which is a bit problematic but not a deal breaker (14k CPU hours to simulate 50M DESI spectra...).  As you noted separately, getting this *correct* was a good first step; making it faster would be helpful but can be treated as a separate issue.

It would also be helpful to make the tests faster, which could be done just by checking multiple things per template set generation instead of re-generating templates multiple times just to check one thing each time.

I'll delete the branch that Ben noticed was still there.

PR #133, opened 2016-06-08
newexp different tileids in parallel

sbailey:
This PR fixes a bug in the newexp-desi data flow where the batch jobs generating different exposures in parallel were all grabbing the same tileid and thus the same targets instead of having a different tileid per exposures.

We'll need to do a deeper refactor later coordinated with fiber assignment and the survey simulations, but this is the minimally useful step to actually get N different tiles instead of N different views of the same tile.

It is still borderline for whether or not pixel simulations can finish within the debug queue 30 minute time limit; I'll work on that in a separate branch + PR.

### DIFF
--- a/bin/newexp-desi
+++ b/bin/newexp-desi
@@ -87,7 +87,10 @@ if opts.exptime is None:
 
 #- Initialize random seeds
 if opts.seed is None:
-    opts.seed = opts.tileid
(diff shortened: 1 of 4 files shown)


sbailey:
Updated batch test to fix the case where DESI_SPECTRO_DATA wasn't set (like for Travis).  Merging now.

PR #137, opened 2016-06-27
fix quickgen mask dtype

sbailey:
This PR fixes a bug in quickgen where a mask was created as a float64 instead of an integer.  This previously silently (and perhaps incorrectly) "worked" until [desispec PR #212](https://github.com/desihub/desispec/pull/212) which enforced that masks should be integers not floats.  None of our tests caught this mismatch (see #96 for a request for quickgen tests).

Thanks to @profxj for reporting this bug.


gdhungana:
I think this is okay to merge, given the issues raised on generating arcs and relevant tests will be addressed in a separate PR.

Issue #138, opened 2016-07-01
problems with resource_filename on templates.py


fjaviersanchez:
Hi,

Maybe I am doing something wrong but when I install the package in my laptop and I try to use ```EMSpectrum``` using anaconda the ```data``` directory is not copied in anaconda's default path. Something like ```anaconda/python2.7/site-packages/desisim.version.egg/desisim```. I copied this directory in the aforementioned path and it works but, when I try to call this class from another package, the ```resource_filename``` function includes an additional ```../../```. I guess this is related to the anaconda installation but, is there any way to circumvent this? (By now I am going to try just adding the absolute path to the function on my laptop)

weaverba137:
Please tell us *exactly* how you installed desisim, and which version you are using.

weaverba137:
In the latest version of desisim, I cannot find any code that uses `resource_filename` in combination with explicit `../../` that isn't commented out or otherwise unused.  However, I did find it necessary to tweak the use of `resource_filename` in a couple of places.  I don't think that was related to this problem.

fjaviersanchez:
I installed desisim using ```python setup.py install```. The version is ```desisim==0.9.2.dev506``` and the installation was made using Anaconda Python:
```
Python 2.7.11 |Anaconda 2.3.0 (x86_64)| (default, Dec  6 2015, 18:57:58) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
```
I found that ```resource_filename``` is used here: https://github.com/desihub/desisim/blob/master/py/desisim/templates.py#L133-135
but not in combination with ```../../```

So I think there are two problems here (I believe that mainly due because of the way I installed it):

1) I installed this using ```python setup.py install``` that makes a copy into the anaconda ```site-packages``` and it doesn't copy the ```data``` directory.

2) Once I copied this directory into the anaconda repository the problem was that I was calling ```EMSpectrum``` inside a package, ```bayez```, that imported ```desisim``` internally. I think that the problem for ```resource_filename``` is that it found ```desisim``` in ```../../desisim``` referenced from the ```bayez``` repository.

fjaviersanchez:
Solved in ```desisim==0.10.0```

PR #139, opened 2016-07-01
Resource filename use

weaverba137:
This PR fixes a minor issue related to the use of `resource_filename()`.  According to http://setuptools.readthedocs.io/en/latest/pkg_resources.html#basic-resource-access, one should not use `os.path.join()` to construct the second argument:

> Note that resource names must be /-separated paths and cannot be absolute (i.e. no leading /) or contain relative names like "..". Do not use os.path routines to manipulate resource paths, as they are not filesystem paths.

Also, I removed any remaining explicit use of `..`, even though they were not being used.

### DIFF
--- a/py/desisim/qso_template/desi_qso_templ.py
+++ b/py/desisim/qso_template/desi_qso_templ.py
@@ -1,5 +1,5 @@
 """
-#;+ 
+#;+
 #; NAME:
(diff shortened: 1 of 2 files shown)


moustakas:
Thanks for fixing this @weaverba137.  This PR also hopefully resolves #138.

PR #145, opened 2016-07-15
Quickgen update

### LINKED ISSUE #136: quickgen doesn't work with newexp-desi arc exposures
Passing on a report from @profxj when trying to run quickgen on a `newexp-desi --flavor arc` exposure:
```
profx.ucolick.org> quickgen --simspec $simdir/simspec-00000001.fits --fibermap $simdir/fibermap-00000001.fits
Reading fibermap file /u/xavier/DESI/DESI_SCRATCH/sim/qa_jun2016/20160614/fibermap-00000001.fits
Now Reading the input file /u/xavier/DESI/DESI_SCRATCH/sim/qa_jun2016/20160614/simspec-00000001.fits
  File "/Users/xavier/anaconda/bin/quickgen", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/xavier/DESI/desisim/bin/quickgen", line 124, in <module>
    wavelengths = simspec.wave['brz']
KeyError: ‘brz'
```

Please fix this, and also add tests (see #96) so that next time we will catch the problem at the time it is first broken, rather than later when we are just trying to use newexp-desi + quickbrick.

rstaten:
This PR addresses several requested updates for quickgen:

- fixes #136 to accomodate arcs and produces output for arc exposures
- accomodates additional object type (BGS, MWS)
- refactor quickgen so scripts has primary functions and bin has top-level
- add unit test/integration test for quickgen

Note: fiberloss for arcs treated as perfect.  BGS treated as LRG, MWS treated as STAR.  Investigating difference between arc normalization via pixsim and quickgen (arc flux in electrons/second.)


moustakas:
Awesome!  Feel free to ignore if this is outside the scope of the current PR, but if you could add a basic test for quickbrick then we could simultaneously close #96.

sbailey:
This looks like a great refactor.  Thanks.  I have attempted to add specsim to the travis install so that the tests will run there.  I also added a default $SPECPROD for the tests, re-enabled the test on flavor='arc', and backed off to fewer flavors of more spectra so that the tests will run faster -- they are dominated by the specsim launch time which happens once per flavor, and things like dark vs. gray vs. elg are largely testing the same thing.

There are also some conflicts with master.  I'll take care of resolving those after the tests are working on Travis.

sbailey:
The travis tests are getting killed while running quickgen.  From watching the memory load on my laptop while this runs, it appears that quickgen is taking 5-6 GB which is probably blowing the limits available to the travis virtual machine.  Do you know why/if quickgen is so memory hungry?  Even at high resolution, the input spectra should only be a few megabytes, not gigabytes.

sbailey:
The memory spike is occurring during
```
qsim.instrument.cameras[i].get_output_resolution_matrix()
```
and in particular it is while getting the 'r' band matrix that the test is killed.

Watching this on my laptop, the memory spike is only temporary; once the resolution is obtained the memory goes back down.

@dkirkby do you know why get_output_resolution_matrix() temporarily uses lots of memory (e.g. does it go via a dense array?) and what could be done about that?

We could potentially work around this for the Travis tests by doing the quickgen test simulations at a much lower resolution, but it would probably be beneficial for quickgen and specsim to use less memory if possible anyway.

sbailey:
I have opened a new desihub/specsim#43 issue to fix why `specsim.instrument.Camera. get_output_resolution_matrix()` is so memory hungry.  In the meantime I'm turning off this test for Travis while leaving it enabled for non-travis environments so that we can move on with merging this PR and proceeding with the python3 upgrade.  Longer term we should fix specsim and/or run these tests on Travis at lower resolution (it wasn't immediately obvious to me how to tell specsim to simulate at a lower resolution).

sbailey:
I resolved the bin/quickgen merge conflict by hand and merged into master.  Closing PR and deleting branch.

PR #146, opened 2016-07-19
QA zfind plot updates

sbailey:
Fixes fibermap OBJTYPE vs. simspec OBJTYPE usage so that fake QSO targets don't count in the QSO statistics.  This removes a vertical stripe at ztrue=0, but doesn't substantially change the interpretation of a large number of QSOs are failing redshift fitting.

Propagates desisim and specter versions to output figure instead of placeholder SIMSPECV=9.999.  Includes value of $SPECPROD and $PIXPROD, with a default of "unknown" if they aren't set. (separate to do: we should have pipeline include those in the file headers, and then get them from there instead relying upon whatever they are currently set to).

Misc plot ranges / axis tweaks.

qaz

### DIFF
--- a/py/desisim/spec_qa/redshifts.py
+++ b/py/desisim/spec_qa/redshifts.py
@@ -201,15 +201,23 @@ def load_z(fibermap_files, zbest_files, outfil=None):
 
         log.info('Reading: {:s}'.format(fibermap_file))
         # Load simspec
-        simspec_fil = fibermap_file.replace('fibermap','simspec')
-        sps_hdu = fits.open(simspec_fil)
+        simspec_file = fibermap_file.replace('fibermap','simspec')
+        sps_hdu = fits.open(simspec_file)
         # Make Tables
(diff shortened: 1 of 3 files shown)


py/desisim/spec_qa/redshifts.py
```diff
@@ -891,12 +901,13 @@ def dz_summ(simz_tab, pp=None, pdict=None, min_count=20):
...
-                        
+            axis.set_xlim(x_min, x_max)
+            
             # Hide overlapping x-axis labels except in the bottom right.
             if overlap and (col < ncols - 1):
-                plt.setp(
-                    [axis.get_xticklabels()[-overlap:]], visible=False)
+                axis.set_xticks(axis.get_xticks()[0:-overlap])
+                # plt.setp(
+                #     [axis.get_xticklabels()[-overlap:]], visible=False)
```
profxj: Remove the commented lines (no longer of interest)

profxj:
One minor comment.  Merge when ready.

PR #148, opened 2016-08-18
add rand seeds for batch pixsim

sbailey:
Adds propagation of random seeds to the `desisim.batch.pixsim` utility functions for batch generation of pixel-level simulations.  This is needed for before/after testing during the python3 migration (and is good to have anyway).

Note: the FITS header CHECKSUM and DATASUM headers include the timestamp of when they were generated, thus causing the sha1sum of the files to differ, even though the data and headers are otherwise identical.  To work around this, I compared the fits files from two batch newexp+pixsim runs using `fitsdiff` commands like:
```
fitsdiff -k CHECKSUM,DATASUM 20160818a/pix-r0-00000003.fits 20160818b/pix-r0-00000003.fits

 fitsdiff: 1.1.1
 a: 20160818a/pix-r0-00000003.fits
 b: 20160818b/pix-r0-00000003.fits
 Keyword(s) not to be compared:
  CHECKSUM DATASUM
 Maximum number of different data values to be reported: 10
 Data comparison level: 0.0

No differences found.
```

### DIFF
--- a/py/desisim/batch/pixsim.py
+++ b/py/desisim/batch/pixsim.py
@@ -109,9 +109,9 @@ def batch_newexp(batchfile, flavors, nspec=5000, night=None, expids=None,
         fx.write('mkdir -p $DESI_SPECTRO_SIM/$PIXPROD/{}\n'.format(night))
         fx.write('\n')
         
-        for i, expid, flavor, tileid in zip(range(nexp), expids, flavors, tileids):
+        for expid, flavor, tileid, seed in zip(expids, flavors, tileids, seeds):
             fx.write(cmd.format(nspec=nspec, night=night, expid=expid,
-                flavor=flavor, tileid=tileid, seed=seeds[i])+' &\n')
+                flavor=flavor, tileid=tileid, seed=seeds)+' &\n')
             
         fx.write('\nwait\n')
         fx.write('\necho Done at `date`\n')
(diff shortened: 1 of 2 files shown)


py/desisim/batch/pixsim.py
```diff
@@ -32,14 +33,17 @@
...
+    nodes=None, pixprod=None, desi_spectro_sim=None, tileids=None, seed=None):
     '''
     Write a slurm batch script for run newexp-desi for the list of flavors
     '''
     nexp = len(flavors)
     timestr = '00:30:00'
     logfile = '{}.%j.log'.format(batchfile)
-    
+
+    np.random.seed(seed)
```
weaverba137: Does `np.random.seed()` accept `None` as an input?

py/desisim/batch/pixsim.py
```diff
@@ -172,10 +180,10 @@ def batch_pixsim(batchfile, flavors, nspec=5000, night=None, expids=None,
 
         fx.write('\necho Starting at `date`\n')
 
-        for expid, flavor in zip(expids, flavors):
+        for i, expid, flavor in zip(range(nexp), expids, flavors):
```
weaverba137: You could also do

```python
for seed, expid, flavor in zip(seeds, expids, flavors):
```

py/desisim/batch/pixsim.py
sbailey: Yes, in fact `None` is the default value if you call `numpy.random.seed()` without an argument.  This is passed to `numpy.random.RandomState` which is more explicit that `None` is valid:

>    If `seed` is ``None``, then `RandomState` will try to read data from
    ``/dev/urandom`` (or the Windows analogue) if available or seed from
    the clock otherwise.

weaverba137:
Assuming the tests pass, I think this can be merged.

PR #151, opened 2016-08-24
fix template fibermap mag; fix quickgen outputs

sbailey:
This PR fixes the MAG calculation for the simulated fibermap, which was broken as part of PR #132.  The previous flattening/slicing/stacking logic was wrong and the same magnitudes were replicated across multiple targets.  This was the core problem for the integration test failures.

Other changes that came long for the ride as a result of debugging this:
* adds exposure flavor "std" to newexp-desi to simulate a bunch of standards; useful for debugging
* several quickgen fixes:
  * include fibermap in frame, cframe output
  * include CAMERA header keyword
  * fiberflats are referenced to 1.0 not 0.0

### DIFF
--- a/bin/newexp-desi
+++ b/bin/newexp-desi
@@ -66,7 +66,7 @@ if opts.night is None:
     opts.night = obs.get_night(utc=time.gmtime())
 
 opts.flavor = opts.flavor.lower()
-if opts.flavor not in ['arc','flat','dark','gray','grey','bright','bgs','mws','lrg','elg','qso']:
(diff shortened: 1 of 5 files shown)


moustakas:
The changes to the code look good.  I removed some now-obsolete commented out code and also made the default redshift to be zero in ```desisim.io.empty_metatable```, to avoid some harmless but annoying divide-by-zero errors downstream.  

My only concern is that ```test_obs.test_newexp``` seems to occasionally fail for reasons that I don't understand (and for reasons that seem to be unreproducible).  We might think about changing/updating this test.

PR #154, opened 2016-08-25
2to3

sbailey:
This PR implements the changes recommended by `2to3`.  It has been tested against python 2.7.11 but not yet under python 3.5.  That will be a separate update and PR after this one has been reviewed and merged.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -12,6 +12,8 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+from __future__ import absolute_import, division, print_function, unicode_literals
+
 import sys
 import os
 import os.path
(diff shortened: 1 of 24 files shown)


moustakas:
I spot-checked about a dozen scripts that I have previously worked on and everything looks pretty good.  And I'm fine with writing ```for key in thisdict``` rather than ```for key in thisdict.keys()``` -- the former is slicker and it's a nice short-hand "non-experts" should/would learn.

And for the record I'm fine with merging this PR before #155, and I'll resolve the conflicts (if any).

sbailey:
I'll take John's review as sufficient and merge this now so that he can review and resolve any conflicts with PR #155.

PR #155, opened 2016-08-25
interpolate stellar templates on an input physical grid for the MWS mocks

moustakas:
This PR addresses a non-existent issue (maybe #72) but is needed to connect the MWS mocks to the templates code, and therefore the current full-survey Data Challenge.  Some groundwork has also been laid to help connect the BGS mocks to the same code.  @apcooper, @forero, @crockosi, @sbailey -- please take a look at the notebook below.

In essence, ```desisim.templates``` can now (optionally) take an ```star_properties``` table, with the following *required* inputs: ```REDSHIFT, MAG, TEFF, LOGG, and FEH``` (actually, ```FEH``` is optional, which means the code can also interface with a WD mock).

To show the new capability for a particular mock generated by @apcooper, I created a short notebook that generates 1000 spectra from a randomly chosen brick:
  https://github.com/desihub/desisim/blob/mwsinterp/doc/nb/mws-mock.ipynb

Some "features" of the current code include:
* If the desired stellar properties are outside of the boundaries of the template grid then no interpolation is attempted and that spectrum simply isn't computed.  I don't know if we want the interpolation code to be smarter or if we want to pre-filter the mock catalog.  Note that in the output metadata table these "non-spectra" are in the same row as in the ```star_properties``` table but have ```TEMPLATEID``` set to ```-1```.
* There is very little error catching to ensure all the inputs are present and correct, so user beware!
* The code currently just does linear interpolation (on a 3D grid), but we could also play around with cubic.  I haven't looked extensively at the output (interpolated) spectra other than some very basic sanity tests, nor have I done any timing tests.  Help on this front would be welcome.
* There's probably a limit to how many objects we can generate at a time, but I haven't found this boundary.  We probably only want to generate of order ~5000 spectra at a time, but we (someone) should test this number.

And thanks to @forero for pointing me to the appropriate ```scipy``` interpolation routine and for writing a minimum working example.


moustakas:
PS.  The Travis tests for ```desisim``` are being flaky -- sometimes they will pass and sometimes they will fail with no rhyme or reason, or they'll fail online but pass on my laptop.  We need to figure this out...

py/desisim/templates.py
```diff
@@ -340,10 +341,9 @@ def __init__(self, objtype='ELG', minwave=3600.0, maxwave=10000.0, cdelt=2.0,
             corresponding to BASEFLUX (Angstrom).
           basemeta (astropy.Table): Table of meta-data [nbase] for each base template.
           pixbound (numpy.ndarray): Pixel boundaries of BASEWAVE (Angstrom).
-          decamwise (speclite.filters instance): DECam2014-* and WISE2010-* FilterSequence
-          gfilt (speclite.filters instance): DECam2014 g-band FilterSequence
           rfilt (speclite.filters instance): DECam2014 r-band FilterSequence
-          zfilt (speclite.filters instance): DECam2014 z-band FilterSequence
+          normfilt (speclite.filters instance): FilterSequence of self.normfilter
+          decamwise (speclite.filters instance): DECam2014-* and WISE2010-* FilterSequence
```
sbailey: docstring seems out of since with actual `__init__` options: there doesn't appear to be a `decamwise` option,
`normfilt` vs. `normfilter`, and it is unclear to me if the `normfilter` option should be a string, a `FilterSequence` object, or either.

moustakas: In the documentation, ```rfilt```, ```normfilt```, and ```decamwise``` are all attributes, not arguments of the Class, so I think I'm in the clear.  And although the naming isn't great, the documentation is also clear that ```normfilter``` is a *string* argument (for the desired normalization bandpasses), whereas ```normfilt``` is a FilterSequence  instance and an attribute of the Class.

moustakas:
* Template tests and output are friendlier, faster, reproducible, and also seem to be more stable.  
* Optional input ```input_data``` was changed to ```star_properties```, per a conversation with @sbailey.
* A major bug in the BGS templates was identified and fixed.  
* The changes in PR #154 were merged with minimal conflicts.

This is ready for final review and merging @sbailey.

sbailey:
Thanks for the normfilt vs. normfilter clarification; I had misread the Args vs. Attributes labels in the docstring.  I think I was expecting the created object attributes to be documented in the class docstring and the arguments to \_\_init\_\_ to be documented in the \_\_init\_\_ docstring, as they do in the [Google docstring style examples](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).  Not a blocker for this PR.  Merging now.

moustakas:
Please let me know if there's a DESI convention we should be adopting regarding the documentation.  I had previously spent a lot of time looking over the docstring style example page you linked to and it's actually ambiguous on this point:
```
    The __init__ method may be documented in either the class level
    docstring, or as a docstring on the __init__ method itself.

    Either form is acceptable, but the two should not be mixed. Choose one
    convention to document the __init__ method and be consistent with it.
```
In the end I decided to document the attributes of the class in the ```___init___``` docstring because many of the arguments to ```___init___``` end up becoming attributes, so it made more sense (to me) to have them documented within the ```___init___``` function itself.

Other expert opinions?

PR #156, opened 2016-08-29
Lya sims

ngbusca:
added the module lya_spectra with a function get_spectra to generate qso-like spectra that includes lyman-alpha forest absorption. 

Example run:

In [41]: import desisim.lya_spectra  
In [42]: flux,wave,meta=desisim.lya_spectra.get_spectra("simpleSpec_0.fits.gz",first=1,last=10)

where the  simpleSpec_0.fits.gz is in /project/projectdirs/desi/users/ngbusca/lya_files/v0/

TODO:
1. From (John Moustakas): "When we go to make the _actual_ spectra for the Data Challenge we'll need to call this function in desisim.targets.get_targets (or wherever), which means we'll probably need a function in desisim.io (and maybe an environment variable) to _find_ the mock Lyman-alpha spectra. "
2. The input_meta ignores the extraneous RA, DEC columns and doesn't propagate them into the output metadata table, so we'll want to either pack them into the output table or keep a separate (row-matched) catalog of inputs.
3. We should put the Lyman-alpha spectra in a standard, versioned directory on NERSC rather than in the current directory
   /project/projectdirs/desi/users/ngbusca/lya_files/v0/simpleSpec_*
4. Need to fix the normalization of the spectra: currently they are normalized before adding the forest
5. Need to make smaller wavelength pixels. Currently they are 2A, certainly too coarse.


sbailey:
Could you review this @moustakas, or perhaps otherwise comment that you don't have the time this week due to the DECaLS meeting?  I'm wondering if this should be integrated into the templates module instead of being standalone.

Let's not merge this until the simpleSpec_0.fits.gz has a data model and a good canonical location (item 3 above).

moustakas:
```
3. Need to fix the normalization of the spectra: currently they are normalized before adding 
the forest.
```
Assuming we want to keep Lyman-alpha forest stuff outside of ```desisim.templates```, we should be able to apply the appropriate correction pretty simply (stealing liberally from existing code...).  At the top:
```
from speclite import filters
normfilter = 'decam2014-g'
normfilt = filters.load_filters(normfilter)
```
And then inside the loop, after the forest attenuation has been applied:
```
padflux, padwave = normfilt.pad_spectrum(flux[i, :], wave, method='edge')
normmaggies = np.array(normfilt.get_ab_maggies(padflux, padwave, 
  mask_invalid=True)[normfilter])
flux[i, :] *= 10**(-0.4*input_meta['MAG'][ii]) / normmaggies
```

```
4. Need to make smaller wavelength pixels. Currently they are 2A, certainly too coarse.
```
We can easily change this of course via the ```cdelt``` optional input, but now that I look at it I'm wondering why I used ```2 A/pixel```.  I think that must have been just for testing; this should be more like ```0.5 A/pix```, or even ```0.1 A/pix```, since the sampling to the correct pixel scale (which is close to ```0.5 A/pix```) and resolution happens in ```specsim```.  

@sbailey Should we change the default ```cdelt``` value in ```desisim.templates``` to ```0.1 A/pix``` or ```0.2 A/pix``` (noting that ```desisim.targets.get_targets``` and other wrapper scripts appear to (incorrectly) call the code with the default ```cdelt``` parameter)?

moustakas:
@sbailey I addressed items (4) and (5), which I'm leaving to @ngbusca to implement. Items (1) and (3) I'd like to offload to @forero, our pipeline/mock overseer/organizer (but please correct me if this is an incorrect assumption!), and point (2) is more of a heads up that will be true of all the mock catalogs.

I don't think this should be part of ```desisim.templates``` because it's tied to a particular set of mocks which are defined at specific redshifts and g-band magnitudes, and all the mocks are going to have/need wrapper scripts like this of varying complexity.  But we may change our mind once we dig a little deeper...

sbailey:
@moustakas I don't understand the point about the default `cdelt`: it seems a tradeoff between speed and accuracy.  In what way is using the current default "incorrect"?

@ngbusca is the format of simpleSpec_0.fits.gz tied to anything specific about Jean-Marc's mocks, or is it sufficiently generic that the format and this code could be reused for any other future LyA mock?  The *contents* are obviously tied to his mocks, but is the *format*?  If the format or code is tied to J-M's mocks, then we should find a location for this that makes room for future kinds of LyA mocks.  If the format is general enough that it can be reused just by updating the file, then the current location is probably ok.

moustakas:
```
@moustakas I don't understand the point about the default `cdelt`: it seems a tradeoff between
speed and accuracy.  In what way is using the current default "incorrect"?
```
It's entirely possible I don't understand how the infinite signal-to-noise, infinite resolution templates returned by ```desisim.templates``` are turned into "real" DESI spectra by ```specsim```, but if ```desisim.templates``` returns spectra that are only sampled every 2 Angstrom and then ```specsim``` downsamples/interpolates them to have 0.5 A pixels (or whatever the detailed pixel scale is, depending on the camera and wavelength) then the emission-line profiles (and therefore the integrated line-fluxes) will not be correct.

sbailey:
Ah!  I thought the choices in play were 0.1 A/pix vs. 0.2 A/pix, not 2 A/pix.  Now I understand, and yes indeed 2 A/pix is way too coarse as a default.  0.2 A/pix seems ok.

`desisim.targets.get_targets()` used the `wave` parameter instead of `wavemin`, `wavemax`, and `cdelt`.  And it is creating that wave array using the equivalent of cdelt=0.2 A/pix, so I think the current `get_targets()` code is ok (in this regard...)

moustakas:
Fantastic.  ```desisim.scripts.quickbrick``` is doing the wrong thing, so I'm going to update the default ```cdelt```.

forero:
@ngbusca  I've already copied the files to the standard place where we are storing all the mocks (`$DESI_ROOT/mocks`). Could you please add a README to have some traceability for those files? 

I am starting to compile all the mock files. This week I will work on having a single module to process `ra,dec,z` and another module to generate the corresponding spectra.

ngbusca:
Fixed the spectral normalization with the code snippet suggested by @moustakas. Also changed the filter to sdss2010-g to match the g-band in Nathalie's QLF measurements. 

As per issue 5. it should be ok once the default cdelt is changed to a smaller value, so left as was.

@sbailey The input-file format is quite generic and makes no reference to te specific's of JM. simulations.

ngbusca:
@forero can you please remind me where DESI_ROOT points at?

moustakas:
```
ioannis@cori03:~% echo $DESI_ROOT
/project/projectdirs/desi
```

ngbusca:
Are there any changes that need to be done before accepting this pull request? Or maybe this feature is meant to be left in a separate branch?

moustakas:
Can you look into why the Travis tests aren't passing?  There have been a lot of changes to desisim and specsim since this was last worked on...

sbailey:
Apologies; I lost track of this one.  I'll do final review and hopefully merge tomorrow.

sbailey:
I made a few changes to `lya_spectra.read_spectra`:
* made `first` option 0-indexed instead of 1-indexed to match python indexing style
* changed `last` option to `nqso` to avoid ambiguity of whether last is included or not.
* added `seed` option
* converted tabs to spaces
* Added RA and DEC columns to the output metadata table.

Will the extra RA DEC columns mess anything up since the metadata table returned by `lya_spectra.read_spectra` now has two more columns than the metadata table returned by `templates.QSO().make_templates`?  i.e. you can't trivially stack it with metadata tables from other object types unless you remove those columns or add RA and DEC to the other tables.  I think this is ok...

For the record: other mock *catalogs* are read via `desitarget.mock.io`, including extracting (ra,dec,z) from these the same set of Lyman-alpha mocks that this uses.  I'm leaving this code here in `desisim` because this does seem similar to template making that is also here, i.e. generating simulated spectra and not just reading in an (ra,dec,z) catalog.

I also added unit tests with a lightweight example Lyman-alpha mock file.

Travis tests just failed, so going to investigate that now...

sbailey:
Update:
* Tests were passing prior to my changes
* I made some changes and added tests to cover the `lya_spectra.read_spectra` code, but they are disabled on Travis because I haven't updated it to install fitsio yet
* The remaining tests run for about 8 minutes and then spend 10 minutes without any log output, which in the past was caused by exceeding the 3 GB memory footprint of the Travis virtual machines.
* Tests were taking 2.7 GB on my laptop; I updated them to save memory and now they only take 1.7 GB on my laptop but they are still hanging on Travis.
* still thinking about what to do next...

moustakas:
@sbailey wrote:
```
Will the extra RA DEC columns mess anything up since the metadata table returned by 
lya_spectra.read_spectra now has two more columns than the metadata table returned by 
templates.QSO().make_templates? i.e. you can't trivially stack it with metadata tables from other 
object types unless you remove those columns or add RA and DEC to the other tables. I think this 
is ok...
```
I standardized the set of columns in the metadata tables precisely because there was a need for them to be stacked on top of one another, e.g., when building sets of targets.  So it's possible that the inclusion of ra, dec here will break some downstream code at some point.  I've lost track of how the template-generating code is interacting with the mock catalogs, however, so I'm not sure what will break.  @forero can you point to the type of code that would call ```lya_spectra.read_spectra``` and handle its output?

sbailey:
@moustakas since the metadata are astropy Tables, they have the ability to vstack even if there are extra columns.  The resulting Table fills in masked values for the missing entries.  Example below.  It might still be a bad idea, but it doesn't just crash the code.

I'm going to proceed with this merge since this PR has been dangling out there for far too long (apologies to @ngbusca).  If these extra columns create a problem, then we can remove them from here and change this interface to return RA,dec differently (Note: there is a separate routine in desitarget for extracting just the (RA,dec) but it would seem wasteful to have to reopen and parse these files with 50k HDUs each just to extract the RA,dec that we already have...)

Travis tests have been updated to sudo=True that gives a different virtual machine with slower boot times but more memory.  I still claim that our tests are under the documented 4 GB memory limit of the previous sudo=false machines, but this is a pragmatic step to move forward.

```python
In [1]: from astropy.table import Table, vstack
   ...: import numpy as np
   ...: a = Table()
   ...: a['x'] = np.arange(3)
   ...: a['ra'] = np.arange(3)
   ...: a['dec'] = 2*np.arange(3)
   ...: b = Table()
   ...: b['x'] = -np.arange(3)
   ...: 

In [2]: a
Out[2]: 
<Table length=3>
  x     ra   dec 
int64 int64 int64
----- ----- -----
    0     0     0
    1     1     2
    2     2     4

In [3]: b
Out[3]: 
<Table length=3>
  x  
int64
-----
    0
   -1
   -2

In [4]: vstack([a,b])
Out[4]: 
<Table masked=True length=6>
  x     ra   dec 
int64 int64 int64
----- ----- -----
    0     0     0
    1     1     2
    2     2     4
    0    --    --
   -1    --    --
   -2    --    --
```

PR #159, opened 2016-09-11
Py3

sbailey:
This PR updates desisim to be compatible with both python 3.5 and 2.7; it requires the recently minted desispec 0.10.0 and desiutil 1.8.0.

@weaverba137 and/or others — please review and comment.

Full disclosure: I had to loosen the tolerance on the DECAM_FLUX check when testing that we get the same templates back when providing a metadata table.  It is still close enough agreement that it doesn't really matter on a practical level, but the disagreement is much worse than machine precision so it is slightly concerning.  This is a known issue after @moustakas last refactor; we may need to chase this down later.


sbailey:
I'll need a new version of specsim for this to work with the Travis tests, since the latest tag of specsim isn't yet compatible with py3.5 (master is, but not v0.4).  I'll ping @dkirkby .

dkirkby:
I just tagged and released specsim v0.5: [docs](http://specsim.readthedocs.io/en/stable/) | [pypi](https://pypi.python.org/pypi/specsim/0.5).

sbailey:
Thanks for the specsim/v0.5 release, which the travis tests now use.  Travis tests now pass on py 2.7 and 3.5; this PR is ready for final review @weaver and/or others.

weaverba137:
This looks good to me.

PR #165, opened 2016-09-20
Get desisim working on ReadTheDocs

weaverba137:
desisim should now build successfully on ReadTheDocs.  You can view the successful build of this branch at https://desisim.readthedocs.io/en/rtd-test/.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,2 @@
 setuptools
-numpy
-scipy
-astropy
-pyyaml
 git+https://github.com/desihub/desiutil.git@1.8.0#egg=desiutil
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,8 +3,4 @@ numpy
 scipy
 astropy
 pyyaml
-speclite
-specsim
-svn+https://desi.lbl.gov/svn/code/desimodel/branches/pip-install#egg=desimodel
-git+https://github.com/desihub/desispec.git@0.8.0#egg=desispec
-git+https://github.com/desihub/desiutil.git@1.6.0#egg=desiutil
+git+https://github.com/desihub/desiutil.git@1.8.0#egg=desiutil


sbailey:
Although we haven't been using requirements.txt particularly rigorously, I'm curious about the motivation for removing so many of the things that the code actually does depend on, while leaving desiutil there.  What's the conceptual direction you are going with this?  I'm also adding @dkirkby to this ticket since I think he is someone who actually does pay attention to requirements.txt and might have opinions on what it should (not) have.

weaverba137:
ReadTheDocs is the only thing that explicitly needs requirements.txt.  I removed everything except setuptools and desiutil in order to speed up the builds.  setuptools and desiutil are needed to make the setup.py file functional.  Remember, ReadTheDocs is only building the documentation, not running any tests.

weaverba137:
PS, if you still want the full listing in requirements.txt, you can also create a separate file with only the requirements needed by ReadTheDocs, and have it use that.  In other words, we'd have two requirements files, that could easily get out-of-sync.

dkirkby:
For reference, an astropy-affiliated pkg (specsim, speclite, ...) identifies three different types of dependencies:
* installation dependencies (ie, other packages necessary for `import thispkg` to work) are specified via `setup.py` and used by `pip install`, `python setup.py`, etc.  These are simply listed inline, e.g. for specsim:
```
install_requires=['astropy', 'scipy', 'pyyaml', 'speclite'],
```
You sometimes see the contents of `requirements.txt` being used here.
* unit test dependencies (to support `python setup.py test`) are specified by astropy-specifc environment variables in .travis.yml, e.g. for specsim:
```
        - NUMPY_VERSION=stable
        - ASTROPY_VERSION=stable
        - PIP_DEPENDENCIES='speclite'
        - CONDA_DEPENDENCIES='scipy pyyaml matplotlib'
```
* documentation dependencies (to support `make html` in docs/) are specified via a file `docs/rtd-pip-requirements` which, for specsim, contains:
```
numpy
matplotlib
Cython
astropy-helpers
astropy
```

Note that matplotlib is not an install dependency, but still needed by a few unit tests and to generate intersphinx documentation links.  I'm not sure why Cython is included in the last set - I should try removing it and see what happens.

sbailey:
OK, thanks for the explanations and background.  I'll merge this now to facilitate getting desisim onto readthedocs.  If someone wants to update and maintain how we specify dependencies, please propose a specific plan (e.g. the one that @dkirkby described for astropy affiliates) and commit to maintaining that.

PR #168, opened 2016-09-23
Tiny one-line fix so that function from desisim.pixsim can be called.

tskisner:
Without this line, I get an error running on python 2.7:

File "/home/kisner/software/desi27/lib/python2.7/site-packages/desisim-0.14.0.dev684-py2.7.egg/desisim/scripts/pixsim.py", line 232, in main
    image, rawpix, truepix = desisim.pixsim.simulate(
AttributeError: 'module' object has no attribute 'pixsim'

This line fixes the error.  I am not sure why this does not get triggered by the desispec integration test!

### DIFF
--- a/py/desisim/scripts/pixsim.py
+++ b/py/desisim/scripts/pixsim.py
@@ -12,6 +12,7 @@ import desimodel.io
 import desispec.io
 
 import desisim
+import desisim.pixsim
 from desisim import obs, io
 from desispec.log import get_logger
 log = get_logger()
--- a/py/desisim/scripts/pixsim.py
+++ b/py/desisim/scripts/pixsim.py
@@ -265,3 +265,4 @@ def main(args=None):
     if mpicomm.rank == 0:
         log.info('Finished pixsim {} expid {} at {}'.format(args.night, args.expid, asctime()))
 
+


weaverba137:
I recommend `from . import pixsim` and removing `import desisim`.

tskisner:
This change is to the main script (desisim.scripts.pixsim.py) not desisim.pixsim.  I guess we could do relative imports from ".." of all the packages at the top of the file that are in desisim.

weaverba137:
Ah, OK.  So do you need to both `import desisim` and `desisim.pixsim`?  Because all `import desisim` gets you is `__version__`

sbailey:
@moustakas it looks like the py3.5 test failed with that intermittent bug of all fluxes being 0.  I think the travis log contains the seeds necessary to reproduce it (after associating which seed was the one actually used for the test): https://travis-ci.org/desihub/desisim/jobs/162317446

I'll let Ben and Ted sort out relative vs. absolute imports.  I am also confused why unit tests and integration tests didn't catch the problem that Ted reports here.

tskisner:
This started out as a one-line fix so that a script would run...  If we want to open up the discussion to changing all the absolute imports of desisim.* that occur inside desisim.scripts.*.py to make them relative, then I am fine with that.  However, I think we should be consistent across all those main scripts.

weaverba137:
Sorry, let's just get the script running.

sbailey:
I think the previous travis failure was from an unrelated intermittent template bug.  I push a whitespace-only change to trigger a new travis test without abandoning the old one that might be useful for debugging.  If this test passes, I'll merge this dangling PR.

sbailey:
3 failures in a row for the PR test but not the push test.  Something else is going on.  Investigating...

sbailey:
I was unable to reproduce this failure on my laptop despite pegging to the same versions of every package and multiple reruns.  Yet another rerun on travis passed.

Something is going on that needs to be fixed, but this PR itself brings in an unrelated but necessary change for the integration test to pass, so I'll merge this now.

PR #166, opened 2016-09-21
quickbrick tests and bug fixes

sbailey:
This PR
 * adds tests for quickbrick
 * fixes #161 (updating how exposure time is set with specsim v0.5)
 * fixes #162 (quickbrick crash)
 * closes #96 (tests for quickgen and quickbrick)

It also fixes some bugs that weren't previously known and were caught by the new unit tests:
 * --seed option wasn't working
 * truth table output was corrupted using py3

However:
  * The tests won't work on Travis until desihub/specsim#43 is addressed to reduce the specsim memory footprint or another workaround is found.  I've flagged these with `@unittest.skipIf('TRAVIS_JOB_ID' in os.environ)` wrappers.  Note that this also impacts the quickgen tests -- you can run them on a laptop with sufficient memory, but not on Travis.  In the meantime, we also won't get accurate coverage reports on coveralls.io for these.
  * The tests are slow (2.5 minutes for just the quickbrick tests).  Although quicksim is fast for N>>1 spectra in a single shot, generating the templates on the fly is not, and quicksim has a fairly large overhead that we encounter multiple times when comparing things like "did making the moon brighter make the spectra noiser?"
  * I'm not sure what the original intent of OBJTYPE vs. TRUE_OBJTYPE was for the truth table.  What I have in there currently at least runs, but it may need further modification.

Related: apologies for the noise I generated on issue #161 where I kept typing "specsim" when I meant "desisim".  @dkirkby and I meant the same thing but I kept typing the wrong thing and misreading the right thing that he was typing...

Issue #161, opened 2016-09-18
Update quickbrick and quickgen to specsim v0.5


dkirkby:
The specsim exposure time parameter moved from `qsim.instrument` to the new`qsim.observation` in v0.5, but quickbrick and quickgen are still setting the value in `qsim.instrument` (which does not raise any exception, since python allows you to add attributes to an object on the fly).  The end result is that the exposure time is always fixed at 1000s.

This issue is to fix this problem, test the fix, and check for any other gotchas with the transition to specsim v0.5.

Thanks to @moustakas for flagging this!

sbailey:
Thanks for starting an issue about this.  We really should have unit tests that catch this sort of change-in-API problems.  This one is a bit tricky because of how the exposure time is set via a directly exposed member variable instead of a function (which generally I like instead of the super-strict C++ style "hide every variable and provide getter/setter functions", but there sometimes are gotchas).  Some basic sanity check unit tests would have at least caught the symptom to prompt us to go find the problem:
* do longer exposure times generate lower S/N spectra?
* do higher airmass observations generate noisier spectra?
* are bright time spectra noisier than dark time spectra?
* etc.

dkirkby:
specsim already has some functional unit tests along those lines, but they don't help when another program isn't using specsim correctly (due to an API change in this case).

The unit test problems in desisim are more basic since the quickbrick program on the master branch does not even run currently, as far as I can tell.  I will open a separate issue about that.

sbailey:
Yes, I specifically was referring to unittests in specsim that should catch if we are not using an external library correctly.

dkirkby:
I can't think of any unit test in specsim that could have caught this problem.   Did you have something in mind?

sbailey:
e.g. the list in my initial reply, in particular "do longer exposure times generate lower S/N spectra".  Although that wouldn't have actually crashed due to this API change, it would have caught that setting `qsim.instrument.exptime` didn't actually seem to have any effect anymore.

moustakas:
I think the units tests @sbailey has proposed belong in the (non-existent) units tests for ```quickbrick```.  Indeed, I "discovered" this issue precisely because I was finding (by running ```quickbrick```) that longer exposure times were not yielding higher-S/N spectra.

dkirkby:
@sbailey The tests you described would have to be in desisim, not specsim, to catch errors like this.  The corresponding unit tests in specsim don't help because they were updated at the same time as the API.

[...]

PR #169, opened 2016-09-28
added unit tests to quickgen and updated exposure time

rstaten:
This PR

- adds airmass unit test for quickgen
- adds exposure time unit test for quickgen
- updates how exposure time is set with specsim v0.5

Note: bright time vs. dark time unit test is not included since quickgen does not yet include a moon phase option as quickbrick does.

### DIFF
--- a/py/desisim/scripts/quickgen.py
+++ b/py/desisim/scripts/quickgen.py
@@ -369,13 +369,13 @@ def main(args=None):
 
             # Get results for our flux-calibrated output file.
             assert output['observed_flux'].unit == u.erg / (u.cm**2 * u.s * u.Angstrom)
-            cframe_observedflux[j, i, :num_pixels] = output['observed_flux']
-            cframe_ivar[j, i, :num_pixels] = output['flux_inverse_variance']
+            cframe_observedflux[j, i, :num_pixels] = 1e17 * output['observed_flux']
+            cframe_ivar[j, i, :num_pixels] = 1e-34 * output['flux_inverse_variance']
 
             # Use the same noise realization in the cframe and frame, without any
             # additional noise from sky subtraction for now.
             frame_rand_noise[j, i, :num_pixels] = output['random_noise_electrons']
-            cframe_rand_noise[j, i, :num_pixels] = (
+            cframe_rand_noise[j, i, :num_pixels] = 1e17 * (
                 output['flux_calibration'] * output['random_noise_electrons'])
 
             # The sky output file represents a model fit to ~40 sky fibers.
--- a/py/desisim/scripts/quickgen.py
+++ b/py/desisim/scripts/quickgen.py
@@ -165,6 +165,11 @@ def main(args=None):
     print("Initializing SpecSim with config '{0}'".format(args.config))
     qsim = specsim.simulator.Simulator(args.config)
 
+    # explicitly set location on focal plane if needed to support airmass
+    # variations when using specsim v0.5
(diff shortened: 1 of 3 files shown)


sbailey:
Thanks for these tests.  Travis skips the quickgen tests because of the specsim memory problem (desihub/specsim#43), but when I run them on my laptop I get failures, e.g.
```
ERROR: test_quickgen_airmass (desisim.test.test_quickgen.TestQuickgen)
  File "/Users/sbailey/desi/git/desisim/build/lib/desisim/test/test_quickgen.py", line 241, in test_quickgen_airmass
    cf0=desispec.io.readframe('cframe-r0-00000100.fits')
AttributeError: module 'desispec.io' has no attribute 'readframe'

ERROR: test_quickgen_exptime (desisim.test.test_quickgen.TestQuickgen)
  File "/Users/sbailey/desi/git/desisim/build/lib/desisim/test/test_quickgen.py", line 265, in test_quickgen_exptime
    cf0=desispec.io.readframe('cframe-r0-00000100.fits')
AttributeError: module 'desispec.io' has no attribute 'readframe'
```
Please update and confirm that they pass on your machine.

Not necessarily for Ryan but also noting:
* I'm also getting errors from test_quickbrick that are fixed in unmerged PR #166.  We should merge that PR and then merge master into this branch to fix those tests.
* the names of the MWS targeting bits changed in desitarget, thus the current desispec code works with desitarget 0.6.1 but not master.  I'll investigate whether we can work around this by just using bit names in our tests that haven't changed, or whether we do need a new tag of desitarget and require that for desisim prior to this merge.

rstaten:
@sbailey This problem has been fixed. However when I now run these tests I get the following error:

FAIL: test_quickgen_airmass (__main__.TestQuickgen)
  File "test_quickgen.py", line 247, in test_quickgen_airmass
    self.assertLess(np.median(cf0.ivar),np.median(cf1.ivar))
AssertionError: 2.7280001849116604e+34 not less than 2.6549968800909387e+34

It is my understanding that increased airmass should yield smaller ivar, so it now seems that the unit test is doing what it should, but perhaps this error should be addressed.

rstaten:
Is it possible for this PR to be merged? If not, I will make changes based on any comments or suggestions.

rstaten:
@sbailey I have changed the flux of the cframe files by a factor of 1e17 and the ivar by a factor of 1e-34. It seems ok to simply add this factor for the flux, but it seems to be a bit of a kludge to simply add a factor in front of inverse variance. Is this acceptable?

sbailey:
I made a small change to make it compatible with desitarget/master (they removed one of the Milky Way Survey mask bit names that we had previously been using as a placeholder).

The remaining test failure is a genuine one (good job!) — changing the airmass with quickgen doesn't actually impact the sky spectrum.  I added a question to issue #161 for converting quickgen and quickbrick to specsim v0.5 .  Let's see what @dkirkby says.  If this is fixable on the desisim side let's do that before this merge.  If it requires a bug fix on specsim, we could flag this as an expected failure test in the meantime and proceed with merging.

rstaten:
I ran these tests using the updated specsim and no longer get any errors or failed tests.

sbailey:
I've added a small update to quickgen to explicitly set `qsim.source.focal_xy` if needed so that the varying airmass tests work with both specsim master (which sets it already) and specsim v0.5 (which doesn't).  After the tests pass I will merge this PR.

PR #173, opened 2016-10-12
Quickgen

rstaten:
This PR

- Adds keyword --frameonly to quickgen.

- Gives quickgen the option to only output uncalibrated frame files.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -7,6 +7,7 @@ desisim change log
 
 * Fix some ``build_sphinx`` errors.
 * Run coverage tests under Python 2.7 for now.
+* Update template Module file to new DESI+Anaconda infrastructure.
 
 0.14.0 (2016-09-14)
 -------------------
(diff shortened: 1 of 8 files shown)


sbailey:
This is a good feature to add, but this implementation involves a lot of replicated code.  Could you implement this with something like:
```
loop over exposures:
    [simulate frames]
    if frameonly:
        continue
    else:
        [simulate sky, stdstars, cframe, etc.]
```
i.e. the code for simulating a frame should be the exact same code whether or not --frameonly is selected, and then any non-frame code should just be skipped if it isn't needed.

rstaten:
The most recent commit fixes this.

sbailey:
Thanks.  Now it is just 3 lines of code instead of N>>1 new lines, and changing anything in how frames are generated will automatically worth with and without --frameonly.

moustakas:
Very excited to see these kinds of developments, @rstaten, thank you.

rstaten:
@moustakas No problem. Let me know if there is something you would like to see implemented.

PR #176, opened 2016-10-13
Quickgen moon

rstaten:
This PR

- Adds moon phase, moon angle, and zenith angle to quickgen via optional keyword arguments.

- Adds unit tests for each additional moon property.

- Updates changes.rst to include recent quickgen developments.

### DIFF
--- a/py/desisim/scripts/quickbrick.py
+++ b/py/desisim/scripts/quickbrick.py
@@ -79,9 +79,9 @@ def parse(options=None):
   
     # Options corresponding to the bright-time survey only.
     bts_parser = parser.add_argument_group('options for Bright Time Surveys (BGS and MWS)')
-    bts_parser.add_argument('--moon-phase', type=float,  help='moon phase (0=full, 1=new)', default=0.7, metavar='')
-    bts_parser.add_argument('--moon-angle', type=float,  help='separation angle to the moon (0-180 deg)', default=50, metavar='')
-    bts_parser.add_argument('--moon-zenith', type=float,  help='zenith angle of the moon (0-90 deg)', default=30, metavar='')
+    bts_parser.add_argument('--moon-phase', type=float,  help='moon phase (0=full, 1=new)', default=None, metavar='')
+    bts_parser.add_argument('--moon-angle', type=float,  help='separation angle to the moon (0-180 deg)', default=None, metavar='')
+    bts_parser.add_argument('--moon-zenith', type=float,  help='zenith angle of the moon (0-90 deg)', default=None, metavar='')
 
     args = None
     if options is None:
@@ -120,9 +120,18 @@ def main(args):
     log.debug('Using OBJTYPE {}'.format(objtype))
     if objtype == 'BGS' or objtype == 'MWS' or objtype == 'BRIGHT_MIX':
         qsim.observation.exposure_time = desiparams['exptime_bright'] * u.s
-        qsim.atmosphere.moon.moon_zenith = args.moon_zenith * u.deg
-        qsim.atmosphere.moon.separation_angle = args.moon_angle * u.deg
-        qsim.atmosphere.moon.moon_phase = args.moon_phase
+        if args.moon-phase is None:
+            qsim.atmosphere.moon.moon_phase = 0.7
+        else:
+            qsim.atmosphere.moon.moon_phase = args.moon-phase
+        if args.moon-angle is None:
+            qsim.atmosphere.moon.separation_angle = 50 * u.deg
+        else:
+            qsim.atmosphere.moon.separation_angle = args.moon-angle * u.deg
+        if args.moon-zenith is None:
+            qsim.atmosphere.moon.moon_zenith = 30 * u.deg
+        else:
+            qsim.atmosphere.moon.moon_zenith = args.moon-zenith * u.deg
     else:
         qsim.observation.exposure_time = desiparams['exptime_dark'] * u.s
 
--- a/py/desisim/scripts/quickbrick.py
+++ b/py/desisim/scripts/quickbrick.py
@@ -79,9 +79,9 @@ def parse(options=None):
   
     # Options corresponding to the bright-time survey only.
     bts_parser = parser.add_argument_group('options for Bright Time Surveys (BGS and MWS)')
-    bts_parser.add_argument('--moon-phase', type=float,  help='moon phase (0=full, 1=new)', default=None, metavar='')
-    bts_parser.add_argument('--moon-angle', type=float,  help='separation angle to the moon (0-180 deg)', default=None, metavar='')
-    bts_parser.add_argument('--moon-zenith', type=float,  help='zenith angle of the moon (0-90 deg)', default=None, metavar='')
+    bts_parser.add_argument('--moon_phase', type=float,  help='moon phase (0=full, 1=new)', default=None, metavar='')
+    bts_parser.add_argument('--moon_angle', type=float,  help='separation angle to the moon (0-180 deg)', default=None, metavar='')
+    bts_parser.add_argument('--moon_zenith', type=float,  help='zenith angle of the moon (0-90 deg)', default=None, metavar='')
 
     args = None
     if options is None:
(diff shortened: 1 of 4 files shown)


sbailey:
Thanks for adding this feature, Ryan.  Right now the moon parameters are only used for flavor='mws' or 'bgs'.  Could we take this one step further, e.g. to allow setting moon parameters for other types like studying ELGs during gray time?  I suggest:

* change the argparse moon parameter defaults to `None` so that we can tell if the user actually specified anything or not
* If the user does specify moon parameters, use them regardless of what the flavor is.
* If the user does not specify moon parameters (`args.moonphase is None`, etc.), then pick defaults that depend upon the flavor:
  * bgs, mws, bright: use the parameters that you currently have (moonphase=0.7, moonangle=50, moonzenith=30) [note: I haven't checked that these are the default bright time parameters used by the BGS WG...]
  * gray, grey: moonphase = 0.1, moonangle=60, moonzenith=80 [Note: I don't think we have defined typical gray-time moon parameters; these are just first guesses to be refined]
  * other: don't set moon parameters for specsim, i.e. treat the moon as being below the horizon

I think this gives us full flexibility and reasonable defaults for each flavor.  In practice when quickgen gets combined with the surveysim, the exact telescope pointing and moon location and phase will be specified for each exposure (requiring more parameters...)

Does that make sense?

moustakas:
These changes would be great, but could the same changes be made to ```quickbrick``` as well, while you're working on it?

```
bgs, mws, bright: use the parameters that you currently have (moonphase=0.7, moonangle=50, 
moonzenith=30) [note: I haven't checked that these are the default bright time parameters used 
by the BGS WG...]
```
These defaults are fine.

rstaten:
I made these changes to both quickgen and quickbrick. The default specsim values are moonphase = 0.5, moonangle=60, moonzenith=100, which are the values taken if no moon arguments are given.

py/desisim/scripts/quickbrick.py
```diff
@@ -120,9 +120,18 @@ def main(args):
     log.debug('Using OBJTYPE {}'.format(objtype))
     if objtype == 'BGS' or objtype == 'MWS' or objtype == 'BRIGHT_MIX':
         qsim.observation.exposure_time = desiparams['exptime_bright'] * u.s
-        qsim.atmosphere.moon.moon_zenith = args.moon_zenith * u.deg
-        qsim.atmosphere.moon.separation_angle = args.moon_angle * u.deg
-        qsim.atmosphere.moon.moon_phase = args.moon_phase
+        if args.moon-phase is None:
+            qsim.atmosphere.moon.moon_phase = 0.7
+        else:
+            qsim.atmosphere.moon.moon_phase = args.moon-phase
```
sbailey: typo!  `moon-phase` vs. `moon_phase`.  The fact that the tests pass is evidence that you don't have tests for this case.  :)  Your penance shall be writing a test...

py/desisim/scripts/quickbrick.py
```diff
@@ -120,9 +120,18 @@ def main(args):
...
+        if args.moon-angle is None:
+            qsim.atmosphere.moon.separation_angle = 50 * u.deg
+        else:
+            qsim.atmosphere.moon.separation_angle = args.moon-angle * u.deg
+        if args.moon-zenith is None:
+            qsim.atmosphere.moon.moon_zenith = 30 * u.deg
+        else:
+            qsim.atmosphere.moon.moon_zenith = args.moon-zenith * u.deg
     else:
         qsim.observation.exposure_time = desiparams['exptime_dark'] * u.s
```
sbailey: In quickgen you also have cases for gray/grey, which don't appear here.  Please add.

py/desisim/scripts/quickgen.py
```diff
@@ -176,6 +179,34 @@ def main(args=None):
...
+            qsim.atmosphere.moon.moon_phase = args.moonphase
+        if args.moonangle is None:
+            qsim.atmosphere.moon.separation_angle = 60 * u.deg
+        else:
+            qsim.atmosphere.moon.separation_angle = args.moonangle * u.deg
+        if args.moonzenith is None:
+            qsim.atmosphere.moon.moon_zenith = 80 * u.deg
+        else:
+            qsim.atmosphere.moon.moon_zenith = args.moonzenith * u.deg
+
```
sbailey: Let's add an explicit `else` clause here to set the parameters for the moon below the horizon.  That happens to be the current specsim defaults, but it would be better for quickgen and quickbrick to control their own default behavior so that a future change in specsim doesn't accidentally silently change the quickgen/quickbrick action for dark time targets.

py/desisim/scripts/quickbrick.py
rstaten: I can certainly write a test for this, but I'm not exactly sure what that test would entail. I changed the keywords in quickbrick to mimic the parameters of specsim, but that seems to be more of a change with respect to convention, rather than being something intrinsically prone to errors.

sbailey: A quickbrick test with flavor=BGS would catch the typos of `args.moon-phase` instead of `args.moon_phase`.  That is more than just a naming convention: `args.moon-phase` would be trying to subtract two non-existence variables `args.moon` and `phase` and would have crashed.  You have fixed that now; good.

I see that we do have quickbrick tests for flavor='BGS' but Travis was skipping them because memory issues with specsim.  Those have also now been fixed with specsim master, so I'll go ahead and merge this now and we can separately enable the quickbrick and quickgen tests for Travis.

sbailey:
FYI, I made a post-merge update directly to master to fix the quickgen and quickbrick unit tests that were broken due to `--moonphase` vs. `--moon_phase` vs. `--moon-phase`.  Please *always* run `python setup.py test` and confirm that tests pass locally before submitting a pull request (and after making changes to the pull request).  The as-merged tests didn't pass but I didn't notice the bug because Travis skipped them and I had assumed that you had been running them locally to catch typos like these.

BTW, seeing how these failed made me realize what the original `args.moon-phase` vs. `args.moon_phase` confusion was about (I think...).  It is fine to have script options like `--moon-phase`; that might even be preferable to `--moon_phase` (underscores are a pain...).  But python doesn't allow hyphens in a variable name so argparse can't convert those into `args.moon-phase`.  Instead, it converts hyphens to underscores and gives you `args.moon_phase` as the internal variable even if the original option was `--moon-phase`.

So: let's do a new PR to go back to how you originally had the options as `--moon-phase`, `--moon-angle`, `--moon-zenith`: those look better and also match how other options to quickbrick work like `--zrange-lrg`.  But please make sure that the tests + the code all work consistently and the tests pass.  Does that make sense?

rstaten:
Yes, that all makes sense. However, as you mentioned previously, including a hyphen in a keyword seems to cause issues. I can change the keywords to --moonphase, etc. to avoid both underscores and hyphens if you think this is okay. If so, should I go ahead and change the other keywords in quickbrick that include hyphens?

moustakas:
I think the hyphens in the keywords are fine/good (for clarity), and there's code (e.g., the bright-time simulations code I've written) that would break by your proposed change.   Internally, ```argparse.ArgumentParser.parse_args``` converts everything to underscores, so let's stick with the current implementation.

sbailey:
@rstaten: to clarify, hyphens are allowed in command line arguments to scripts, e.g. `--zrange-lrg`, `--moon-phase`, etc..  But they aren't allowed in python variables names, so that becomes `args.zrange_lrg`, `args.moon_phase` internally.  For clarity/consistency with other options, let's update these with hyphens.  The only change should be to lines like
```python
parser.add_argument('--moon_phase', type=float,  help='moon phase (0=full, 1=new)', default=None, metavar='')
```
which become
```
parser.add_argument('--moon-phase', type=float,  help='moon phase (0=full, 1=new)', default=None, metavar='')
```
And the tests will need to be updated to call quickgen/quickbrick with `--moon-phase` instead of `--moon_phase`.  You can probably make those changes faster than I could type this message.  Ping me by email if this is still confusing.

Issue #167, opened 2016-09-22
dedicated integration test for quickgen + spectro pipeline


sbailey:
Context: quickgen uses specsim to simulate the full spectro pipeline output at multiple levels: fiberflats, uncalibrated extracted spectra, sky model, stdstar models, flux calibration vectors, fully calibrated spectra.  In principle the full pipeline should be able to pick up at any of these intermediate stages and continue running as if they had been generated by the full pipeline instead of quickgen.

quickgen has unit tests but we've occasionally had problems with it outputting files that can't be used as input to the next pipeline step.  e.g. the last case was when the pipeline started requiring the the fibermap be included in the frame files, but we forgot to add that to quickgen too (that has been fixed).

In anticipation of using quickgen more and more, please write a dedicated quickgen integration test to make sure that its outputs still work as pipeline inputs.  e.g. something like:

* Use newexp-desi to generate a simspec file with ~25 spectra
* Run quickgen on those to generate frame, fiberflat, sky, stdstars, calib, and cframe files
* Run the desispec pipeline steps to confirm that they work with the quickgen files as inputs, e.g.
  * desi_compute_fiberflat (frame -> fiberflat)
  * desi_compute_sky (frame & fiberflat -> sky)
  * desi_fit_stdstars (frame & fiberflat & sky -> stdstars)
  * desi_compute_fluxcalibration (frame & fiberflat & sky & stdstars -> fluxcalib)
  * desi_process_exposure (everything -> cframe)
  * desi_make_bricks (cframe -> bricks)

This would probably take long enough that we wouldn't run it as part of the unit tests but we'd run it as a separate test similar to (but separate from) py/desispec/test/integration_test.py for pixsim+spectro pipeline.

rstaten:
@sbailey I have written this integration test and I will push it to a branch soon. The only issue that arrises when sending quickgen output through the full pipeline is due to astype with respect to the output wavelength array in the quickgen calib file. I can add .astype('f4') to line 91 in desispec.io.fluxcalibration to fix this, but if I do not then I get an error when running desi_process_exposure.

Also, I do not include the first step that you mention (frame -> fiberflat) because quickgen only outputs fiberflat files for a flat exposure and skips the process of outputting frame files. If this needs to be tested as well, then I can update quickgen to allow for frame files as output for flat exposures.

rstaten:
[...]

PR #179, opened 2016-10-20
add integration test for quickgen

rstaten:
This PR

- Provides an integration test for quickgen which verifies that quickgen output can be used as input for the full pipeline.

The integration test:

- Checks that all necessary environment variables are set.
- Generates 25 spectra flat, arc, and dark exposures via newexp-desi, then simulates this data using quickgen.
- Uses quickgen frame and fiberflat files as input for desi_compute_sky.
- Uses quickgen frame, fiberflat, and sky files, and star templates as input for desi_fit_stdstars.
- Uses quickgen frame, fiberflat, and sky files, and stdstars file generated by previous step as input for desi_compute_fluxcalibration.
- Uses quickgen frame, fiberflat, sky, and calib files as input for desi_process_exposure.
- Runs desi_make_bricks.

Note: desi_process_exposure fails due to 32 bit vs. 64 bit wavelength arrays generated by desispec.io.frame vs. desispec.io.fluxcalibration. Changing fluxcalib.wave to fluxcalib.wave.astype('f4') in line 91 of desispec.io.fluxcalibration seems to solve this issue.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,9 +5,9 @@ desisim change log
 0.15.1 (unreleased)
 -------------------
 
-* No changes yet
+* add integration test for quickgen
 
-0.15.0 (2016-10-14)
+0.15.0 (2016-10-20)
 -------------------
 
 * add moon phase, moon angle, and zenith angle to quickgen
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,35 +5,9 @@ desisim change log
 0.15.1 (unreleased)
 -------------------
 
-* add integration test for quickgen
+* Add integration test for quickgen
 
-0.15.0 (2016-10-20)
--------------------
-
-* add moon phase, moon angle, and zenith angle to quickgen
-* add a unit test for each moon property
-
-0.14.0 (2016-10-13)
--------------------
-
-* change $PRODNAME to $SPECPROD in quickgen
-* change print statement to log.info() in quickgen
-* change os.path.join to desispec.io.findfile in quickgen
-
-0.14.0 (2016-10-12)
--------------------
-
-* add keyword frameonly to quickgen
-* allow for only uncalibrated frame files to be output by quickgen
-
-0.14.0 (2016-10-12)
--------------------
-
-* add airmass unit test for quickgen
-* add exposure time unit test for quickgen
-* update how exposure time is set in quickgen with specsim v0.5
-
-0.14.0 (2016-09-28)
+0.15.0 (2016-10-14)
 -------------------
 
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks for this quickgen integration test.  It already caught the genuine problem of whether the fluxcalibration wavelength should be written as 'f4' or 'f8'.  I updated desispec master to fix the wavelength as 'f4' vs. 'f8', as well as fixing some BUNIT syntax (e.g. 'A' vs. 'Angstrom'). I thought I had included that with desihub/desispec#284 but I guess not...).  I'll admit that I broke my own rules and forgot to run tests before committing and got caught by Travis and had to do a second commit for it really to work...

Please update the integration test to use `desispec.io.findfile` instead of creating the paths by hand.  e.g.
```python
framefile = desispec.io.findfile('frame', night, expid, camera)
fiberflatfile = desispec.io.findfile('fiberflat', night, flat_expid, camera)
skyfile = desispec.io.findfile('sky', night, expid, camera) + '.test'
com = "desi_compute_sky --infile {} --fiberflat {} --outfile {}".format(framefile, fiberflatfile, skyfile)
```
Note that `expid` and `flat_expid` should be integers instead of zero-padded strings for using `desispec.io.findfile`.

This highlights a misunderstanding in the currently code, which constructed paths starting with `$DESI_SPECTRO_REDUX/exposures/` instead of `$DESI_SPECTRO_REDUX/$SPECPROD/exposures/`.  Using `desispec.io.findfile` is supposed to save you from that kind of problem.

rstaten:
Should I also remove all directories/files generated by this test? It doesn't seem that the pixsim integration test does this, which is why I currently do not include this feature.

sbailey:
A few small things:
* expid and flat_expid should not be zero padded.  Just plain integers 0 and 2.
* in changes.rst, you should never change the date of the tag.  The tag is a historical event on a certain date and the bullet list description below that describes what changes went into that tag.
New changes to into the new unreleased undated tag at the top.  In some cases we may also retroactively update the description of a tag to include something that we forgot to comment on previously.  But don't change its date.

Like the desisim integration test, let's *not* automatically remove files at the end of this test.  Unit tests should cleanup after themselves, but for these larger tests it can be useful to inspect the files afterward.

rstaten:
These last changes should have fully fixed issue #167. Ok to merge?

sbailey:
There was a missing `import sys` that was needed in the case that the necessary environment variables were not set.  I just added that.  Otherwise this looks good and I will merge it as soon as the Travis tests pass again.

Issue #180, opened 2016-10-25
comparison of quickgen and quickbrick


rstaten:
Following @rkehoe discussion with @sbailey, this issue itemizes observed differences between quickgen and quickbrick to facilitate generation of common  code as raised by @dkirkby (eg. #178).  The main difference comes from the fact that quickgen uses files generated by newexp-desi as input for specsim, while quickbrick goes directly from templates to specsim simulation.

Relevant similarities include:
- Using specsim as a simulator for all three channels
- Taking moon parameters for bright time (quickgen also includes gray time)
- Outputting calibrated frame files

Relevant differences include:
- Quickgen uses a fibermap file generated by newexp-desi while quickbrick generates an empty fibermap file
- Quickgen obtains parameters such as airmass, exposure time, etc. from simspec files while quickbrick takes these parameters as input arguments
- In addition to calibrated files, quickgen also outputs sky, calibration, and uncalibrated frame files

Options for development include 1) leaving as is, 2) they could be combined by either essentially having if/else statements for quickgen vs. quickbrick, or 3) or by having input files for one (eg. quickbrick) transformed in code to the other (eg. make a mock simspec file from templates plus user parameters).   The latter option would potentially maximize reused code and allow new possibilities such as quickbrick running on simspec files or quickgen outputting to bricks on configuration.

moustakas:
I very much support consolidating the functionality of these two scripts (i.e., either option 2 or 3, with option 3 being the most flexible but also the most amount of work).  In particular, I'll note that neither ```quickbrick``` nor ```quickgen``` (via ```newexp-desi```) currently allows the user to vary many/most of the myriad optional inputs to the template-generating code (for example, suppressing the color-cuts applied or changing the input apparent magnitude priors), which means that when these options are added they'll need to be added to *both* scripts --- unless those scripts are consolidated as proposed, of course.

rkehoe:
Good to know of the potential downstream issues.  @rstaten has already a draft of 2) and it seems to suggest 3).  I favor option 3), but it needs to be done in such a way as to preserve the ease-of-use for the WGs that motivated quickbrick to begin with.

rstaten:
I am also in favor of option 3. As @rkehoe mentioned, I have already written a working version of option 2, but it is very redundant. Both scripts follow essentially the same steps, so an if/else version wasn't difficult to produce, but I am starting to look into how they can perform these steps in unison. This would of course also include a refactoring of the unit tests and recently added integration test.

[...]

PR #184, opened 2016-11-04
Quickgen refactor

rstaten:
I have moved forward with the refactor of quickgen/quickbrick in order to consolidate common code, as suggested in #180.

This PR:

- Refactors quickgen to include quickbrick functionality, allowing the user to either provide 
simspec and fibermap files or brick name to perform simulations.
  - Usage of quickgen to incorporate bricks is identical to quickbrick: i.e. 'quickbrick -b brickname' becomes 'quickgen -b brickname'.
  - Other command line options of quickbrick are included in quickgen.
- Updates quickgen unit tests to include tests for brick options.

Note: This implementation was natural following the refactor of quickgen/quickbrick code. If one script for quickgen and quickbrick use is not the preferred refactor, other options can be pursued. If quickbrick needs to remain a command line option, then a refactor that includes common code, while still allowing both quickgen and quickbrick to be given by the user, can be provided.


sbailey:
I'm not sure this refactor is really accomplishing the goal of having quickgen and quickbrick share the same code.  Although they are now within the same `desisim.scripts.quickgen.main` function and share a single call to `qsim.simulate()`, the actual code is mostly split up by blocks of:
```python
if args.simspec:
    # do something
elif args.brickname:
    # do something very similar but slightly different
```
We're trying to get rid of duplicated code like
```python
    if args.simspec:
        if simspec.flavor in bright_objects:
            if args.moon_phase is None:
                qsim.atmosphere.moon.moon_phase = 0.7
            else:
                qsim.atmosphere.moon.moon_phase = args.moon_phase
        #...
```
vs.
```python
    elif args.brickname:
        if objtype in bright_objects:
            if args.moon_phase is None:
                qsim.atmosphere.moon.moon_phase = 0.7
            else:
                qsim.atmosphere.moon.moon_phase = args.moon_phase
        #...
```

I'm looking for a deeper refactor that would accomplish something like:
```python
if args.simspec:
    # read inputs from simspec file
else:
    # create inputs on the fly...

# Do stuff that is common to both quickgen and quickbrick without a bunch
# of if/then blocks for the two cases

if args.brickname:
    # write a brick file with all the calibrated spectra and exit
else:
    # write frame, sky, fluxcalib, cframe, etc.
```
Let's keep working on this...

rstaten:
@sbailey The most recent push should encompass more of what you are looking for in the refactor. However, when running the updated unit tests, the moon phase test is failing when using brick files and I am still looking into this before a potential merge.

rstaten:
The unit tests now all pass, and this can potentially be merged if it is the desired refactor.

py/desisim/scripts/quickgen.py
```diff
@@ -412,10 +508,21 @@ def main(args=None):
...
             cframe_observedflux[j, i, :num_pixels] = 1e17 * output['observed_flux']
             cframe_ivar[j, i, :num_pixels] = 1e-34 * output['flux_inverse_variance']
 
+            # Fill brick arrays from the results.
+            if i==0:
+                camera = 'b'
+            elif i==1:
+                camera = 'r'
+            else:
+                camera = 'z'
```
sbailey: I think the qsim.camera_output results are not guaranteed to be in the order b, r, z.  There must be a better way to derive their name.

py/desisim/scripts/quickgen.py
```diff
@@ -1,109 +1,127 @@
...
+    parser.add_argument('--zrange-lrg', type=float, default=(0.5, 1.1), nargs=2, metavar='',
+                        help='minimum and maximum redshift range for LRG')
+    parser.add_argument('--zrange-bgs', type=float, default=(0.01, 0.4), nargs=2, metavar='',
+                        help='minimum and maximum redshift range for BGS')
+    parser.add_argument('--rmagrange-bgs', type=float, default=(15.0, 19.5), nargs=2, metavar='',
+                        help='Minimum and maximum BGS r-band (AB) magnitude range')
+    parser.add_argument('--sne-rfluxratiorange', type=float, default=(0.1, 1.0), nargs=2, metavar='',
+                        help='r-band flux ratio of the SNeIa spectrum relative to the galaxy')
+    parser.add_argument('--add-SNeIa', action='store_true', help='include SNeIa spectra')
+
```
sbailey: Please add some comment documentation about:
* how to run this in "quickgen mode", outputting frame files, cframe files, etc.
* how to run this in "quickbrick mode", outputting only bricks
* which options apply to only one or the other vs. both

rstaten:
@sbailey These changes have been implemented, so this refactor should be ready to be merged. For the record, I ran the command discussed in issue #312 on this refactored version, but did not get the reported error messages.

sbailey:
I updated the camera name code to get it directly from the specsim results instead of mapping the mean wavelength to a name:
```python
camera = output.meta['name']
```
PR #189 changed `test_quickbrick.py`, but those tests have been moved to `test_quickgen.py` in this PR, causing a merge conflict with master.  I ported the updated test into `test_quickgen.py` in this branch.  After confirming that the Travis tests still pass, I'll take care of the command line merge to resolve the conflicts.

PR #185, opened 2016-11-04
testing Travis and memory

sbailey:
This is a fake PR to trigger travis tests to try to figure out some memory issues.  Don't merge yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ os:
     # - osx
 
 # sudo option affects which type of container is available
-sudo: false
+sudo: required
 
 # The apt packages below are needed for sphinx builds, which can no longer
 # be installed with sudo apt-get.
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ os:
     - linux
     # - osx
 
-# Setting sudo to false opts in to Travis-CI container-based builds.
-sudo: true
+# sudo option affects which type of container is available
+sudo: false
 
 # The apt packages below are needed for sphinx builds, which can no longer
 # be installed with sudo apt-get.
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ os:
     # - osx
 
 # Setting sudo to false opts in to Travis-CI container-based builds.
-sudo: false
+sudo: true
 
 # The apt packages below are needed for sphinx builds, which can no longer
 # be installed with sudo apt-get.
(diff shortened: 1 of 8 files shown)


sbailey:
Tests hung under python 2.7 (but not 3.5) after re-enabling quickgen and quickbrick, and only for the "push" version of the tests but not the "pr" version.  `/usr/bin/time -lp` reports 332 MB memory used by test_quickgen.py, 350 MB for test_quickbrick.py, and 490 MB for the full `python setup.py test` suite.  Maybe this is something different from a memory issue?  I'm not sure what our best route is to get back to stably working Travis tests.

sbailey:
rerunning the failed travis test without making any code changes passed.  Perhaps Travis is having a bad day...

sbailey:
A bit more digging.  The problem does seem to be caused by some change on the Travis side.  I contacted their technical support at the end of Friday but haven't heard back yet.  Some details for the record:

This branch started with master 2dd0b72, which passed Travis tests on November 3:

https://travis-ci.org/desihub/desisim/builds/173054282

After a one-line documentation change to produce branch test_travis commit f51da91, Travis tests timed out:

https://travis-ci.org/desihub/desisim/builds/173350004

Confirming the diff:
```
[master*] desisim $ git diff 2dd0b72..f51da91
diff --git a/doc/changes.rst b/doc/changes.rst
index 01981fe..aa03a7a 100644
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -6,6 +6,7 @@ desisim change log
 -------------------
 
 * Add integration test for quickgen
+* Tuning memory for Travis tests
 
 0.15.0 (2016-10-14)
 -------------------
```
Note that this wasn't the first time the Travis tests became problematic; I created this branch and PR while trying to isolate the Travis tests failures in PR #156, but since that has real code changes it was harder to isolate whether the new code or delayed merge was causing problems, or Travis itself.  I now blame Travis.

sbailey:
closing vestigial test PR without merging.

rainwoodman:
There is a button on the travis log page to rerun the test suite, I sometimes click on that if the failure smells like it is due to travis.

sbailey:
Yes, I used that button to retry tests.  I forgot to add some other close notes:

Moving to the sudo-enabled test environment gives a larger memory footprint and our tests work, at the expense of slower startup times for the tests.  My memory profiling of our tests indicate that they should have fit within the 4 GB of the faster non-sudo environments.

Travis tech support has offered to open up a backdoor so that we can ssh into the test virtual machine while the tests are running to poke around if we would fine that useful.  Now that tests are running again this seems low priority, but we should keep that possibility in mind if we need it in the future.

PR #189, opened 2016-11-17
fix tests to work with latest desitarget master

sbailey:
desihub/desitarget#113 split `desitarget.cuts.isQSO` into `isQSO_colors` and `isQSO_cuts`.  That broke desisim without us noticing.  This PR fixes that, while still supporting tag 0.7.0 and prior that have the `isQSO` function (while printing an error message about needing to upgrade).

It also fixes a problem where the lyman alpha test data wasn't installed by setup.py; I'm not sure how we passed tests previously without this.

I need this fix to get other tests to pass on an unrelated upcoming PR so I will self merge this in an hour if there are no objections.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -2,10 +2,10 @@
 desisim change log
 ==================
 
-0.17.0 (unreleased)
+0.16.1 (unreleased)
 -------------------
 
-* No changes yet
+* fixes tests for use with latest desitarget master
 
 0.16.0 (2016-11-10)
 -------------------
--- a/py/desisim/templates.py
+++ b/py/desisim/templates.py
@@ -1566,7 +1566,12 @@ class QSO():
         self.objtype = 'QSO'
 
         if colorcuts_function is None:
-            from desitarget.cuts import isQSO as colorcuts_function
+            try:
+                from desitarget.cuts import isQSO_colors as colorcuts_function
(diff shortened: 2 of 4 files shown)


moustakas:
I was fixing this on an uncommitted branch but got side-tracked. Sorry.  Looks good to merge.

sbailey:
I'm getting intermittent failures of test_quickbrick where the observed error vs. flux appears to have changed.  I'm going to chase that a bit more before merging...

sbailey:
Well, I got the failure twice on Travis and once on my laptop this afternoon.  This evening I haven't been able to reproduce it in 20 runs of the unit tests.  I added additional info so that if it fails we'll have exact command including the random seed in the log so that we can reproduce the failure to study more.  After these tests pass (will they?) I'll merge this, and spawn a new ticket the next time we catch this happening.

sbailey:
test_quickbrick_options fails when using seed=784077896 .  That produces some templates that are especially bright in blue and beyond the validity range of the test coefficients describing error vs. flux.  i.e. it is the test that is wrong, not the template or quickbrick code.  I'll update tomorrow with a more robust solution.

sbailey:
I updated the coefficients and filter out objects that randomly are simulated above their range of validity.  Merging now.

PR #190, opened 2016-11-19
Add observing conditions to quicksurvey

sbailey:
This PR integrates quicksurvey with surveysim.  The `obslist_all.fits` file from surveysim is now an optional input for quicksurvey.  If provided, it will be used to define the observing conditions for each tile, to be passed to quickcat.  If not provided, this PR also updates the default observing conditions to use the airmass and E(B-V) from `desi-tiles.fits`.

Internally, quicksurvey was refactored to build the list of what tiles go
with what epochs at the very start and then subsequent simulation code uses
those lists.  The list of tiles per epoch can come from either the original method of
`epoch*.txt` files, or by providing both `obslist_all.fits` and the list of dates
when fiberassign should be rerun (as one YEARMMDD or YEAR-MM-DD per line of a text file).

When using `obslist_all.fits` + epochs files, it checks that the tiles in the epochs files are present in `obslist_all.fits` but does not require their order or grouping to make sense based upon the dates in `obslist_all.fits`.  This is a feature, not a bug -- it might be useful to test different epoch orderings while still using the same random observing conditions from a surveysim run.  The standard usage, however, would be to provide `obslist_all.fits` and the list of dates for running fiberassign.

This has been tested with dark tiles and surveysim, but I have not yet finished testing it with the latest surveysim that also has bright tiles.

Note: surveysim is still run as a separate step before running quicksurvey.  quicksurvey+fiberassign+quickcat affect what targets will be observed when we get to a tile, but doesn't actually change what tiles get observed in what order.


sbailey:
Now that Travis tests are passing, this is ready for final review before merging.

sbailey:
Heads up on this branch and PR: merging #191 (quickcat ELG model) generated merge conflict for this.  I'm going to rebase this branch as part of cleaning it up for merging.

PR #191, opened 2016-11-22
Model quickcat for ELGs

cballand:
This is a first implementation of the efficiency and error models for ELGs in quickcat.
I have tested get_redshift_efficiency() and get_observed_redshifts() using the 5000 ELG training sample of zdc1 as an input, from which I created a flux dict containing OII fluxes. I turned obsconditions off in get_redshift_efficiency to test the efficiency from fluxes only. I have not tested the full quickcat yet.


sbailey:
I fixed the tests so that it would test ELGs in quickcat, and then fixed a few bugs found via that.  I refactored how the targets and truth information is passed into quickcat and the underlying functions and update the I/O to use `np.loadtxt` instead of a custom ascii file reader.

I think this is ready to go, except that it will break quicksurvey since that doesn't yet fill in the OIIFLUX truth and the DECAM_FLUX photometry.  I'm going to go fill that in before merging this so that we can try to maintain a fully working simulation chain.

cballand:
I implemented obs conditions for ELGs by varying the OII flux threshold as function of the observation conditions params rather than directly altering the efficiency. I tuned the parametrization so that resulting efficiency variations are consistent with Kyle's predictions. I commented the previous implementation so that it is easy to swithch back to it if prefered.

sbailey:
Test failed because no ZWARN flags were set even for the `perfect=False` case.  It appears that that is because the test objects are bright enough and the efficiencies high enough that failure should be rare.  Lowering the minimum [OII] flux in test_quickcat.py line 71 to
```python
        truth['OIIFLUX'][isELG] = np.random.normal(2e-17, 2e-17, size=nELG).clip(0)
```
Allowed the tests to pass (i.e. ZWARN was sometimes set as expected), but it generated the concerning warnings:
```
/Users/sbailey/desi/git/desisim/py/desisim/quickcat.py:259: RuntimeWarning: invalid value encountered in true_divide
  simulated_eff = eff_model(true_oii_flux/oii_flux_threshold,sigma_fudge,max_efficiency)
/Users/sbailey/desi/git/desisim/py/desisim/quickcat.py:501: RuntimeWarning: invalid value encountered in less
  jj = r > z_eff
```
Some NaNs or divide by zeros or something are floating around in there.

sbailey:
I restored the observing conditions model from Kyle: the terms were tuned as a correction factor to be multiplied by the overall redshift efficiency, not a correction to the [OII] flux threshold, etc.  We'll still want to continue vetting the redshift efficiency model, but I'm going to merge this now so that I can proceed with resolving conflicts and merging #190 (obsconditions input from surveysim) before we get too much further out of sync.

cballand:
Thanks for putting this together. This is fine, although it sounds more physical to directly alter the OII flux threshold rather than the efficiency _a posteriori_ as the OII flux threshold is equivalent to a signal-to-noise ratio that should be directly impacted by obs conditions. Moreover, multiplying the overall efficiency by a correction factor that may exceed one can potentially leads to efficiencies greater than one. Let's talk on how to improve the model at OSU next week.

sbailey:
I agree that it would be more physical to parameterize how the observing conditions impact S/N so that they have a more direct connection to the parameterization of how the redshift efficiency depends upon S/N.  But that isn't what Kyle's parameterization is — applying it to the [OII] flux threshold doesn't produce the variations in efficiency that he intended.  Efficiencies greater than 1 are effectively clipped to 1.

After we have the more realistic spectral simulations (e.g. chromatic distortions, varying fiber size, and impact of varying seeing), we could use those to make an updated quickcat model based upon varying S/N instead of ad-hoc post-corrections to the efficiencies.  But we need those more realistic spectral simulations first to be able to tune what the updated quickcat model should be.

PR #199, opened 2016-12-22
quicker quickcat maybe

sbailey:
This PR saves memory in quickcat by dropping some unused columns in the input catalogs; this might make it somewhat quicker.

It also fixes a bug when resuming quicksurvey at a given epoch after several epochs had previously been run (it reads in the previous MTL and zcat instead of treating them as None).

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,7 @@ desisim change log
 0.17.2 (unreleased)
 -------------------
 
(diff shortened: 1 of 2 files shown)


sbailey:
I forgot this wasn't merged yet; this branch is what was used for the December 2016 quicksurvey data challenge.  merging now.

Issue #200, opened 2016-12-26
write some code to deal with the DB white dwarf models


moustakas:
The white dwarf models now include both DB and DA white dwarfs (as indicated by the ```WDTYPE``` tag in the metadata table).  Need to write some code to generate the requested white dwarf type.  See also https://github.com/desihub/desisim-testdata/pull/6.

sbailey:
For consideration: the redshift fitting outputs include a "SUBTYPE" column.  Its exact usage remains to be defined, but it is a placeholder for things like stellar classifications wanting to record a subtype while QSO classifications might want to record a subtype from a different set of options.

Should/could template metadata do the same? — a catch-all SUBTYPE column whose options depend upon what type of template it is, rather than carrying a WDTYPE column that is meaningful for only a small fraction of the overall targets.

moustakas:
Implementing this would require a somewhat annoying regeneration of the metadata tables (at least for white dwarfs and stars).  Could we just do this on-the-fly as needed when the template-generating code gets called, e.g., in ```newexp-desi```?  We do something like this (i.e., picking and choosing what we want from the template metadata tables and renaming some columns) anyway...

moustakas:
Addressed in #213.

PR #204, opened 2017-01-18
added wd templates

rstaten:
This PR:

- Adds white dwarfs to the quickgen brick mode.


moustakas:
Thanks for this.  Note that the newest DB white dwarf models are not yet supported (see #200) but this was still needed.

PR #208, opened 2017-02-03
add option for restframe templates

sbailey:
This PR adds an option for generating restframe templates for subclasses of SUPERSTAR and GALAXY (but not yet QSO).


moustakas:
Nice "which --> that" catch!

Issue #207, opened 2017-02-02
speeding up desisim.lya_spectra


forero:
Running [`desisim.lya_spectra`](https://github.com/desihub/desisim/blob/master/py/desisim/lya_spectra.py) to generate LyaQSO spectra takes on average ~2 seconds per spectrum (~1 month for the whole set of objects in the mocks!) on a login node.  Most of the time is spent in [`qso.make_templates`](https://github.com/desihub/desisim/blob/master/py/desisim/templates.py#L1601). What would be the best strategy to speed it up?

sbailey:
Was this timing test done for a single spectrum, or requesting N>>1 and dividing the time by N?

If we can't speed up the individual operations, parallelizing over bricks (or even within bricks and recombining) would be a brute force option for shortening the walltime.

moustakas:
Echoing @sbailey's question, ```qso.make_templates``` is vectorized and should be able to generate thousands of spectra relatively quickly (although I haven't formally timed the code). So can you provide a few more details about how you tested this?

forero:
I tested the code generating 1000 spectra at a time, running from ipython in a login node at NERSC, using %timeit. The 2sec/spectra come from getting the best timing and dividing by N=1000.

forero:
any hints on how to speed up the code @sbailey @moustakas ?

moustakas:
I'm sorry I have not had time to look into this.  The core algorithm was written by @profxj but then I modified it to accept input redshift and magnitude arrays.  Could you profile the code to see where specifically in ```qso.make_templates``` it's spending most of its time?

forero:
@moustakas @sbailey cProfile says that 70% of the time is spent here https://github.com/desihub/desisim/blob/master/py/desisim/qso_template/desi_qso_templ.py#L348 doing cosmology calculations. I will explore some solutions tomorrow.

sbailey:
Adding a @profxj mention as the original author of the code that uses that.

We could make an approximate lookback_distance vs. z spline or polynomial and replace all these calculations with a much faster evaluation of those.

profxj:
I've taken a look at this and have some thoughts.

Let me start from the code I wrote: desi_qso_templates()
  This module was written to generate N_perz mock spectra per redshift interval with the default
zmnx=(0.4, 4) and stepsize z_wind=0.2.  Someone (probably me) has modified it so that
it may create a spectra at any input redshift (or redshifts).

[...]

PR #210, opened 2017-02-14
Lyaspeedup

forero:
Solves #207.

I did some benchmarking using this script (called `tt.py`)

```python
import desisim.lya_spectra
import argparse
parser = argparse.ArgumentParser()

parser.add_argument("nqso", type=int)
args = parser.parse_args()
print(args.nqso)
a, b, c = desisim.lya_spectra.get_spectra("/global/homes/f/forero/mocks/simpleSpec_0.fits.gz", nqso=args.nqso)
```

Using the old version I got
```bash
$time python tt.py 100

real	2m42.291s
user	16m30.520s
sys	1m23.836s

$time python tt.py 1000

real	26m7.781s
user	151m19.616s
sys	10m39.264s
```

Using the new version I get

```bash
$time python tt.py 100

real	0m23.315s
user	5m26.872s
sys	0m21.212s

$time python tt.py 1000

real	2m1.740s
user	50m55.360s
sys	3m11.264s
```

### DIFF
--- a/py/desisim/lya_spectra.py
+++ b/py/desisim/lya_spectra.py
@@ -3,6 +3,7 @@ from desisim.io import empty_metatable
 import scipy as sp
 from scipy import interpolate
 from speclite import filters
+import numpy as np
 
 def get_spectra(infile, first=0, nqso=None, seed=None):
  
@@ -46,24 +47,30 @@ def get_spectra(infile, first=0, nqso=None, seed=None):
     rand = sp.random.RandomState(seed)
     seed = rand.randint(2**32, size=nqso)
 
-    input_meta = empty_metatable(objtype='QSO', nmodel=nqso)
+    input_meta = empty_metatable(objtype='QSO', nmodel=1)
 
     filter_name = 'sdss2010-g'
 
     normfilt = filters.load_filters(filter_name)
     qso = QSO(normfilter=filter_name)
 
-    input_meta["REDSHIFT"]=zqso
-    input_meta["MAG"]=mag_g
-    input_meta["SEED"]=seed
-
-    flux, wave, meta = qso.make_templates(input_meta=input_meta)
-    
-    # Add RA,DEC to output meta
-    meta['RA'] = ra
-    meta['DEC'] = dec
+    flux = np.zeros([nqso, len(qso.wave)])
+    meta = {}
 
     for i,head in enumerate(h[first+1:first+1+nqso]):
+        input_meta["REDSHIFT"] = zqso[i]
+        input_meta["MAG"] = mag_g[i]
+        input_meta["SEED"] = seed[i]
+
+        f, wave, meta_qso = qso.make_templates(input_meta=input_meta)
+        ##append the meta information
+        try:
+            for k in meta_qso.keys():
+                meta[k] = np.append(all_meta[k], meta_qso[k])
+        except:
+            for k in meta_qso.keys():
+                meta[k] = meta_qso[k]
+
         ## read lambda and forest transmission
         la = head["LAMBDA"][:]
         tr = head["FLUX"][:]
@@ -73,13 +80,19 @@ def get_spectra(infile, first=0, nqso=None, seed=None):
         ## will interpolate the transmission at the spectral wavelengths, 
         ## if outside the forest, the transmission is 1
         itr=interpolate.interp1d(la,tr,bounds_error=False,fill_value=1)
-        flux[i,:]*=itr(wave)
-        padflux, padwave = normfilt.pad_spectrum(flux[i, :], wave, method='edge')
+        f *= itr(wave)
+        padflux, padwave = normfilt.pad_spectrum(f, wave, method='edge')
         normmaggies = sp.array(normfilt.get_ab_maggies(padflux, padwave, 
                                mask_invalid=True)[filter_name])
-        flux[i, :] *= 10**(-0.4*input_meta['MAG'][i]) / normmaggies
+        f *= 10**(-0.4*input_meta['MAG'][0]) / normmaggies
+        flux[i,:] = f.copy()
 
     h.close()
+
+    # Add RA,DEC to output meta
+    meta['RA'] = ra
+    meta['DEC'] = dec
+    
     return flux,wave,meta
 
 
--- a/py/desisim/templates.py
+++ b/py/desisim/templates.py
@@ -1704,6 +1704,7 @@ class QSO():
             mag = input_meta['MAG'].data
 
             meta = empty_metatable(nmodel, self.objtype)
+
         else:
             meta = empty_metatable(nmodel, self.objtype)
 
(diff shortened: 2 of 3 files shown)


forero:
Fixed all comments by @sbailey.
The running time went up a little bit, but it's still pretty good (0.22 sec/spectrum)

```bash
$time python tt.py 100
real	0m26.526s
user	6m26.940s
sys	0m23.844s

$time python tt.py 1000
real	2m41.130s
user	67m17.408s
sys	3m57.004s
```

forero:
I've updated the TEMPLATEID to start at `first+1`, corresponding to the header unit in the lya mock. The lya mocks don't have any ID we can add. 

Should we also add a `filename` field in the `meta` Table?

I guess that for `templates.QSO.make_templates(input_meta=meta)` to recover the QSO you should pass `meta` in the same way the spectra were built, i.e. row by row. Not as a bulk `meta` Table. Building them row by row is what gives us the factor 10 speedup.

I've included two new comments in `doc/changes.rst`

sbailey:
Thinking aloud about options for how to trace the LyA mocks back to their input files:

Adding a filename column to the meta table will results in a large number of repeated strings that seems wasteful.  We could keep the filename in the metadata of the metadata table (astropy Tables have a `.meta` OrderedDict for this purpose), but that then would prevent stacking tables from different input files without some extra processing to resolve the metadata conflicts.

Or we could keep the path that should be shared by everyone in the metadata header, while keeping a column of just the filename, or just the integer of the "simpleSpec_0.fits.gz" filename, though that becomes constraining on how the files are named.  Or keep an integer that means "index of the file used in that directory when sorted alphabetically".  If the LyA mock releases are really locked down that would work, but becomes fragile if someone does something like add "README.txt" file.

Or keep a mapping like the `map_id_filename.txt` files in the quicksurvey data challenge.

It would be helpful if we assigned mocks a range of integers to use, and every mock provided unique MOCKIDs even if they span multiple files.

Other ideas?

forero:
Keeping `map_id_filename.txt` is certainly the easiest because it's already implemented and nothing new should be added to the `lya_spectra` routine. I like that one, of course.

sbailey:
I confirmed that passing the meta row-by-row to `QSO.make_templates(input_meta=...)` works to recover the pre-absorption continuum up to a scale factor.

Mapping the output back to the input mocks will still require some wrapper code to update template ids and make a `map_id_filename.txt` or equivalent.  That seems messy but I don't have a better suggestion right now.

I'm ok to merge this, but I'd like to get @moustakas opinion on this mapping too.  We could also merge this as an improvement in speed, and deal with mapping in a separate set of changes.

moustakas:
I'm traveling but I should have a chance to review this in the next couple days.  Thanks. --JM

py/desisim/lya_spectra.py
```diff
@@ -3,8 +3,10 @@
 import scipy as sp
 from scipy import interpolate
 from speclite import filters
+import numpy as np
+import astropy.table 
 
-def get_spectra(infile, first=0, nqso=None, seed=None):
+def get_spectra(infile, first=0, nqso=None, seed=20):
```
moustakas: Why is the `seed` set to 20 rather than `None`?

py/desisim/templates.py
```diff
@@ -1671,7 +1671,7 @@ def make_templates(self, nmodel=100, zrange=(0.5, 4.0), rmagrange=(21.0, 23.0),
             ignored.
           nocolorcuts (bool, optional): Do not apply the fiducial rzW1W2 color-cuts
             cuts (default False).
-
+          N_perz (int, optional): Number of templatex per redshift bin or redshift value.
```
moustakas: Type (templatex --> templates).

py/desisim/lya_spectra.py
```diff
@@ -46,40 +48,45 @@ def get_spectra(infile, first=0, nqso=None, seed=None):
...
-    input_meta["SEED"]=seed
-
-    flux, wave, meta = qso.make_templates(input_meta=input_meta)
-    
-    # Add RA,DEC to output meta
-    meta['RA'] = ra
-    meta['DEC'] = dec
-
+    flux = np.zeros([nqso, len(qso.wave)])
+    meta = None
```
moustakas: Optionally, `flux` could be explicitly initialized to `f4` precision.

py/desisim/lya_spectra.py
```diff
@@ -46,40 +48,45 @@ def get_spectra(infile, first=0, nqso=None, seed=None):
...
     for i,head in enumerate(h[first+1:first+1+nqso]):
+        f, wave, meta_qso = qso.make_templates(nmodel=1,
+                                               redshift=np.array([zqso[i]]), mag=np.array([mag_g[i]]), seed=seed[i])
+
```
moustakas: So we still have to loop here?  Can't we generate N>>1 pure QSO spectra and then just do the normalization and IGM attenuation (below) in the loop?  I think I asked this before and realized that this wasn't viable but I can't recall why not.

py/desisim/lya_spectra.py
```diff
@@ -46,40 +48,45 @@ def get_spectra(infile, first=0, nqso=None, seed=None):
...
-
-        ## will interpolate the transmission at the spectral wavelengths, 
-        ## if outside the forest, the transmission is 1
-        itr=interpolate.interp1d(la,tr,bounds_error=False,fill_value=1)
-        flux[i,:]*=itr(wave)
-        padflux, padwave = normfilt.pad_spectrum(flux[i, :], wave, method='edge')
+        if len(tr):
+            ## will interpolate the transmission at the spectral wavelengths, 
+            ## if outside the forest, the transmission is 1
+            itr = interpolate.interp1d(la,tr,bounds_error=False,fill_value=1)
```
moustakas: Minor quibble: PEP 8? (here and where `flux` gets populated, below).

py/desisim/lya_spectra.py
forero: We can generate N>>1 QSO spectra, but it's superslow. Having the loop here is the key to the 10x speedup.

py/desisim/lya_spectra.py
forero: got back `seed=None`

py/desisim/lya_spectra.py
forero: It is initialized now to `f4`

py/desisim/templates.py
forero: solved

py/desisim/lya_spectra.py
forero: solved

forero:
@sbailey CI test is failing randomly with error

`The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install graphviz texlive-latex-extra dvipng" failed and exited with 100`

what would be a possible workaround?

weaverba137:
This failure appears to be genuinely random.  The particular Ubuntu server it was downloading from appears to have been offline at that moment.  The only way to see if it is still failing is to submit another test.

forero:
The tests pass now. Merging.

PR #212, opened 2017-02-26
update the API of lya_spectra to work with mock-->spectra connection

moustakas:
This updates the API of `lya_spectra` (building off of #210) in order to accommodate the needs of https://github.com/desihub/desitarget/issues/136 when generating Lyman-alpha QSO spectra.

[The Travis tests were failing on my laptop because it couldn't load `speclite.filters` (which I don't understand), but let's see what happens here.]


moustakas:
This is holding up some other work so I'm merging now.  Tests pass.  Please open a new issue if this creates any other problems.

PR #213, opened 2017-02-27
support DA and DB white dwarf subtypes

moustakas:
This PR addresses #200 (and a need of https://github.com/desihub/desitarget/issues/136) to support DB- and DA-type white dwarfs via a `subtype` optional input string to the `WD` class. As requested by @sbailey in #200, the subtype is also written to the output metadata table, which allows us to potentially expand this column to include other finer-grain properties of the various templates (see also the discussion in https://github.com/desihub/desitarget/issues/149).  I also added a couple basic unit tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,10 +44,10 @@ env:
         - ASTROPY_VERSION=1.2.1
         # - SPHINX_VERSION=1.3
         # - DESIUTIL_VERSION=1.8.0
-        - DESIUTIL_VERSION=master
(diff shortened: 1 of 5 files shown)


moustakas:
Tests are failing because the tests are using the `v2.2` version of the basis templates.  I think I'm ready to upgrade the templates in `trunk` to `v2.3`.  I'll do that and report back.

moustakas:
Tests pass on my laptop but Travis seems to be having problems...

PR #214, opened 2017-03-04
Fix test failures

sbailey:
A variety of test failures had crept into desisim mostly due to non-backwards compatible changes in other packages.  These went unnoticed in desisim since the Travis tests were pinned to older versions.  These changes get desisim back into compatibility with the latest versions of desispec, desitarget, and specter.

desisim itself will be ready for a fresh tag after confirming that these tests pass on Travis and merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,9 +50,9 @@ env:
         - DESISIM_TESTDATA_VERSION=0.3.3
         - SPECLITE_VERSION=0.5
         - SPECSIM_VERSION=v0.6
-        - SPECTER_VERSION=0.6.0
-        - DESISPEC_VERSION=0.12.0
-        - DESITARGET_VERSION=0.8.1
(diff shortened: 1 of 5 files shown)


sbailey:
Looks like there was a bug in desispec installation of data/ccd/ccd_calibration.yaml .  I think I've fixed that in desispec master; testing here.  If this works I'll update the desispec tag, make this branch of desisim depend on that tag, and try again...

i.e. comments welcome but don't merge this just yet

moustakas:
I tested and verified that all tests now pass, so I'm merging.

Note for the future:  I was getting many failures on my laptop when testing `quickgen` because of this line:
```
    assert output['observed_flux'].unit == 1e17 * fluxunits
```
but it turns out that I had version 1.3.0 of `astropy` rather than `1.2.1` installed in my conda environment.

PR #220, opened 2017-03-11
Include DLAs in LyaForest

profxj:
Added routines to inject DLAs into the Lya Forest of simulated QSO templates.

Briefly, the DLAs are:

- Drawn from NHI = 19.5 - 22.5
- Inserted according to the observed incidence (i.e. dN/dz)
- There is a small boost in the incidence to account for the fact that the NHI range goes down to 19.5
- Only Lya is modelled.  

New test, no new docs aside from advances to the QSO_templates Notebook.


ngbusca:
Sorry for the delay, I'll take a close look by tomorrow

py/desisim/lya_spectra.py
```diff
@@ -6,9 +6,17 @@
 """
 
 from __future__ import division, print_function
+import pdb
```
sbailey: leftover import?

py/desisim/lya_spectra.py
```diff
@@ -6,9 +6,17 @@
...
+
+import numpy as np
+from scipy.special import wofz
+
+from pkg_resources import resource_filename
+
+c_cgs = 29979245800.0  # cm/s
```
sbailey: preferable to use `scipy.constants.speed_of_light` (check units) or `astropy.constants.c.to('cm/s').value`

py/desisim/lya_spectra.py
```diff
@@ -104,10 +114,267 @@ def get_spectra(lyafile, nqso=None, wave=None, templateid=None, normfilter='sdss
         normmaggies = np.array(normfilt.get_ab_maggies(padflux, padwave, 
                                                        mask_invalid=True)[normfilter])
         flux1 *= 10**(-0.4 * mag_g[ii]) / normmaggies
+
+        # Inject a DLA?
+        if add_dlas:
+            if np.min(wave/1215.67 - 1) < zqso[ii]: # Any forest?
+                if debug:
```
sbailey: preferable to use `log=desiutil.log.get_logger(); log.debug(message)` rather than `if debug: print...`.  Or remove altogether if you are done debugging.

py/desisim/lya_spectra.py
```diff
@@ -104,10 +114,267 @@ def get_spectra(lyafile, nqso=None, wave=None, templateid=None, normfilter='sdss
...
+    with a given redshift
+    Args:
+        wave:
+        zem:
+        rstate:
+        seed:
+        fNHI:
+        **kwargs:
+
+    Returns:
```
sbailey: Please finish this documentation including units, how `**kwargs` is used, and whether `insert_dlas` updates any of the inputs in-place vs. returning a new spectrum with DLAs while leaving the original untouched.  Thanks for carefully documenting the docstrings of the other functions added here.

py/desisim/lya_spectra.py
```diff
@@ -27,6 +35,9 @@ def get_spectra(lyafile, nqso=None, wave=None, templateid=None, normfilter='sdss
           SEED.
         qso (desisim.templates.QSO, optional): object with which to generate
           individual spectra/templates.
+        add_dlas (bool): Inject damped Lya systems into the Lya forest
+          These are done according to the current best estimates for the incidence dN/dz
```
sbailey: Could you add a citation for "current best estimates" and guidance somewhere of what should be changed if we wanted to update this?

For consideration: From the name of `fN_spline_z24.fits.gz` it sounds like you may be storing spline parameters, which may be a pain to understand/update if we want to update in the future (splines seem straightforward but every implementation seems to parameterize them slightly differently...).  Should we just store the raw histogram numbers to which a spline is fit on the fly?

moustakas:
Thanks, @profxj.  It would also be helpful if you could time the code with and without this new feature.  Speed has been an issue for the Data Challenge (most recently addressed by @forero in #210).

ngbusca:
The code looks good and it's clearly written.

I've verified that for a given set of z_abs and nhi I get the same absorption profile with an independent calculation.

I have two minor requests though:
* could you add the DLA info to the output metatable?
* when calling get_spectra multiple times with the same random seed, I get the same continua but different DLAs. Could you propagate the random state to the function insert_dlas (l. 123 in lya_spectra.py)?

After you add the meta-data info, it would be good to produce a histo of zabs and nhi to make sure they correspond to the desired input from you paper.

Thanks!!

moustakas:
How much additional metadata is going to be needed?  Remember that the template-level metadata (seed, redshift, normalizing magnitude, etc.) is "stacked" across the various spectral classes (e.g., when processing the mocks), so we should be careful that it doesn't grow without bound.

Perhaps a dedicated, DLA-specific metadata table / optional output?

ngbusca:
@moustakas It's tricky for the DLAs because you don't know how many are inserted in each spectrum, most of the times there are no DLAs but you can have up to a few per line of sight. We need to keep track of two numbers per DLA: zabs and nhi. I'm not sure what's the best data structure to keep track of this...

sbailey:
Thanks for catching the random number non-reproducibility, @ngbusca .  @profxj please add a unit test to verify that the same seed results in the same DLAs and different seeds result in different DLAs.  This is something that we frequently break with refactors, so we should have a unit test to protect against that.

DLA metadata: since there could be 0 or 1 or multiple DLAs per spectrum, how about a DLA-specific metadata table (as @moustakas suggested) while making it one-row-per-DLA instead of one-row-per-target.  One of the columns would be which target that DLA was added to.  That DLA metadata table would only be returned if DLAs were added.  The target mapping would need to be general enough that we could run different patches of sky in parallel and stack the DLAmeta tables without losing track of which target each DLA goes with.

andreufont:
Hi everyone. I found this thread at random while googling related stuff, and I was very happy to see all this progress :-)

Just a comment: from what I gathered above, we add the Lyman alpha forest from an input file, then re-normalize the spectrum, and then add the DLA. Shouldn't we add the DLA before re-normalizing the spectrum?

profxj:
Well we 'add' a DLA multiplicatively.  So I don't think it matters (although I will double-check).

py/desisim/lya_spectra.py
profxj: removed

py/desisim/lya_spectra.py
profxj: done

py/desisim/lya_spectra.py
profxj: Citation added (it was below already)

For the spline, I fear it is much more complicated than that although
I do agree with your concern.

py/desisim/lya_spectra.py
profxj: removed

py/desisim/lya_spectra.py
profxj: done

profxj:
Added info to meta tables in keys:

- n_DLA -- Number of DLAs in the sightline
- z_DLA -- Redshifts of the DLAs (padded to 100 entries)
- NHI_DLA -- Log NHI of the DLAs

@moustakas , I suggest that @ngbusca do the profiling.  Would take me 10x longer to do so.

I can't pass the rstate (unless we return it from the template code) but I am passing in the seed now (had meant to).

Modified the seed in the test to insure at least 1 DLA every time

ngbusca:
I've done test of drawing 1000 spectra with and without DLAs (average over 10 draws):
* without DLAs: 2m58s 
* with DLAs: 3m15s
So about 8% slower when adding DLAs. I think this is acceptable. I'll let @moustakas comment on the increased size of the metadata.

I agree with @andreufont that the DLA absorption should be added to the absorption field before normalization (I didn't understand @profxj argument of why it wouldn't make a difference). 

One final comment: passing the random seed to the DLA code fixes the issue or reproducibility. But it's the same seed as what's passed to the continuum generator, meaning that the random state is reset to what it was before the generation of the continua. It's unlikely that this will pose problems but, @sbailey is this acceptable? Shouldn't there be a global random generator accessible from anywhere in the code? (I realize the complications of multithreading that, but there must be a known solution).

ngbusca:
@profxj The absorption lyman-alpha field is added between lines l.115 to l. 120 (the block that follows the if len(tr): statement. I would add the dla absorption here, so that when the spectrum is normalized (l. 122-124) the normalization correctly accounts for the extra dla absorption.

profxj:
ok, got it.  I think we are good aside from @moustakas thinking about meta bloat.

moustakas:
I've finally had a chance to look at this.  This enhancement -- while obviously great -- is definitely going to cause some issues with how the spectra are connected to the mocks.  A tiny minority of the mock targets are going to be Lyman-alpha QSOs, and an even smaller subset of those are going to have DLAs, but the way the code is currently structured we would have to carry around a lot of extra zeros in the metadata (aka 'truth.fits') output tables.

What if we made an enhanced set of Lya/QSO mocks in which we had *pre-computed* the numbers, redshifts, and other properties of the DLAs such that the properties of each DLA belongs to the mock.  In the spectrum-generating code we then generate the DLA on-the-fly (using the desired properties as inputs) and an end-user can use the unique mockid mapping to look back at those properties as needed.  

Thoughts?

profxj:
Seems a lot easier to carry around a few extra zeros.  All of which will
compress trivially (if you are compressing the meta data).

sbailey:
I agree with @moustakas concerns about the extra metadata burden.  Could we switch to a DLA-specific metadata table suggested earlier in this thread?  Although extra zeros could be externally gzip'ed to efficiently pack the zeros, that doesn't help with internal fits compression (not fully supported for tables yet) nor for memory burden.

profxj:
grumble. ok.

andreufont:
About metadata for DLAs in mocks. Even though in the current implementation DLAs are distributed randomly, in the future we will certainly need to be able to place them in a way that is correlated with the density in the simulated boxes.

This means that we will need an option to add DLAs to read Z_DLA and N_HI from a file equivalent to the one we use to read the Lyman alpha forest.

It would be good if during the current discussion on meta data we had this in mind.

profxj:
Failing Travis for some doc string error (I think).

Otherwise ready for @moustakas + @sbailey to give the thumbs up.

Note that the passback from get_spectra() now depends on 
whether add_dlas is True/False.  We could alternatively always
have it be 4 items with the last one None if add_dlas=False.

moustakas:
Thanks for this PR, @profxj.  It looks like I created some conflicts in your notebook which I'll clean up, but we'll also want this refactor to take into account the significant changes merged in #234.  In other words, it would be a plus to have `desisim.templates` generate the DLAs on-the-fly as well as `desisim.lya_spectra`.  Will get to this as soon as I can --

profxj:
Pinging @moustakas

andreufont:
Hi everyone, 

This PR request has been stalled for a very long while. Looking through the comments above, it looks like there were two issues back then:
 - conflicts in the documentation notebooks, and Travis checks failing.
 - it was not clear the best way to store the DLA-specific meta data. 

Related to the last point, I mentioned that in the long term we'll want to be able to add DLAs to the mocks in positions read from an external file. This is equivalent to the way we add random forest fluctuations: we can do this on the fly (using LyaMockP1D) or we can do this by reading the fluctuations from a file. We can leave this for a future PR, though, and get something running for now so that we can test the DLA finders on DESI mocks.

@moustakas - have there been major changes in desisim since we had this discussion in March?

profxj:
Bullet 2 has been dealt with (storing the meta data).

I believe the rest is on @moustakas but he should
scream if he believes otherwise.  ;)

moustakas:
Yes, this is on me -- apologies for the delay.  @profxj could you resolve the conflicts fix the failing tests first and then I'll get this into the mock spectra code.

profxj:
Merged and passing tests except for some doc
string error I can't find.  Maybe @sbailey knows
how to squash these..

All yours now @moustakas

sbailey:
@weaverba137 can you help us figure out the correct way to format the return values portion of the docstring in a way that is both sphinx/rst legal and human readable (e.g. when using pydoc, ipython help, or just browsing code)?  It doesn't seem to like the multi-line entries, even though the same basic formatting is used for the "Args:" section.
```
    Args:
        lyafile (str): name of the Lyman-alpha spectrum file to read.
        nqso (int, optional): number of spectra to generate (starting from the
          first spectrum; if more flexibility is needed use TEMPLATEID).
        wave (numpy.ndarray, optional): desired output wavelength vector.
        templateid (int numpy.ndarray, optional): indices of the spectra
          (0-indexed) to read from LYAFILE (default is to read everything).  If
          provided together with NQSO, TEMPLATEID wins.
        [...]
    Returns:
        flux (numpy.ndarray): Array [nmodel, npix] of observed-frame spectra
          (erg/s/cm2/A).
        wave (numpy.ndarray): Observed-frame [npix] wavelength array (Angstrom).
        meta (astropy.Table): Table of meta-data [nmodel] for each output spectrum
          with columns defined in desisim.io.empty_metatable *plus* RA, DEC.
        dla_meta (astropy.Table): Table of meta-data [ndla] for the DLAs injected
          into the spectra.  Only returned if add_dlas=True
```
It is parsing the Args: portion fine, but complaining about indentation within the "Returns" section.

weaverba137:
The return value of that function is a tuple containing four objects correct?  So the return value is a tuple.  Describe the four entries in the tuple:
```
Returns:
    foo (tuple): A tuple containing flux (numpy.ndarray)...; wavelength (numpy.ndarray) ...;
        metadata (astropy.Table) ...; ...
```
Alternatively see if numpydoc formatting allows multiple entries in the Returns section.

sbailey:
Tests pass including docstring formatting.  Can this be merged now without breaking anything, or does it still require updates from @moustakas or others?

`get_spectra` only returns the `dla_meta` entry if `add_dlas=True` (default False), so I don't think it would break any existing code, even if other code isn't yet updated to use DLAs.  Is that correct?

moustakas:
Thank you for cleaning up the remaining issues, @sbailey.

Issue #219, opened 2017-03-10
templates.QSO().make_templates is slow


sbailey:
Spinning a conversation with @profxj into an issue:

`templates.QSO().make_templates(n=100)` is surprisingly slow, especially given that `desisim.lya_spectra.get_spectra()` is fast and calls `templates.QSO().make_templates()`.

Figure out why and fix it.

moustakas:
One difference is that `lya_spectra` passes the redshift to `QSO.make_templates` as an input.

moustakas:
Fixed in #225.

PR #225, opened 2017-03-24
fix N^2 scaling of QSO.make_templates

sbailey:
This PR contributes the 4-character fix to solve the `QSO.make_templates` scaling problem reported in #219.  The core problem was that it was looping over redshifts but then calling `desi_qso_templ.desi_qso_templates` with the full redshift array every time, thus it was generating N^2 QSOs while only picking N of them.


moustakas:
Thank you for finding this bug.

Issue #209, opened 2017-02-13
need to speed up desisim.templates


moustakas:
Related to #207, I'm finding *significant* overhead in ```desisim.templates``` due to ```pixelsplines.gauss_blur_matrix```, which convolves each spectrum to a desired (intrinsic) velocity dispersion.  It looks like most of the overhead is in ```scipy.coo_matrix.tocsr```.  

We need to resolve this in order to make the ```desitarget.mock.targets_truth``` mocks-->spectra refactor tractable (see #136).

Is anyone available to help with this?

moustakas:
One possibility would be to velocity-blur each spectrum to a restricted set of velocity dispersions and then pre-compute and store the sparse matrices needed before building each spectrum.  In the spirit of adding as much variety as possible to our simulated spectra, doing this isn't my first choice. Moreover, the code would need a decent amount of refactoring to deal with this.

Or are there other (fast) alternatives to the necessary (flux-conserving) smoothing operation?

moustakas:
To be quantitative:  skipping ```desisim.templates.GALAXY.vdispblur``` entirely leads to a factor of ~22 speed-up in the "generate a spectrum" step (from ~1.55 sec/spectrum to 0.07 sec/spectrum). Yikes!

moustakas:
This issue was also discussed in #58.

moustakas:
Addressed in #229.

moustakas:
I'm reopening this because although #229 led to a reasonable speed-up of the velocity dispersion broadening portion of `desisim.templates`, it's still too slow for production-scale simulations (e.g., for the various data challenges).  Moreover, the code as written is a memory hog.

@dkirkby mentioned to me off-list that he had some ideas for dealing with these issues. Input is welcome!

moustakas:
Some very nice ideas here --
http://adsabs.harvard.edu/abs/2017MNRAS.466..798C

moustakas:
With the current version of `desitarget/bin/select_mock_targets` we're generating hundreds of thousands of spectra without too much trouble, so although the velocity dispersion broadening *is* a slow part of the spectral generation, it's not the only one.

PR #229, opened 2017-04-04
significant speed-up of desisim.templates for galaxies

moustakas:
This PR addresses #209 by using a smaller subset of unique, ungridded velocity dispersions (10% of the number of spectra being generated or 40, whichever is smaller) when generating galaxy (ELG, LRG, BGS) templates.  

I wasn't sure where to put it, but I added a small timing script to `desisim/bin/time-vdisp-cache` which compares the performance of the code with and without the incorporated changes (see below).  Building 100 templates is now a factor of ~12(!) faster with the new code vs the old code.  

The other change was to allow the [OIII]/H-beta line-ratio to meander outside of the prior range specified by the `oiiihbeta` optional input, which saved another ~20% of time.  This is a minor issue that's in the weeds for the current state of the code and can be revisited at a future time.

(Tests are still running but submitted this now.)

### DIFF
--- a/bin/time-vdisp-cache
+++ b/bin/time-vdisp-cache
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+
+import numpy as np
+import argparse
+
+from desisim.templates import ELG
+from desisim.io import empty_metatable, read_basis_templates
+
+def make_elg(nobj=50, seed=123, fracvdisp=(0.1, 40), no_cache=False):
+    
+    rand = np.random.RandomState(seed)
+
+    if no_cache:
+        vdisp = 10**rand.normal(1.9, 0.15, nobj)
+    else:
+        nvdisp = int(np.min((np.fix(nobj * fracvdisp[0]), fracvdisp[1], 1)))
+        vdisp = 10**rand.normal(1.9, 0.15, nvdisp)
+
+    meta = read_basis_templates('ELG', onlymeta=True)
+
+    input_meta = empty_metatable(nobj, objtype='ELG')
+    input_meta['TEMPLATEID'] = rand.choice(meta['TEMPLATEID'], nobj)
+    input_meta['REDSHIFT'] = rand.uniform(0.8, 1.2, nobj)
+    input_meta['MAG'] = rand.uniform(21, 23, nobj)
+    input_meta['SEED'] = rand.randint(2**32, size=nobj)
+    input_meta['VDISP'] = rand.choice(vdisp, nobj)
+
+    ELG().make_templates(input_meta=input_meta, nocolorcuts=True)
+    #ELG().make_templates(nmodel=nobj, seed=seed, nocolorcuts=True)
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-n', '--nobj', type=int, help='Number of spectra to simulate.')
+    parser.add_argument('--no-cache', action='store_true', help='Do not cache the velocity dispersion (slow!)')
+
+    args = parser.parse_args()
+    make_elg(args.nobj, no_cache=args.no_cache)
--- a/py/desisim/templates.py
+++ b/py/desisim/templates.py
@@ -371,8 +371,11 @@ class GALAXY(object):
     def lineratios(self, nobj, oiiihbrange=(-0.5, 0.2), oiidoublet_meansig=(0.73, 0.05),
                    agnlike=False, rand=None):
         """Get the correct number and distribution of the forbidden and [OII] 3726/3729
-           doublet emission-line ratios.  Note that the agnlike option is not
-           yet supported.
+        doublet emission-line ratios.  Note that the agnlike option is not yet
+        supported.
+
+        Supporting oiiihbrange needs a different (fast) approach.  Suppressing
+        the code below for now until it's needed.
 
         """
         if agnlike:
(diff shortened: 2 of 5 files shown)


moustakas:
Here are the timing tests --

Old code with 100 templates:
```
(desi) ioannis@atlas:~/repos/desihub/desisim% time python time_vdisp_cache.py -n 100 --no-cache
WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.
WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.
INFO:io.py:613:read_basis_templates: Reading /Users/ioannis/research/projects/desi/spectro/templates/basis_templates/v2.3/elg_templates_v2.0.fits metadata.
INFO:io.py:625:read_basis_templates: Reading /Users/ioannis/research/projects/desi/spectro/templates/basis_templates/v2.3/elg_templates_v2.0.fits

real    1m46.413s
user    1m26.831s
sys     0m16.777s
```

New code with 100 templates:
```
(desi) ioannis@atlas:~/repos/desihub/desisim% time python bin/time-vdisp-cache -n 100
WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.
WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.
INFO:io.py:613:read_basis_templates: Reading /Users/ioannis/research/projects/desi/spectro/templates/basis_templates/v2.3/elg_templates_v2.0.fits metadata.
INFO:io.py:625:read_basis_templates: Reading /Users/ioannis/research/projects/desi/spectro/templates/basis_templates/v2.3/elg_templates_v2.0.fits

real    0m10.164s
user    0m9.260s
sys     0m0.537s
```

bin/time-vdisp-cache
```diff
@@ -0,0 +1,38 @@
...
+from desisim.io import empty_metatable, read_basis_templates
+
+def make_elg(nobj=50, seed=123, fracvdisp=(0.1, 40), no_cache=False):
+    
+    rand = np.random.RandomState(seed)
+
+    if no_cache:
+        vdisp = 10**rand.normal(1.9, 0.15, nobj)
+    else:
+        nvdisp = int(np.min((np.fix(nobj * fracvdisp[0]), fracvdisp[1], 1)))
```
sbailey: I think you meant `np.max` rather than `np.min` here.  Your test is always testing the case of nvdisp=1 regardless of nobj.

py/desisim/templates.py
```diff
@@ -692,8 +730,8 @@ def make_galaxy_templates(self, nmodel=100, zrange=(0.6, 1.6), magrange=(21.0, 2
                     if nocontinuum or novdisp:
                         blurflux = restflux[this, :] * magnorm[this]
                     else:
-                        blurflux = (self.vdispblur((restflux[this, :] - thisemflux), vdisp[ii]) + \
-                                    thisemflux) * magnorm[this]
+                        blurflux = ((blurmatrix[blurindx[ii][0]][1] * (restflux[this, :] - thisemflux)) + \
```
sbailey: For long term maintainability, please add come comments to `_blurmatrix` explaining what the structure of `blurindx` and `blurmatrix` are.  The lists of tuples of (integer,matrix) ... make my head spin.  Could this structure be simplified?  This appears to be the only place that `blurmatrix` and `blurindx` are used, which would argue for simplifying the return structures of `_blurmatrix` so that the magic extra levels of `[0]` and `[1]` indexing wouldn't be needed here.  i.e. could you turn this into `blurmatrix[blurindx[ii]]`?

For consideration: if `blurmatrix` was a `dict` keyed by `vdisp[ii]` instead of a `list` indexed by `bluridx[ii]`, this would all boil down to `blurmatrix[vdisp[ii]]` which seems more understandable to me, and would simplify `_blurmatrix` too since you wouldn't need to do the `blurindx` bookkeeping.  Note: any concerns about two different values of vdisp being epsilon different would also apply to `list(set(vdisp))` in `_blurmaxtrix` so you're not avoiding that issue by using `blurindx`.

py/desisim/templates.py
```diff
@@ -365,20 +389,27 @@ def lineratios(self, nobj, oiiihbrange=(-0.5, 0.2), oiidoublet_meansig=(0.73, 0.
...
-            siihbeta[need] = samp[:,3]
-            oiiihbeta[oiiihbeta<oiiihbrange[0]] = -99
-            oiiihbeta[oiiihbeta>oiiihbrange[1]] = -99
+        samp = EMSpectrum().forbidmog.sample(nobj, random_state=rand)
+        oiiihbeta = samp[:, 0]
+        oiihbeta = samp[:, 1]
+        niihbeta = samp[:, 2]
+        siihbeta = samp[:, 3]
+
+        if False:
```
moustakas: Done.

py/desisim/templates.py
```diff
@@ -535,10 +566,13 @@ def make_galaxy_templates(self, nmodel=100, zrange=(0.6, 1.6), magrange=(21.0, 2
                 mag = rand.uniform(magrange[0], magrange[1], nmodel).astype('f4')
 
             if vdisp is None:
+                # Limit the number of unique velocity dispersion values.
+                nvdisp = np.max((np.min((np.round(nmodel * self.fracvdisp[0]), self.fracvdisp[1])), 1)).astype(int)
```
moustakas: Yes, it's correct.  I cleaned up the spacing to be more clear.

py/desisim/templates.py
moustakas: This is much much smarter; done.

bin/time-vdisp-cache
moustakas: Ugh, right.  Oh well, factor of 12 was a little too crazy to believe anyway.

bin/time-vdisp-cache
```diff
@@ -0,0 +1,38 @@
...
+    input_meta['TEMPLATEID'] = rand.choice(meta['TEMPLATEID'], nobj)
+    input_meta['REDSHIFT'] = rand.uniform(0.8, 1.2, nobj)
+    input_meta['MAG'] = rand.uniform(21, 23, nobj)
+    input_meta['SEED'] = rand.randint(2**32, size=nobj)
+    input_meta['VDISP'] = rand.choice(vdisp, nobj)
+
+    ELG().make_templates(input_meta=input_meta, nocolorcuts=True)
+    #ELG().make_templates(nmodel=nobj, seed=seed, nocolorcuts=True)
+
+if __name__ == '__main__':
```
moustakas: Right.  I moved it to `etc` as requested and we can remove it downstream if it starts to clutter.

moustakas:
This is ready for re-review; all the requested changes have been made and tests pass.

PR #231, opened 2017-04-07
do not cache blurmatrix when novdisp=True

moustakas:
This logic should have been included in #229.  When `novdisp=True` I was incorrectly still caching the memory-hungry velocity dispersion convolution matrices.

### DIFF
--- a/py/desisim/templates.py
+++ b/py/desisim/templates.py
@@ -358,7 +358,7 @@ class GALAXY(object):
         blurmatrix = dict()
         for uvv in uvdisp:
             sigma = 1.0 + (self.basewave * uvv / LIGHT)
-            blurmatrix[uvv] = pxs.gauss_blur_matrix(self.pixbound, sigma)
+            blurmatrix[uvv] = pxs.gauss_blur_matrix(self.pixbound, sigma).astype('f4')
 
         return blurmatrix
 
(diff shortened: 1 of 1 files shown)


moustakas:
Because unit tests pass and the proposed changes are so minor -- and I believe only affect me -- I'd like to self-merge this soon.  Please holler if this creates unintended problems!

py/desisim/templates.py
```diff
@@ -358,7 +358,7 @@ def _blurmatrix(self, vdisp, log=None):
         blurmatrix = dict()
         for uvv in uvdisp:
             sigma = 1.0 + (self.basewave * uvv / LIGHT)
-            blurmatrix[uvv] = pxs.gauss_blur_matrix(self.pixbound, sigma)
+            blurmatrix[uvv] = pxs.gauss_blur_matrix(self.pixbound, sigma).astype('f4')
```
sbailey: I'm not convinced that keeping the blur matrix as f4 helps, since it is likely upcast back to f8 every time it is used, perhaps creating even more memory churn since an f8 version would be created multiple times and the garbage collector may or may not keep up.  Are you sure this helps?

moustakas: I'll create a new issue on the use of double- vs single-precision in the templates code but for now (i.e., for the mocks-to-spectra effort) this part of the code isn't being used.

PR #233, opened 2017-04-12
Faster quickcat

apcooper:
Calling np.in1d in the loop over tileids in tileid quickcat.get_redshift_efficiency seemed to be a bottleneck (taking ~hours) for whole-survey simulations. This PR replaces that loop by argsorting the targetids and doing a one-off match of the concatenated list of targets in each tile.

For first epoch, 73260852 targets, 3134 tiles to gather in zcat:

Before:
```
Tue Apr 11 15:58:14 2017 QC Adding ZERR and ZWARN
Tue Apr 11 19:08:04 2017 QC Adding NUMOBS column
```

After:
```
Tue Apr 11 12:28:32 2017 QC Adding ZERR and ZWARN
Tue Apr 11 12:29:26 2017 QC Adding NUMOBS column
```

So speedup in this case is ~3hrs down to < 1min.


sbailey:
Wow, super-duper.  This always felt like it was taking way longer than it should but I hadn't found what was causing the bottleneck.  Thanks for contributing this.  I just updated changes.rst too and will merge now.

Issue #232, opened 2017-04-09
pragmatically avoiding negative values in QSO templates


sbailey:
The current QSO template generation code excludes PCA combinations that have negative flux ([desi_qso_templ line 346](https://github.com/desihub/desisim/blob/master/py/desisim/qso_template/desi_qso_templ.py#L346)).  This requires generating many more templates than are needed in the hopes that enough will pass this cut.  The current padding factor `ipad` of 20x isn't always sufficient, causing the code to crash.

This issue documents some implementation ideas for how to be robust against negative flux while not wasting memory and compute time on much larger ipad values.

#### The current negative flux check is for restframe wavelengths 900 < lambda < 5000 Angstrom.  Change that to checking only within the observer frame wavelength visibility window. ####

Ideally the templates would be valid over the full wavelength range for every realization, but my experience with fitting PCA to spectra is that often they have a good positive fit within the observer frame visibility window while allowing non-physical negativity outside of that.  We have tuned the coefficient means and sigmas to the fits to data within each redshift bin, and those coefficients likely represent combinations that wouldn't pass our current non-negativity cuts.

#### Work (more) iteratively ####

Rather than tune `ipad` to be some larger number for the worst case and have to generate many more templates than are needed for the typical case, work iteratively, e.g.

1. Generate N_perz spectra
2. Keep the good ones
3. If we don't have enough good ones yet, repeat until we do or some upper iteration limit is reached

We do still want to have some upper iteration limit to avoid wasting a huge amount of time in case input parameters mean that it really can't generate enough good QSOs, but the loop above would allow us to finish faster and in less memory in the typical case, while increasing the ipad upper iteration limit.

Note: the current code partially does this, only calculating the lookback distance and filling in the final output arrays for templates that pass non-negativity cuts, but it still generates a full `N_perz*ipad` templates as a starting point.  That could become prohibitive if we need larger `ipad`.

#### Use Gaussian or actual distribution of coefficients rather than uniform ####

Currently the coefficients are drawn uniformly within +/- 2sigma of their mean for each coefficient within each redshift bin ([L332-L333](https://github.com/desihub/desisim/blob/master/py/desisim/qso_template/desi_qso_templ.py#L332-L333)), meaning that they are spending more time exploring corners of phase space rather than the more peaky actual distributions.  Pulling from the actual distributions or from Gaussians would likely generate more combinations that pass cuts.

[...]

PR #234, opened 2017-04-13
logging tweaks to desisim.templates and one bug fix

moustakas:
The changes in this PR were made while working on https://github.com/desihub/desitarget/issues/169.  The biggest change is that `desisim.templates.QSO` now uses the `desitarget.cuts.isQSO_colors(optical=True)` option (whereas previously it unsatisfactorily didn't apply *any* color-cuts).  

Because of this, the Travis tests temporarily have to depend on desitarget/master.  @sbailey if we could cut a tag of `desitarget` post-https://github.com/desihub/desitarget/pull/172 then I can update the `.travis.yml` file here before merging.  

Hmm, tests are failing because of an issue in `desisim.test.test_targets`.  Debugging now but opening this PR anyway.


moustakas:
`desisim.test.test_targets.test_random` is failing because of the issue mentioned [here](https://github.com/desihub/desisim/issues/232#issuecomment-293856255).  The fastest way to "fix" this would be to not apply color-cuts when generating QSO templates in `desisim.targets.get_targets`, but I don't know what / how that impacts code that @sbailey is working on.  I don't think it's a big deal because the distribution of targets we care about are coming from mocks, so we could resolve this at a later date.  I'll submit the requisite change anyway so we can discuss.

sbailey:
desitarget 0.11.0 has been tagged

moustakas:
This is officially a poorly named branch as I decided to do a major refactor of the QSO template-generating code on this branch to address #232 and to also (by default) include Lyman-alpha forest absorption in each QSO template using code provided by @andreufont--see [here](https://github.com/igmhub/LyaMockP1D). 

In essence, in addition to incorporating ~all the suggestions made by @sbailey in #232, I consolidated the code in `desisim.qso_template.desi_qso_templ` entirely into `desisim.templates.QSO`.  In detail, for those willing to review the changes please take a look at:
  https://github.com/desihub/desisim/pull/234/files#diff-9131b35ddb55549286cd73decb05cdcb

The templates also by default include Lyman-alpha forest, which can be suppressed by setting `lyaforest=False`.

andreufont:
Hi @moustakas - I went through the lya_mock_p1d related code, and I believe it looks good. A couple of comments:  
 - I noticed that you call get_lya_skewers() for each quasar, instead of calling get_lya_skewers(nmodel). Is this because you want to make sure that each quasars uses its own templateseed[ii] instead of just passing a global seed for all the quasars? I don't think this matters much, since this code should be quite fast, but if at some point this mattered we could modify get_lya_skewers to accept a list of random seeds, or similar.
 - The notebook says "Generate 100" and then only generates 10. But what I loved from the notebook is the last plot, with a a negative continuum and an inverted forest! I'm aware that this will be fixed in another pull request, but it blew my mind :-)

sbailey:
I'm still getting problems with the code hitting the maximum number of iterations without getting a good spectrum and returning all zeros.  e.g.
```python
import desisim.templates
qsomaker = desisim.templates.QSO()
flux, wave, meta = qsomaker.make_templates(100, seed=1)
```
gives
```
WARNING:templates.py:1901:make_templates: Maximum number of iterations reached on QSO 20, z=3.66191.
WARNING:templates.py:1901:make_templates: Maximum number of iterations reached on QSO 23, z=3.32587.
WARNING:templates.py:1901:make_templates: Maximum number of iterations reached on QSO 59, z=3.99063.
WARNING:templates.py:1901:make_templates: Maximum number of iterations reached on QSO 68, z=2.99034.
WARNING:templates.py:1901:make_templates: Maximum number of iterations reached on QSO 77, z=3.44711.
WARNING:templates.py:1901:make_templates: Maximum number of iterations reached on QSO 84, z=3.30221.
WARNING:templates.py:1901:make_templates: Maximum number of iterations reached on QSO 86, z=3.32487.
WARNING:templates.py:1901:make_templates: Maximum number of iterations reached on QSO 97, z=2.99546.
WARNING:templates.py:1908:make_templates: 8 spectra could not be computed given the input priors!
```
This is worse than master, which doesn't generate any bad spectra for this case.  Am I missing something (e.g. updated templates or a desitarget version or something)?  Please try this.

moustakas:
These "failures" are largely due to the fact that the Lyman-alpha forest scatters the QSO out of the optical color-box defined in `desitarget.cuts.isQSO_colors`, which is more of an issue of the Targeting WG.  Perhaps @Cyeche can look into these sources?  I'm generally finding that the simulated QSO spectra that include Lyman-alpha forest in the optical have much bluer r-z and g-r colors than lower-redshift QSOs.

You can turn this off with either
```
flux, wave, meta = qsomaker.make_templates(100, seed=1, lyaforest=False)
```
or
```
flux, wave, meta = qsomaker.make_templates(100, seed=1, nocolorcuts=True
```

That said, there still seems to be cases where the normalization factor (i.e., the synthesized magnitude) ends up being negative, which I'm looking into now.

All this heartache---by the way---is one of the many reasons I despise PCA!

moustakas:
OK I just pushed a bug fix (which cause the forest in emission, @andreufont--glad you got a chuckle!) and also updated the notebook with a color-color plot to show exactly what's going on:
  https://github.com/desihub/desisim/blob/tweaks/doc/nb/QSO_templates.ipynb

Cyeche:
John,

I'm confused by your statement " I'm generally finding that the simulated QSO spectra that include Lyman-alpha forest in the optical have much bluer r-z and g-r colors than lower-redshift QSOs." Do you mean that the Ly-alpha QSO are bluer than the low-z QSO? In principle, as there is absorption in Ly-a region, the low-z QSO should be bluer, see attached plot. If I have correctly understood your sentence (bluer and redder is always confusing for me, like right and left ;-)) I would suggest that you debug this point with the persons in charge of Ly-a simulations.

Christophe.

sbailey:
I suspect that we are double applying Lyman-alpha mean absorption.  The PCA templates/coefficients are binned in redshift, and thus probably already include the mean Lyman-alpha absorption at that redshift.  We are then additionally applying an absorption field skewer, further decrementing the forest beyond the mean, making the QSOs too blue and failing selection cuts.

moustakas:
@Cyeche (or anyone) Do you have a g-r vs r-z color-color plot with the (spectroscopic) z>2.1 QSOs identified?  (Your message referred to an attached figure, but it didn't come through on Github). This would be very helpful (and apologies if this is somewhere obvious like in the FDR).

sbailey:
I'll go ahead and merge this now so that we don't hold up the other fixes, while spinning the QSO cuts and LyAF into a separate ticket.

Here is a histogram of generated QSO redshifts (blue) and those that didn't find a spectrum that passes the color cuts after 20 tries (orange).  We're basically knocking out a redshift window around 3.1<z<3.6 and likely selecting non-typical QSO PCA realizations around that.

badqso

moustakas:
Thanks, @sbailey.  This may also be related to my incorrect implementation of the attenuation blueward of the Lyman-limit (on roughly line 1856 of `desisim.templates`).  If any of the QSO experts in the room have time to take a look I'd be grateful.  

Also, if @profxj could comment on the extent to which the output QSO spectra have double the amount of Lya forest absorption that would also be very helpful.

Issue #235, opened 2017-04-18
z>3 QSO templates often fail color cuts: doubly applying LyA absorption?


sbailey:
First identified in #234: when we add Lyman-alpha absorption to QSO templates, they often fail the targeting color cuts for QSOs with 3.1 < z < 3.7 .  I suspect that we are double applying LyA mean absorption and thus are generating overly blue QSOs.  @xavier @andreufont @ngbusca could you take a look and confirm / fix?

ngbusca:
@sbailey Could you put a snippet of code that illustrates the problem?

andreufont:
I don't know if the PCA templates include mean absorption, if so we should
fix this.

But if this was the case, adding a double lya absorption would make the
quasars redder, isn't it? The forest eats blue light.

Andreu

sbailey:
@andreufont yes, you are right.  Redder.

@ngbusca example of them failing:
```python
import desisim.templates
qsomaker = desisim.templates.QSO()
flux, wave, meta = qsomaker.make_templates(100, seed=1)
```

John's notebook has example spectra and g-r vs. r-z colors with and without LyAF at the bottom:

https://github.com/desihub/desisim/blob/master/doc/nb/QSO_templates.ipynb

I'm just guessing about the double applying LyAF mean absorption.  First task is to check whether the PCA templates have mean absorption or not (I'm guess they do).

andreufont:
About that notebook: cell #8 it shows spectra for the 10 quasars, with and
without the forest. Blue-ward of Lyman alpha these are the same for all
quasars but the second, that has larger continuum after adding forest. I
could see this happening if we are rescaling the spectra after adding the
forest, to match the input magnitude, but then why does this only happens
for the second quasar?

Andreu

ngbusca:
The plot below shows the median templates near z=2.5 and z=3.5. It looks like there is indeed absorption in the templates.

The way the templates are constructed is by sampling the pca coefficients of BOSS spectra near the input redshift. My guess is that those coeffs were obtained including the forest in the pca fits. If yes, then the fix would be to redetermine them masking the forest, which implies updating the pca input files.

![image](https://cloud.githubusercontent.com/assets/4727509/25151965/3516eb6c-2490-11e7-893e-a8f8fb37e269.png)

ngbusca:
@andreufont by default, r-mags (5500 < lambda < 6500, approx) are used to normalize the spectra. The second spectrum is the only one that has some forest in the r-band.

[...]

PR #236, opened 2017-04-19
remove lya absorption below lyman-lymit 

ngbusca:
This PR intends to address issue #235 by removing the lyman-alpha absorption below the lyman-limit (which is already taken into account in the template).

### DIFF
--- a/py/desisim/templates.py
+++ b/py/desisim/templates.py
@@ -1815,9 +1815,11 @@ class QSO():
 
             # Interpolate the Lya forest spectrum.
             if lyaforest:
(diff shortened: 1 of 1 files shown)


ngbusca:
@andreufont can you take a look?

andreufont:
@ngbusca, I'm not sure what you mean. In principle there is nothing wrong with having Lyman alpha forest down to the blue end of the spectrograph, right? Regardless of rest-frame wavelength.

ngbusca:
I've now found a different problem  at the red end when adding the forest described in #235 (bottom plots). I removed my previous (wrong) "fix"  and added a red-end fix. @moustakas I seem to be able to generate z~3.5 qsos with forest without a problem using this branch, can you double check?

Issue #97, opened 2016-03-07
review objtype / flavor usage


sbailey:
Our usage of flavor and objtypes is getting a little out of hand in desispec.  ELG, LRG, QSO seemed straightforward enough, but we are starting to get things like MWS vs. MWS_STAR, STD vs. STD_FSTAR vs. FSTD, etc.  This issue is a placeholder to think about the big picture of how this should work and do some cleanup.

moustakas:
With data now in-hand, I think our choices are basically now baked in. Closing with no action.

PR #243, opened 2017-06-12
fix tests and code related to flavor=science

sbailey:
This PR fixes some pieces of desisim that were broken by desihub/desispec#386 which became more picky about allowable exposure flavors when writing Frame objects.

The whole usage of "flavor" vs. "objtype" vs. "program" still needs to be cleaned up (see #97), but this at least gets us back to a working state on desisim + desispec master.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -2,19 +2,17 @@
 desisim change log
 ==================
 
-0.19.1 (unreleased)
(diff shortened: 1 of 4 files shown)


sbailey:
merging now to get this in place before the nightly updates at NERSC.

PR #244, opened 2017-06-16
simulating desi spectra notebook tutorial

moustakas:
This PR adds a notebook tutorial for generating simulated spectra using `quickgen`.  However, the tutorial falls apart at the end when calling `desispec.scripts.makebricks` because `$DESI_SPECTRO_SIM`, `$DESI_SPECTRO_DATA`, and `$DESI_SPECTRO_REDUX` are not playing nice and I'm utterly confused about what to update without breaking things.  

Here's the direct link to the notebook -- @sbailey and @geordie666 please take a look if you get a chance.
  https://github.com/desihub/desisim/blob/using-quickgen/doc/nb/simulating-desi-spectra.ipynb

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,9 @@ desisim change log
 0.19.1 (unreleased)
 -------------------
 
-* No changes yet.
+* Adds tutorial on simulating spectra (`PR #244`_).
+
+.. _`PR #244`: https://github.com/desihub/desisim/pull/244
 
 0.19.0 (2017-06-15)
 -------------------
--- a/doc/nb/simulating-desi-spectra.ipynb
+++ b/doc/nb/simulating-desi-spectra.ipynb
@@ -6,10 +6,13 @@
    "source": [
     "# Simulating DESI Spectra\n",
     "\n",
-    "The goal of this notebook is to demonstrate how to generate some simple DESI spectra using the *quickgen* utility.  For simplicity we will only generate 1D spectra and skip the more computationally intensive (yet still instructive!) step of extracting 1D spectra from simulated 2D spectra (i.e., so-called \"pixel-level simulations\").\n",
+    "The goal of this notebook is to demonstrate how to generate some simple DESI spectra using the `quickgen` utility.  For simplicity we will only generate 1D spectra and skip the more computationally intensive (yet still instructive!) step of extracting 1D spectra from simulated 2D spectra (i.e., so-called \"pixel-level simulations\").\n",
     "\n",
-    "For additional (albeit somewhat outdated) information and documentation about *quickgen* see  \n",
+    "For additional (albeit somewhat outdated) information and documentation about `quickgen` see  \n",
     "https://desi.lbl.gov/DocDB/cgi-bin/private/ShowDocument?docid=1429\n",
+    "\n",
+    "The heart of `quickgen` is the `SpecSim` package, which you can read out here:  \n",
+    "http://specsim.readthedocs.io/en/stable\n",
     "\n",
     "If you identify any errors or have requests for additional functionality please create a new issue on  \n",
     "https://github.com/desihub/desisim/issues\n",
@@ -54,12 +57,30 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Populating the interactive namespace from numpy and matplotlib\n"
+      "WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.\n",
+      "WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.\n",
+      "WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.\n"
+     ]
+    },
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/usr/local/anaconda3/envs/desi/lib/python3.5/site-packages/matplotlib/__init__.py:1401: UserWarning:  This call to matplotlib.use() has no effect\n",
+      "because the backend has already been chosen;\n",
+      "matplotlib.use() must be called *before* pylab, matplotlib.pyplot,\n",
+      "or matplotlib.backends is imported for the first time.\n",
+      "\n",
+      "  warnings.warn(_use_error_msg)\n"
      ]
     }
    ],
    "source": [
-    "%pylab inline"
+    "import desispec.io\n",
+    "import desisim.io\n",
+    "from desisim.obs import new_exposure\n",
+    "from desisim.scripts import quickgen\n",
+    "from desispec.scripts import group_spectra"
    ]
   },
   {
@@ -73,28 +94,21 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.\n",
-      "WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.\n",
-      "WARNING:log.py:19:get_logger: desispec.log is deprecated, please use desiutil.log.\n"
+      "Populating the interactive namespace from numpy and matplotlib\n"
      ]
     }
    ],
    "source": [
-    "import desispec.io\n",
-    "import desisim.io\n",
-    "from desisim.obs import new_exposure\n",
-    "from desisim.scripts import quickgen\n",
-    "from desispec.scripts import makebricks\n",
-    "\n",
-    "from desiutil.log import get_logger\n",
-    "log = get_logger()"
+    "%pylab inline"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Make sure we have all the right environment variables set (assuming the bash shell).  If any of these environment variables are missing please set them in your *.bashrc* file (and then restart this notebook) or create them for just this notebook session using the *%set_env* magic command, as we demonstrate below."
+    "#### Check and set your environment\n",
+    "\n",
+    "Next, make sure we have all the right environment variables set (assuming the bash shell).  If any of these environment variables are missing please set them in your `.bashrc` file (and then restart this notebook) or create them for just this notebook session using the `%set_env` magic command, as we demonstrate below."
    ]
   },
   {
@@ -125,13 +139,13 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "DESIMODEL environment set to /Users/sbailey/desi/git/desimodel/svn/trunk\n",
-      "DESI_ROOT environment set to /data/desi\n",
-      "DESI_SPECTRO_SIM environment set to /data/desi/spectro/sim\n",
-      "DESI_SPECTRO_DATA environment set to /data/desi/spectro/data\n",
-      "DESI_SPECTRO_REDUX environment set to /data/desi/spectro/redux\n",
-      "SPECPROD environment set to debug\n",
-      "PIXPROD environment set to debug\n"
+      "DESIMODEL environment set to /Users/ioannis/repos/desihub/desimodel\n",
+      "DESI_ROOT environment set to /Users/ioannis/research/projects/desi\n",
+      "DESI_SPECTRO_SIM environment set to /Users/ioannis/research/projects/desi/spectro/sim\n",
+      "DESI_SPECTRO_DATA environment set to /Users/ioannis/research/projects/desi/spectro/data\n",
+      "DESI_SPECTRO_REDUX environment set to /Users/ioannis/research/projects/desi/spectro/redux\n",
+      "SPECPROD environment set to dailytest\n",
+      "PIXPROD environment set to dailytest\n"
      ]
     }
    ],
@@ -143,7 +157,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Let's reassign the *\\$SPECPROD* environment to something other than \"dailytest\" so that we don't conflict with the outputs of the standard DESI integration test.  In addition, we need to make raw data input *\\$DESI_SPECTO_DATA* match *\\$DESI_SPECTRO_SIM/\\$PIXPROD* where the simulated data will be written."
+    "Let's reassign the `$SPECPROD` environment to something other than `dailytest` so that we don't conflict with the outputs of the standard DESI integration test.  In addition, we need to make raw data input `$DESI_SPECTO_DATA` match `$DESI_SPECTRO_SIM/$PIXPROD` where the simulated data will be written."
    ]
   },
   {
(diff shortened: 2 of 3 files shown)


sbailey:
There is some confusion (not unique to you) about the various environment variables and directories.  Attempting to clarify:

The intension is that environment variables define default locations, while every script/function should have a way to override that (not every script/function has that override yet).

Raw data are in `$DESI_SPECTRO_DATA/{yearmmdd}/*.fits`.

Simulated raw data (including simspec and fibermap files) are written to `$DESI_SPECTRO_SIM/$PIXPROD`.  The idea is that you set `$DESI_SPECTRO_SIM` once for your system, and then vary `$PIXPROD` for different isolated simulation runs.

As an unfortunate consequence, when running downstream steps, you have to set `$DESI_SPECTRO_DATA=$DESI_SPECTRO_SIM/$PIXPROD` since the downstream steps aren't supposed to know that their inputs are simulated and thus look at `$DESI_SPECTRO_DATA`.

i.e. `$DESI_SPECTRO_DATA` and `$DESI_SPECTRO_SIM` refer to simulated *inputs*, not pipeline/quickgen *outputs*.  That seems to be the key confusion in this notebook.

Outputs are written to `$DESI_SPECTRO_REDUX/$SPECPROD`, with a similar philosophy: you set `$DESI_SPECTRO_REDUX` once and then vary `$SPECPROD` for different runs.  `$SPECPROD` might match `$PIXPROD`, or it might be different due to different runs on the same input with different code versions.

`desisim.io.simdir(night=night)` returns the directory where to write the simulated input for that night, i.e. `$DESI_SPECTRO_SIM/$PIXPROD/{night}/`.

In your example call to make_bricks, you don't actually need to specify `--specprod` since the code will find it from `$DESI_SPECTRO_REDUX/$SPECPROD`, but if you do specify it for extra clarity, it should match those, i.e.
```
args = makebricks.parse(['--night', night, '--verbose'])
makebricks.main(args)
```
or
```
specprod_dir = os.path.join( os.getenv('DESI_SPECTRO_REDUX'), os.getenv('SPECPROD') )
args = makebricks.parse(['--night', night, '--specprod', specprod_dir, '--verbose'])
makebricks.main(args)
```
or
```
specprod_dir = desispec.io.specprod_root()
args = makebricks.parse(['--night', night, '--specprod', specprod_dir, '--verbose'])
makebricks.main(args)
```
I'll save a more fundamental issue for a separate comment to be written in a moment

sbailey:
My previous comment should get you past the brick making, but let me know if that didn't make sense or didn't work.

More fundamentally, we are deprecating the brick grouping and brick format in favor of the healpix grouping and spectra format that merges b,r,z channels into a single file.  These are generated with `desi_group_spectra`, but that code currently only works in the context of a successful pipeline run, and requires the pipeline yaml cookie crumbs to correctly reflect the state of processing instead of using the actual cframe*.fits files on disk from quickgen.  Even if you create a production after the fact to get the graph of tasks and files, `desi_group_spectra` doesn't work.  e.g.
```
(desi) example $ desi_pipe
Working with production /data/desi/spectro/redux/example :
  Updating plans ...
[...]
  Generating scripts ...

(desi) example $ desi_group_spectra
mpi4py not found, using only one process
INFO:group_spectra.py:61:main: Starting at Fri Jun 16 13:51:07 2017
INFO:group_spectra.py:62:main:   using raw dir /data/desi/spectro/redux/example
INFO:group_spectra.py:63:main:   using spectro production dir /data/desi/spectro/redux/example
INFO:group_spectra.py:119:main: Distributing 0 spectral groups among 1 processes
INFO:group_spectra.py:222:main: Finishing at Fri Jun 16 13:51:07 2017
(desi) example $ ls spectra/
(desi) example $ desi_group_spectra --nights 20170615
mpi4py not found, using only one process
INFO:group_spectra.py:61:main: Starting at Fri Jun 16 13:51:22 2017
INFO:group_spectra.py:62:main:   using raw dir /data/desi/spectro/redux/example
INFO:group_spectra.py:63:main:   using spectro production dir /data/desi/spectro/redux/example
INFO:group_spectra.py:119:main: Distributing 0 spectral groups among 1 processes
INFO:group_spectra.py:222:main: Finishing at Fri Jun 16 13:51:22 2017

(desi) example $ ls spectra

(desi) example $ 
```

@tskisner can you take a look at this and recommend a path forward?  Options that occur to me are:
  * I've misunderstood something and there is some way to get `desi_group_spectra` to work with quickgen output as is
  * update `quickgen` to mimic the pipeline yaml files (ugly; yaml files are being refactored away anyway...)
  * update `desi_group_spectra` to work from the real files on disk rather than requiring / trusting the state of the yaml files
  * continue to use brick files for this example (and thus train the collaboration in a format we don't intend to use in the future...)
  * something else?

tskisner:
Hi folks,

The desispec pipeline is designed to go from input data, through a series of steps to the final redshifts.  You can definitely override steps of the pipeline with your own code and customize it in other ways.  For example, you could pretty easily make pipeline steps use quickgen instead of the full PSF estimation and extraction.  Note this is NOT what has been done so far- instead people have been dumping quickgen outputs into a pre-created pipeline production.

For this discussion, it "feels" like you want run a small production, using quickgen for some steps, but you want to do it without actually running the pipeline tools.

Some steps of the pipeline can be run manually on input files to create manually specified output files.  The spectral grouping step requires knowing which cframe files are touched by each healpix pixel.  That in turn requires going through all of the cframe files, parsing the fibermap, and computing the healpix pixels for each spectrum from the target RA/DEC.  All this information is computed once in the pipeline production and then loaded by the regrouping step.

So the path(s) forward are:

1.  Add a new function that parses all the cframe files (basically that replicates the piece of the pipeline production setup).  Then modify the spectral grouping commandline entry points to either use pipeline metadata or compute it on the fly.

2.  Simply add pipeline workers that call quickgen, and then try to convince people to run a pipeline.

Since I don't expect to make much progress convincing people to use desi_pipe and then run the generated bash scripts, I'll start work on item (1) instead.  However, for the record, implementing (2) would be:

A) write a trivial worker for quickgen that performs the "extraction" step.
B) Run desi_pipe with "--fakeboot" and "--fakepsf".
C) Modify the options.yaml file to select the quickgen extraction worker and any options you want to pass to it.
D) Run the generated pipeline bash scripts on your laptop.

tskisner:
Trying to figure out exactly how you want this command line entry point to work...  Am I allowed to use $DESI_SPECTRO_REDUX and $SPECPROD?  In other words, should I assume that I can use desispec.io.find_file?

Or do you want to pass in a full list of cframe files on the command line and the output spectra directory?

Looking at the other command line scripts in desispec.scripts, it seems they do not use the environment variables, so that is the model I will stick to.  This means that the user will also need to specify the "chunking factor" to use for the top level grouping of files (unless that is now standardized by Stephen in the desispec code somewhere).  It looks like the pipeline code in master is still using my original healpix grouping...

sbailey:
I just submitted PR for updating desispec.io.findfile to match dc17a2 output directory structure; will merge after tests pass (or someone else can do it for me if they noticed that the tests have passed and I haven't gotten to it yet).  Since that only affects where the *output* spectra go, it shouldn't be a blocking factor for updating `desi_group_spectra` to not require a production to have been run.

Two options for inputs here:
1. if you are starting from the fibermap files in a raw data directory (default `$DESI_SPECTRO_SIM/$PIXPROD`), then ok to use `desispec.io.findfile` to locate the matching `cframe` files.
2. alternatively, you could start from a base redux exposures dir (default `$DESI_SPECTRO_REDUX/$SPECPROD/exposures`) and use `os.walk` or `desispec.io.iterfiles` to completely agnostically find all cframe files

In either case, use `desispec.io.findfile` to derive where to put the output spectra.

(2) is more generic (perhaps overly so), has the advantage of working purely from redux output without requiring a separate raw data dir, but has the disadvantage of opening 30 frame files looking at pieces of fibermaps instead of 1 input fibermap file (some clever caching could reduce that to 10).  (1) is more restrictive but probably fine.  For the purposes of today, do whatever is fastest to implement that will work with John's tutorial in this PR.

tskisner:
It begs the question- if we are starting from a production directory, then....  why not just run the pipeline script?  I thought running desi_pipe was in fact the way that you got a $DESI_SPECTRO_REDUX/$SPECPROD/exposures directory in the first place?  Or do you create that manually?  Or does quickgen create it?

For each step of the pipeline, some "workers" each run one or more little "tasks".  The command line tools in desispec.scripts are basically wrappers around running one "task".  The scripts in here do not use any external environment variables, just commandline options.

In the pipeline, the regrouping step is a special case since it has only one (very big) task.  If you really want this script to be "aware" of the production that it is running in, then OK.  Just wanted to point out that it is different from the other entry points.

tskisner:
I will go with your (2) above, since that is the easiest to implement.

sbailey:
In practice, we often run `quickgen` first and then after-the-fact run `desi_pipe` to populate the plan.  Either way appears to work.

But it appears that `desi_group_spectra` appears to need more than just the plan graph yaml file to have been created by `desi_pipe`.  If you think it should work from just the plan file but not the "yes this step is done" files, then maybe we are fixing the wrong thing.  i.e. try running John's notebook, substituting the `makebricks` step at the end with
```
args = makebricks.parse(['--night', night, '--verbose'])
makebricks.main(args)
```
and augment that with `desi_pipe` anywhere along the chain as you see fit, and try to get a production directory that works with `desi_group_spectra`.  I haven't found a working recipe for that.

Or proceed with your current update.

moustakas:
Thanks, I've been following the conversation.  So everything in the notebook looked OK except in and around cell [17] instead of calling `makebricks` I'll call....and I just saw that @sbailey just commented.  So I'm standing by and will defer to whatever @tskisner can most easily implement.

tskisner:
The current desi_group_spectra loads the plans for all nights listed in the command line option.  Those "planning" yaml files (one per night) include the dependencies between every spectra file and every frame, computed when desi_pipe is run.  The "state" files in the current pipeline are used for nothing except monitoring the status with desi_pipe_status- they are not relevant here.

I'll get this finished this evening, but it won't be for a couple hours due to domestic chores.

sbailey:
OK, then this goes back to my first possibility of "I've misunderstood something and there is some way to get desi_group_spectra to work with quickgen output as is".  Thanks for the help getting this recipe+code into a working state.

tskisner:
If you run desi_pipe with "--fakeboot --fakepsf --fakepix", then the resulting pipeline scripts will skip bootcalib, psf estimation, and extraction.  So this assumes that "something else" will create the frame-* files.  Then you could run the calibration pipeline step to produce cframe files.  And then run the regrouping.

If quickgen is simulating the already-calibrated cframe files, then you don't need to run the calibration step, just run the $SPECPROD/run/scripts/spectra.sh script to regroup the cframe files.  So to summarize:

(If you have no environment variables set- sourcing that setup file will define them)
```
$> desi_pipe --fakeboot --fakepsf --fakepix --data <raw data dir> \
--redux <reduction directory> --prod <prod name>
$> source <reduction director>/<prod name>/setup.sh
```
(If you already manually set $DESI_SPECTRO_DATA, $DESI_SPECTRO_REDUX, $SPECPROD)
```
$> desi_pipe --fakeboot --fakepsf --fakepix
```
Then use quickgen to make frame files (or calibrated cframe files?)

If you just have frame files:
```
$> cd $DESI_SPECTRO_REDUX/$SPECPROD/run/scripts
$> ./run_shell.sh
```
Which will run the pipeline (only the calibration step in this case).  If you already have cframe files from quickgen, then just do:
```
$> cd $DESI_SPECTRO_REDUX/$SPECPROD/run/scripts
$> ./spectra.sh
```
To run just the regrouping step.  The bash flavor of these scripts just run the pipeline serially (or using local mpirun, if you use some other options to desi_pipe).

sbailey:
My previous attempts at running `desi_pipe` in this notebook were silently sort-of failing due to `$DESI_SPECTRO_DATA` being set incorrectly, resulting in `desi_pipe` not finding anything but also not complaining about it.

I updated the notebook to:
* clarify (?) / standardize the environment variable usage at the beginning so that `desi_pipe` finds
  the fibermaps that we were writing
* encourage reading HDUs by name not number
* add `desi_pipe --fakeboot --fakepsf --fakepix`

However, this highlighted a new problem in the quickgen / desi_pipe interaction: apparently `desi_pipe` can't plan a production for which there weren't any arcs or flats, generating the traceback below.

Note that we can still proceed with quickgen; we're just blocked at the `desi_group_spectra` step.

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-17-6319d2d14e00> in <module>()
      1 from desispec.scripts import pipe_prod
      2 args = pipe_prod.parse(['--fakeboot', '--fakepsf', '--fakepix'])
----> 3 pipe_prod.main(args)

/Users/sbailey/desi/git/desispec/py/desispec/scripts/pipe_prod.py in main(args)
    326     print("  Updating plans ...")
    327     expnightcount, allpix = pipe.create_prod(nightstr=args.nights, 
--> 328         extra=extra, specs=specs, fakepix=args.fakepix, hpxnside=args.nside)
    329     totcount = {}
    330     totcount["flat"] = 0

/Users/sbailey/desi/git/desispec/py/desispec/pipeline/plan.py in create_prod(nightstr, extra, specs, fakepix, hpxnside)
    182             os.makedirs(nlog)
    183 
--> 184         grph, expcount, npix = graph_night(nt, specs, fakepix, hpxnside)
    185 
    186         for pix in npix:

/Users/sbailey/desi/git/desispec/py/desispec/pipeline/plan.py in graph_night(rawnight, specs, fakepix, hpxnside)
    505         cam = "{}{}".format(band, spec)
    506         flatid = None
--> 507         for fid in sorted(flatexpid[cam]):
    508             if (flatid is None):
    509                 flatid = fid

KeyError: 'r0'
```

@tskisner please try out this notebook and advise.  Thanks.

https://github.com/desihub/desisim/blob/using-quickgen/doc/nb/simulating-desi-spectra.ipynb

tskisner:
So it sounds like basically you just want to support pipelines that have no calibration exposures.  Ok, I can hack in that feature...   In hindsight, we might have considered creating alternate pipeline workers that called the quick simulations, but not on the 2 day timescale.

tskisner:
I just opened desispec pull request #405, which fixes a couple things and enables the spectra grouping to work without pipeline infrastructure.  So now there are 3(!) ways to call this code.  In your python notebook, the desi_pipe call you have should work now.  Then run quickgen as you say.  Then do one of these:

(this works on a login node / laptop)
```
$> cd run/scripts
$> ./spectra.sh
```
OR (at NERSC)
```
$> cd run/scripts
$> sbatch spectra.slurm
```
OR
```
$> desi_group_spectra
```

sbailey:
Notebook updated with new desi_regroup_spectra.  I'm ready to hand off to John for final check / updates, or merging now.  Would be nice to add final examples of reading and plotting frames and spectra, though that will be also covered in a different tutorial and I didn't want to hold a blocking lock on this for too long.

tskisner:
Note that the "hpixnside" option to desi_group_spectra (when run outside the pipeline script), should match the value passed in the "--nside" option to desi_pipe.  The default for both is 64, so maybe just leave it out in this example?

moustakas:
This is ready to merge once tests complete.  The notebook looks pretty good (I think!).  Thanks for all the help @tskisner, @sbailey.

PR #247, opened 2017-07-07
fix QSO template extrapolation issue

moustakas:
This PR needs https://github.com/desihub/desispec/pull/415 before it can be merged, but it fixes https://github.com/desihub/desispec/issues/414.  In essence, the QSO templates will no longer have a funny step function at long wavelengths when extrapolating (see cell [18] of the updated python notebook).


moustakas:
I thought I would be able to merge this now that https://github.com/desihub/desispec/pull/415 has been merged, but Travis tests are failing on an `import healpy as hp` error when importing `desitarget.cuts`.  This is odd since the `.travis.yml` file depends on the master branch of `desitarget`.

weaverba137:
I'm not sure what the desispec merge has to do with desitarget.  & in any case, you are not requiring the presence of healpy in the .travis.yml file for desisim.

moustakas:
`desisim` depends on `healpy` only through `desitarget`.  I didn't think I needed to declare `desitarget`'s dependencies in the `desisim` Travis file.   I'll add that here if that's the case.

And FYI this PR was dependent on https://github.com/desihub/desispec/pull/415, which is why I was waiting for that merge.

sbailey:
Tests ran fine on my laptop for this branch, so I added a few debugging printouts to this branch to see what is going on with Travis.

sbailey:
We needed specsim v0.9 to get a units propagation bug fix from desihub/specsim#63.

I also updated travis to use desimodel 0.7.0 but I don't think that was necessary.  The failure on commit 2e2ab6b was the infrequent LRG flux=0 problem (used to be more frequent, not appears very occasionally).

moustakas:
Thank you, @sbailey.

PR #238, opened 2017-04-24
refactored newexp

sbailey:
Disclaimer: this is a rather unsatisfying refactor since so much more could be done and it just doesn't "feel" right yet, but in the spirit of the 2% sprint I'm going to stop shuffling deck chairs and provide something that works for now.

Please review this but let's hold off on merging until we've confirmed that it properly handles the full 2% mock spectra input (I debugged it on a much earlier run).  Tests pass on my laptop but I see I still have some Travis cleanup to do...

This refactors `newexp-desi` (now just `newexp`) to accept inputs from a mock target catalog, mock truth spectra, fiber assignment, and observing conditions from surveysim instead of making up a random exposure on the fly.  It also switches to using `specsim` for the sky/lunar model and fiberloss calculations, also laying the groundwork for optionally using galsim for detailed per-object fiberloss calculations.

A random dark/bright/etc. exposure can still be generated from a python API (old `desisim.obs.new_exposure` code) but not via a command-line script anymore.

`newarc` and `newflat` are split out as separate scripts since they take different parameters than a new science exposure (e.g. no observing conditions or fiber assignment inputs...)

This also provides a `simulate_spectra` convenience function that wraps specsim to provide a translation from obsconditions and fibermap tables into how to construct and run a specsim Simulator.  At its simplest this is just `sim = simulate_spectra(wave, flux)`, but it optionally can take surveysim inputs for specifying observing conditions or a target/fiberassign/fibermap table to specify target locations and targeting bits to derive fiber input losses.

simspec format was simplified (?) to use binary tables with columns e.g. WAVELENGTH, FLUX, SKYFLUX instead of one image HDU per variable.  If this mucks up BGS or other work that may have been writing simspec formats on their own, I could consider expanding `desisim.io.read_simspec` to support both formats.

Examples:

Generate new arc and flat exposures to be written to `$DESI_SPECTRO_SIM/$PIXPROD/{night}/*-{expid}.fits`:
```
newarc --night 20190830 --expid 1
newflat --night 20190830 --expid 3
```
Write output to a different location:
```
newarc --night 20190830 --expid 1 \
    --simspec temp/simspec-arc.fits --fibermap temp/fibermap-arc.fits
```
Generate a new science exposure using fiberassignment, surveysim, and mock spectra inputs:
```
newexp \
    --fiberassign fiberassign/tile_37299.fits \
    --obslist two_percent_DESI/twopct.ecsv \
    --mockdir mocktargets/ \
[...]

Issue #221, opened 2017-03-15
need to be able to simulate (noisy) DA and DB white dwarf spectra


moustakas:
The ability to generate subtype DA and DB white dwarf templates was added in #200 but `newexp-desi`, `pixsim-desi`, and `quickgen` do not handle this "flavor" of target.  We should fix this with the new refactor. 

This issue is *kind of* related to #203 and #97.

moustakas:
Simulating both DA and DB WDs is now a standard part of the spectra generated with `desitarget/bin/select_mock_targets` (as long as the yaml configuration file points to an appropriate WD mock, see, e.g., [here](https://github.com/desihub/two_percent_DESI/blob/master/dc17b/select-mock-targets.yaml)).  

In addition, smaller-scale simulations of WDs can be done trivially by calling `desisim/bin/quickspectra` (by way of a simple example, see [this notebook](https://github.com/desihub/desisim/blob/master/doc/nb/bgs-redshift-efficiency.ipynb), which simulates 1000 stand-alone BGS spectra).

Members of the MWS Working Group should reopen this issue if they have more specific needs not covered here.

PR #250, opened 2017-07-28
refactored newexp (again)

sbailey:
This is the replacement PR for #238 that got bogged down in rebasing and merge conflicts due to keep the branch and PR open too long.  Trying again.

Tests pass on cori; moving on to Travis now.  Will also re-confirm with integration tests prior to final merge.


moustakas:
@sbailey Does this branch address #221?

sbailey:
This PR has gotten out of hand already.  Will merge prior to implementing smaller isolated changes like #221 (Adding support for WD subtypes DA and DB).  Running final (?) travis tests now, while checking simulating spectra tutorial in parallel.

sbailey:
Close notes:
 
This will require minor changes to the simulating spectra tutorial since the return values of `desisim.obs.new_exposure` changed, and the simspec HDU 'METADATA' changed name to 'TRUTH'.  I'll submit that after this is merged.

I'll also submit an update to the desispec integration test to call `newexp-random` instead of `newexp-desi`.

quickgen runs and produces working cframe files, but the quicklook integration test fails on the sky subtraction step.  It does work on master (after a minor desispec.pipeline.runcmd -> desispec.util.runcmd fix) so I broke something here, but I haven't been able to chase that down yet and don't want that to block this PR.

`astropy.io.fits.BinTableHDU(meta)` generates a broken HDU with our metadata tables from make_templates(). `astropy.io.fits.convenience.table_to_hdu(Table(meta))` is the workaround.  Unit tests didn't catch that problem, which is one symptom of this new code not having good test coverage.

A few features that I discussed with @moustakas and @akremin that didn't make it in here yet:
* WD DA/DB
* passing bgs_kwargs dict into `obs.new_exposure()` to propagate down through `get_targets` and eventually to `BGS.make_templates()`

Issue #251, opened 2017-08-01
gmag bug in quickcat?


moustakas:
@forero Is [this code block](https://github.com/desihub/desisim/blob/master/py/desisim/quickcat.py#L207-L218) in `quickcat` a bug?

`DECAM_FLUX` returns fluxes in nanomaggies but line 218 treats the g-band flux as a magnitude.

[If this is indeed a bug I can fix it on the branch I'm working on now, which is updating the `targets` and `truth` tables to the latest imaging data model.]

forero:
I don't know if that's a bug. It looks like it is one. 
Consult first with @cballand . He wrote those lines.

moustakas:
Apologies, I thought you wrote `quickcat`, @forero!

forero:
Don't worry! `quickcat` has been a real collaborative effort!

cballand:
I had a quick look and I'm  not sure it's a bug (but I should check this in more detail): as far as true_gmag and qso_gmag_threshold are in the same units (nanomaggies), the flux ratio should be expressed as it is on line 218 (i.e., it should only depend on the difference true_gmag - qso_gmag_threshold).

cballand:
In fact you're right, this is a bug, true_gmag is a flux in nanomaggies and should be converted into a magnitude before being used in line 218.

moustakas:
OK, great.  I've fixed this in my branch and will merge after Stephen's PR.

cballand:
Ok, thanks @moustakas

moustakas:
Fixed in #252.

PR #252, opened 2017-08-01
Update datamodel for templates metadata table

moustakas:
This PR builds on https://github.com/desihub/desitarget/pull/189 by updating the `templates` code to return a metadata table which conforms to the latest imaging data model (DR4 and beyond).  Specifically, `DECAM_FLUX[:, 6]` and `WISE_FLUX[:, 2]` have now been split into individual `FLUX_G`, `FLUX_R`, `FLUX_Z`, `FLUX_W1`, `FLUX_W2` columns.  See the [DR4 data release webpage](http://legacysurvey.org/dr4/catalogs) for more details.

Any code that makes assumptions about the format of the metadata table will need to be updated -- I only fixed code within `desisim`.  

I think I've also addressed the bug identified in #251 but @forero should confirm.

I'm now going to update the associated code called by `select_mock_targets`.


moustakas:
@sbailey The failing test is in `test_newexp` with a "possibly wrong units" error.  Before I dig into this and fix it, should I change / update the spectra returned by `templates` to have the `1e-17` factor in them (and also possibly explicitly use the correct astropy units)?  I think this is the only place in either `desisim` or `desispec` where this standard is not (currently) being enforced.

sbailey:
We chatted on phone, but coming up for air on github too: I originally supported "change the datamodel variable names first, reintegrate end-to-end, and then change units later" but from the git traffic it looks like units/tests are getting in the way, so if you need to bit the bullet and convert everything (including the tests) to 1e-17 now, we can update the downstream pieces too.

moustakas:
@sbailey I decided to move forward with converting the output flux vector from `templates` to `1e-17 cgs` and I *think* I've fixed most of the unit test failures, at least within desisim.  Are there other places (e.g., the integration tests) that I should check?

sbailey:
If you can get the desisim unit tests working, I'll take care of the integration tests and any knock-on effects to other repos.

moustakas:
Unless you feel strongly otherwise, in the metadata table I'm going to keep `OIIFLUX` and `HBETAFLUX` in plain erg/s/cm2/A, i.e., without the factor of 1e-17.

I'll note here that these line-fluxes do not carry astropy units with them; we should revisit the use of units throughout the simulation workflow at a future date.

moustakas:
This is ready for review / to be merged, @sbailey.

py/desisim/templates.py
```diff
@@ -704,11 +708,11 @@ def make_galaxy_templates(self, nmodel=100, zrange=(0.6, 1.6), magrange=(21.0, 2
...
-                        gflux=synthnano[:, 1],
-                        rflux=synthnano[:, 2],
-                        zflux=synthnano[:, 4],
-                        w1flux=synthnano[:, 6],
-                        w2flux=synthnano[:, 7])
+                        gflux=synthnano[:, 0],
+                        rflux=synthnano[:, 1],
+                        zflux=synthnano[:, 2],
+                        w1flux=synthnano[:, 3],
+                        w2flux=synthnano[:, 4])
```
sbailey: `synthnano` as a 2D array predates this PR, but it is somewhat fragile to have these magic indices tied to the order of the filters specified in the constructor `self.decamwise = filters.load_filters...`.  Especially since part of the motivation for unpacking the `decam_flux` 2D arrays is to get away from these magic indices, could you do the same here with code something like:
```python
synthnano = dict()
for key in enumerate(maggies.columns):
    synthnano[key] = 1E9 * maggies[key] * magnorm # nanomaggies
...
colormask = self.colorcuts_function(
    gflux=synthnano['decam2014-g'],
    rflux=synthnano['decam2014-r'],
    zflux=synthnano['decam2014-z'],
    w1flux=synthnano['wise2010-W1'],
    w2flux=synthnano['wise2010-W2])
```
Same comment applies to several other sections with magic indices into `synthnano` arrays. This caught my eye because the the mapping of indices to filters changed; using a dictionary keyed by filter name would avoid that problem.

py/desisim/templates.py
```diff
@@ -753,9 +760,9 @@ def make_galaxy_templates(self, nmodel=100, zrange=(0.6, 1.6), magrange=(21.0, 2
                         format(np.sum(success == 0)))
 
         if restframe:
-            return outflux, self.basewave, meta
+            return outflux * 1e17 * u.erg / (u.cm**2 * u.s * u.Angstrom), self.basewave * u.Angstrom, meta
         else:
-            return outflux, self.wave, meta
+            return outflux * 1e17 * u.erg / (u.cm**2 * u.s * u.Angstrom), self.wave * u.Angstrom, meta
```
sbailey: Units bug. This returns an astropy Quantity object with true units of 1e-17 erg/(s cm2 Angstrom) but reported units without the 1e-17.

My preference would be to not use astropy units at all and just return `outflux * 1e17` and put in the docstring that the units are `1e-17 erg/(s cm2 Angstrom)`, though I recognize reasonable arguments for why to provide a units-ful object, especially at certain critical interfaces like this.  You also include units in the accompanying metadata table, which is good, which could be an additional argument in favor of using units here.

If you do want to include astropy units to return a Quantity object instead of a vanilla numpy array (your choice), then you shouldn't include the 1e17 factor here.  

i.e. don't try to force astropy to return something where `str(flux.unit).startswith('1e-17')`; astropy will always strip the prefactor for the units and apply it to the numbers to simplify the units.  e.g. this would have worked while carrying around unnecessary cancelations:
```python
scaledfluxunits = 1e-17 * u.erg / (u.cm**2 * u.s * u.Angstrom)
return (outflux*1e17) * scaledfluxunits
```
but even then, flux.unit would be `erg / (Angstrom cm2 s)` and the values would be O(1e-17)...

Same comment applies to other units-ful return values in other `make_templates` functions.

py/desisim/io.py
```diff
@@ -917,8 +917,11 @@ def empty_metatable(nmodel=1, objtype='ELG', subtype='', add_SNeIa=None):
                            data=np.zeros(nmodel)))
     meta.add_column(Column(name='MAG', length=nmodel, dtype='f4',
                            data=np.zeros(nmodel)-1))
-    meta.add_column(Column(name='DECAM_FLUX', shape=(6,), length=nmodel, dtype='f4'))
-    meta.add_column(Column(name='WISE_FLUX', shape=(2,), length=nmodel, dtype='f4'))
+    meta.add_column(Column(name='FLUX_G', length=nmodel, dtype='f4'))
+    meta.add_column(Column(name='FLUX_R', length=nmodel, dtype='f4'))
+    meta.add_column(Column(name='FLUX_Z', length=nmodel, dtype='f4'))
+    meta.add_column(Column(name='FLUX_W1', length=nmodel, dtype='f4'))
+    meta.add_column(Column(name='FLUX_W2', length=nmodel, dtype='f4'))
```
sbailey: You include unit metadata for the other columns in this table; let's do it for these new columns too.

moustakas:
This is ready for re-review and merging -- I think I addressed all the issues that were raised.  The failing test is a sphinx documentation problem that I'm having trouble debugging, but I'm out of time right now.  I'll fix this but if @weaverba137, @sbailey, or anyone else has insight into what I'm doing wrong I'd appreciate it!

moustakas:
Thank you for the documentation formatting changes, @sbailey.  Please merge when ready.

sbailey:
I fixed sphinx formatting, albeit in a clunky manner.  Enough to proceed, but we should revisit how to format our docstrings in a human and sphinx friendly way.  Merging now.

sbailey:
For the record: I merged this before remembering that I was going to check it with downstream integration tests too.  Those have probably broken now, but would need to be fixed anyway.  I can test some of this with NERSC still out, but other tests will need NERSC to come back (preferably Cori, which is also out tomorrow).  But we are probably in a state where master of everything isn't self-consistent.

PR #253, opened 2017-08-10
BGS Simulation parameter propagation

akremin:
Created new functions to specify target parameters in templates.py. Code now propagates a dictionary to get_targets and unpacks into make_templates. All defaults maintained. Old function calls should not see any changes.

### DIFF
--- a/py/desisim/obs.py
+++ b/py/desisim/obs.py
@@ -29,7 +29,7 @@ import desisim.simexp
 from .simexp import simulate_spectra
 
 def new_exposure(program, nspec=5000, night=None, expid=None, tileid=None,
-                 seed=None, obsconditions=None, testslit=False, exptime=None,
+                 seed=None, obsconditions=None, specify_targets=None, testslit=False, exptime=None,
                  arc_lines_filename=None, flat_spectrum_filename=None):
     """
     Create a new exposure and output input simulation files.
@@ -45,6 +45,9 @@ def new_exposure(program, nspec=5000, night=None, expid=None, tileid=None,
         tileid : integer tile ID
         seed : random seed
         obsconditions: str or dict-like; see options below
+        specify_targets: dict;  Define target properties like magnitude and redshift
+                                see simspec.templates.specify_galparams_dict() or 
+                                simsepc.templates.specify_starparams_dict()
         exptime: float exposure time [seconds], overrides obsconditions['EXPTIME']
 
         testslit : simulate test slit if True, default False; only for arc/flat
@@ -149,7 +152,8 @@ def new_exposure(program, nspec=5000, night=None, expid=None, tileid=None,
         return truth, fibermap, None, None
 
     #- all other programs
-    fibermap, (flux, wave, meta) = get_targets_parallel(nspec, program, tileid=tileid, seed=seed)
+    fibermap, (flux, wave, meta) = get_targets_parallel(nspec, program, tileid=tileid, \
+                                          seed=seed, specify_targets=specify_targets)
 
     if obsconditions is None:
         if program in ['dark', 'lrg', 'qso']:
--- a/py/desisim/obs.py
+++ b/py/desisim/obs.py
@@ -29,7 +29,7 @@ import desisim.simexp
 from .simexp import simulate_spectra
 
 def new_exposure(program, nspec=5000, night=None, expid=None, tileid=None,
-                 seed=None, obsconditions=None, specify_targets=None, testslit=False, exptime=None,
+                 seed=None, obsconditions=None, specify_targets=dict(), testslit=False, exptime=None,
                  arc_lines_filename=None, flat_spectrum_filename=None):
     """
     Create a new exposure and output input simulation files.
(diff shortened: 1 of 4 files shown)


py/desisim/targets.py
```diff
@@ -279,19 +282,28 @@ def get_targets(nspec, program, tileid=None, seed=None, specmin=0):
...
             fibermap['DESI_TARGET'][ii] = desi_mask.LRG
 
         elif objtype == 'BGS':
             from desisim.templates import BGS
             bgs = BGS(wave=wave)
-            simflux, wave1, meta1 = bgs.make_templates(nmodel=nobj, seed=seed)
+            if specify_targets is None:
+                simflux, wave1, meta1 = bgs.make_templates(nmodel=nobj, seed=seed)
+            else:
+                simflux, wave1, meta1 = bgs.make_templates(nmodel=nobj, seed=seed), **specify_targets
```
sbailey: Typo in where the parentheses close.  Please run "python setup.py test" locally before pushing to github to catch problems like this before Travis does.  Tests will run faster if you get the desisim-testdata first:
```
git clone https://github.com/desihub/desisim-testdata
source desisim-testdata/setup-testdata.sh
```

py/desisim/scripts/brightsims_aug2017.py
```diff
@@ -0,0 +1,137 @@
+"""
+desisim.scripts.brightsims
```
sbailey: OK to add this file for now, but we should be careful about the boundary between generally useful scripts (ok for this repo) vs. one-off scripts or functions that are tied to a particular analysis (better in a different repo custom for that analysis).  Proceed with this but we may refactor later when we know more about what is generally useful vs. bespoke.

py/desisim/targets.py
```diff
@@ -236,6 +236,9 @@ def get_targets(nspec, program, tileid=None, seed=None, specmin=0):
     Options:
         tileid: (int) tileid, used for setting RA,dec
         seed: (int) random number seed
+        specify_targets: dict;  Define target properties like magnitude and redshift
+                                see simspec.templates.specify_galparams_dict() or 
+                                simsepc.templates.specify_starparams_dict()
```
sbailey: Sigh; I think sphinx doesn't like this kind of multi-line syntax.  if you make each item a bullet list I think it will work, e.g.
```
Options:
    * tileid: ...
    * seed: ...
    * specify_targets: ...
                    see simspec...
    * specmin: ...
```

`python setup.py build_sphinx` can pre-test your docs before pushing, but note that you'll need sphinx 1.5 not latest 1.6.2.

Please try that out and holler for help if you can't get it working.

py/desisim/obs.py
```diff
@@ -29,7 +29,7 @@
 from .simexp import simulate_spectra
 
 def new_exposure(program, nspec=5000, night=None, expid=None, tileid=None,
-                 seed=None, obsconditions=None, testslit=False, exptime=None,
+                 seed=None, obsconditions=None, specify_targets=None, testslit=False, exptime=None,
```
sbailey: This implementation with `specify_targets` will work for mono-target programs like "BGS" but not multi-target programs like "BRIGHT" or "DARK" where you may want different selections for different types of templates.  I think we need something a bit messier but more general:

`new_exposure` has separate options for each target class with a default as a blank dict, e.g.
```
def new_exposure(..., bgs_kwargs=dict(), elg_kwargs=dict(), lrg_kwargs=dict(), ...)
```
and then down in targets.select_targets, each individual class would use it individual options,
```
simflux, wave1, meta1 = elg.make_templates(nmodel=nobj, seed=seed, **elg_kwargs)
simflux, wave1, meta1 = lrg.make_templates(nmodel=nobj, seed=seed, **lrg_kwargs)
...
```
by making the default a `dict()` instead of None, you can also avoid the if/else logic for whether any keyword args are specified or not.

akremin:
Hi Stephen,

Thank you for these comments. I'm a bit of a newbie when it comes to this
level of testing and sphinx. I'll try to get up to speed fast so I'm not
bogging you down with these basic failures.

Question - I see your point about applicability to the 'program' flavors of
new_exposure. How many of these kwargs did you imagine? If we go on
templates there are 8 different classes, or new_exposure has 7 objtype's
(if we include bad_qso).

A more compact but even more obscure way would be a dict of dicts with each
key being the object type used in new_exposure and the values being a dict
of their kwargs.

object_kwargs['BGS'] = bgs_kwargs        <--    where type(bgs_kwargs) =
> dict()


We can still have the default as an empty dict for object_kwargs. Then have
a loop in get_targets that assigns any of the objtype's without a
corresponding key to have an empty dict value:

for key in list_of_objtypes:
>    if key not in object_kwargs.keys():
>        object_kwargs[key] = dict()


Let me know your thoughts. Tomorrow morning I can make those edits and
submit another pull request after more deeply vetting for test failures.

Thank you again,
Anthony

sbailey:
I like the idea of `object_kwargs[objtype]`.  Go for it.  That makes it easier to adapt the code for future target classes too.

akremin:
All changes have been made and tests pass.

I'm happy to merge if you are.

PR #255, opened 2017-08-15
QA Polishing

profxj:
Changes from input at desi data telecon (15 Aug 2017):

* Added BGS, MWS to redshift QA
* Calculate S/N per Ang instead of per pix
* Change to velocity units for 2D redshift histograms
* Extend [OII] flux range for ELGs below requirement threshold
* Clip before calculating RMS of redshift deviates

Here is the BGS output:

dc17a2_zfind_bgs

### DIFF
--- a/py/desisim/spec_qa/redshifts.py
+++ b/py/desisim/spec_qa/redshifts.py
@@ -242,8 +242,9 @@ def load_z(fibermap_files, zbest_files, outfil=None):
     # Return
     return simz_tab # Masked Table
 
+
 def obj_requirements(zstats, objtype):
-    '''Assess where a given objtype passes the requirements
+    """Assess where a given objtype passes the requirements
     Requirements from Doc 318 (August 2014)
 
     Parameters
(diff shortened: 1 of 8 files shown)


py/desisim/spec_qa/redshifts.py
```diff
@@ -257,12 +258,14 @@ def obj_requirements(zstats, objtype):
...
         Pass/fail dict
-    '''
+    """
     log = get_logger()
     pf_dict = {}
     #
     all_dict=dict(ELG={'RMS_DZ':0.0005, 'MEAN_DZ': 0.0002, 'CAT_RATE': 0.05, 'EFF': 0.90},
         LRG={'RMS_DZ':0.0005, 'MEAN_DZ': 0.0002, 'CAT_RATE': 0.05, 'EFF': 0.95},
+                  BGS={'RMS_DZ':1.0000, 'MEAN_DZ': 1.0000, 'CAT_RATE': 1.00, 'EFF': 0.0}, # MAKE BELIEVE
+                  MWS={'RMS_DZ':1.0000, 'MEAN_DZ': 1.0000, 'CAT_RATE': 1.00, 'EFF': 0.0}, # MAKE BELIEVE
```
sbailey: Even though we don't have formal requirements for BGS and MWS, let's pick something semi-reasonable so that some errors are triggered if they fall apart.  e.g. same as LRG, or maybe a bit better if we're feeling optimistic (current sims are so optimistic that no one believes them...)

profxj: ok, have copied the LRGs.

profxj:
@sbailey -- Note that I have snuck in a change related to 
calculating the fiber area.

sbailey:
I updated Travis config to use desimodel/master to get the fiber area calculation.  After the next desimodel tag we can switch this to use that instead.  OK to merge as soon as we confirm that travis tests are passing.

profxj:
Passing travis.  Merging

PR #261, opened 2017-10-04
Fast fiber acceptance

julienguy:
Use the fast fiber acceptance of specsim in desisim (why two packages ...)

 - In desisim/simexp.py , for source_type = ELG or LRG , use DISK or BULGE source profiles with 0.45" and 1" half light radii. For BGS, I implemented the same parameters as for LRG. We can change this if this is not the best choice. We may also include there a range of half light radii correlated with magnitudes if needed. 

 - Added in bin/quickspectra a --source-type option to use the correct fiber aperture loss.

Code tested with quickspectra (S/N dropping by a factor 2.4 when ran with --source-type lrg or star )

### DIFF
--- a/py/desisim/scripts/quickspectra.py
+++ b/py/desisim/scripts/quickspectra.py
@@ -21,7 +21,7 @@ from desispec.spectra import Spectra,spectra_dtype
 from desispec.resolution import Resolution
 import matplotlib.pyplot as plt
 
-def sim_spectra(wave, flux, program, spectra_filename, obsconditions=None, expid=0, seed=0):
+def sim_spectra(wave, flux, program, spectra_filename, obsconditions=None, sourcetype=None, expid=0, seed=0):
     """
     Simulate spectra from an input set of wavelength and flux and writes a FITS file in the Spectra format that can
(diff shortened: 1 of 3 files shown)


julienguy:
The travis tests are failing because the master version of specsim is used, not the fast-fiber-acceptance branch ...

sbailey:
@moustakas could you provide median BGS parameters to use?
Typical half_light_radius (arcsec) and disk (exponential) vs. bulge (devaucouleurs)?

Otherwise looks good to merge.  Thanks!

julienguy:
I added BGS parameters based on the analysis of J. Moustakas for the SDSS main sample.
https://github.com/desihub/desitarget/blob/bgs-properties/doc/nb/bgs-morphology-properties.ipynb

    bulge_fraction = 0.48
    disk_fraction = 0.52
    bulge_half_light_radius = 1.3 arcsec
    disk_half_light_radius = 4.7 arcsec
    note we use De Vaucouleurs' law , which correspond to a Sersic index n=4 when the mean value for SDSS is 2.3

PR #262, opened 2017-10-05
Scaling improvements for pixsim

tskisner:
This work adds a new script / entrypoint "pixsim_nights" which processes one or more nights of simspec files in parallel.  It also includes optional MPI parallelization of pixsim itself for a single exposure.  In that case the number of processes per camera is specified, and the grouping of processes is done automatically based on the number of cameras per exposure and the total number available.

This was tested on a medium-sized sim, with 150 nodes (6 procs per node for memory reasons).  In that case there were 5 groups of 30 nodes working on 45 exposures.  Each group got 9 exposures, and for each exposure one node (6 processes) worked on each camera.  The PSF projection was done in parallel with MPI.


julienguy:
I tried to follow the MPI parallelization. It looks great.

One suggestion. It's not super important but anyways. 

The randomization of cosmic rays is done in desisim.io.read_cosmics.
It is called in desisim.script.pixsim by rank 0 **before** the selection of the random seed if I am not mistaken.
Also, an argument of desisim.io.read_cosmics is an exposure id which also has to be randomized. In the current version of the code it is the pixsim script argument expid. If I am not mistaken it means that all cameras from a given exposure see the same cosmics (with the additional 'jitter' performed in desisim.io.read_cosmics).

So I suggest we set the random seed before calling read_cosmic, and we also use a random parameter cosmics expid per simulated expid and camera (making sure it is the same parameter expid for all bundles)

tskisner:
Ok, that is a small fix, I'll do those small changes after my 11am telecon.  Also will look at why the travis tests failed.  Don't merge yet...

julienguy:
For Stephen, I checked the MPI parallelism is done for the projection of the true flux.
The randomization of pixel values is done by rank 0, so there is no problem of duplicated read noise.

tskisner:
Even if these checks pass, don't merge yet- I need to add some more logic to detect whether an exposure has completely finished, otherwise we have partial desi-*.fz files.

julienguy:
Ted, please do merge when ready.

PR #264, opened 2017-10-18
scale the lya spectra by 1e17

moustakas:
This PR fixes a minor bug whereby the QSO+Lya spectra generated by `lya_spectra.get_spectra` were not being scaled by `1e17`, causing an inconsistency in the "truth" spectra generated as part of the Data Challenge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,17 +63,17 @@ env:
         # These packages will only be installed for documentation builds.
         - CONDA_SPHINX_DEPENDENCIES="scipy pyyaml sphinx==1.5"
         # These packages will only be installed if we really need them.
-        - CONDA_ALL_DEPENDENCIES="scipy matplotlib coverage==3.7.1 pyyaml qt=4 healpy"
+        - CONDA_ALL_DEPENDENCIES="scipy matplotlib coverage pyyaml"
         # These packages will always be installed.
         - PIP_DEPENDENCIES=""
         # These packages will only be installed if we really need them.
-        - PIP_ALL_DEPENDENCIES="speclite==${SPECLITE_VERSION} coveralls fitsio"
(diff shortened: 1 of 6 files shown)


moustakas:
Tons of failing tests but I don't see how they're related to the 1-line change I made.  @sbailey can you take a look as the failures seem to be related to `new_exposure`...

sbailey:
I don't understand why these travis tests are failing.  The tests appear to be using astropy 1.3.3 which is what we have been using before and which works on my laptop.  Since the last successful test, default numpy version increased from 1.13.1 to 1.13.3.  I've added some debugging messages to get a bit of a better view into what is happening, but it is still guessing at this point.

[and thanks for the fix...]

sbailey:
@weaverba137 can you help with this?

Our travis configuration pins astropy=1.3.3, but something down the line is still forcing an "upgrade" to 2.0.2 which we don't support yet.  The portion of the raw log that is forcing that upgrade is surrounded by text about having difficulty finding healpy on conda and eventually giving up and falling back to a pip installation, but I can't tell if this astropy upgrade is a direct consequence of attempting to install healpy via conda or not.

I have temporarily turned off quickgen tests for faster turnaround, but test_obs will fail due to astropy Table units issues, and I dumped some explicit printouts of the table and astropy/numpy versions at the point of failure.  Search the raw logs for "DEBUGING TRAVIS" (alas, note the typo, not DEBUGGING).  After this is fixed we should revert that code and re-enable quickgen tests before merging.

weaverba137:
Acknowledged, I'll take a look.

weaverba137:
It looks like the fix was to move the healpy installation explicitly to a pip install.

moustakas:
Excellent, thanks for the sleuthing, @weaverba137.  @sbailey do you want to take out all the test code you inserted before merging?

PR #265, opened 2017-10-22
fixing MPI wrappers for pixsim

sbailey:
Submitting this PR on behalf of @tskisner after iterations with him and @julienguy about this branch and pixsim_fix (latest commit 54c4234) which addresses the same issue in a different way.

This fixes up the pixsim_nights_mpi wrapper around pixsim to work with multiple spectrographs and non-ideal packings of exposures into nodes.  It has been tested under various concurrencies and job sizes up to ~5 short nights of pixel-level simulations.  We'll proceed with this for larger runs for pipeline scaling tests.

### DIFF
--- a/bin/pixsim_mpi
+++ b/bin/pixsim_mpi
@@ -1,0 +1,22 @@
+#!/usr/bin/env python
+
+"""
+MPI Parallel version of pixsim
+"""
+
+from desispec.parallel import use_mpi
+
+comm = None
+if use_mpi:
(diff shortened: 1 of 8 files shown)


sbailey:
I just merged master into mpifix to get the same travis/healpy fix identified and fixed in PR #264; too many copies of this branch were in the wild to do a safe rebase, so this PR will appear to have a few non-pixsim changes that really came from that merge.

PR #266, opened 2017-10-23
Quickspectra

sbailey:
This PR makes several small improvements to quickspectra:
* uses wavemin/wavemax from `desimodel.io.load_params()` if those parameters are available, thus enabling quickspectra to work without specter.  This requires desihub/desimodel#68 to be merged and installed, or otherwise it will continue to use specter via desimodel.io.load_throughput() to get the wavemin/max values.
* zero pads input spectra if the input wavelength coverage doesn't span the full DESI wavelength range.  It prints a warning if it does this, but it no longer crashes due to insufficient coverage.  I think this is handy, but one could reasonably argue that it is dangerous and that that user should be forced to realize that their coverage is insufficient and do something about it themselves.
* Allows an input fits spectra file to have either WAVE or WAVELENGTH, thus enabling quickspectra to directly work from a simspec file output by newexp-random or newexp-mock.


julienguy:
I checked this works

PR #271, opened 2017-11-01
drop columns from tables used in quicksurvey

forero:
Small change to save RAM. Some columns are dropped in the tables used in `quicksurvey`.
Also forces garbage collection after deleting `mtl` tables.


forero:
Small change. Tests pass. Merging.

PR #275, opened 2017-11-09
Exposures as obslist

### LINKED ISSUE #273: Use exposures.fits file as input observation list
`wrap-newexp` needs to be set up to read an exposures.fits file produced by `desisurvey.progress.Progress.get_exposures()`.  This input is specifically designed to make certain standardization steps unnecessary.

This issue will track some specific remaining questions prior to creating a PR:

1. If we write out the exposures list, after flats and darks are added, what do we call the file?  exposures.fits, but in a different directory?
2. When writing out the exposures list, the 'PROGRAM' for flats and darks is 'FLAT' and 'DARK', correct?
3. Did we already have specific dummy values in mind (for darks and flats) for things like 'MOONSEP', or would any arbitrary (but obviously dummy) value work?

weaverba137:
This PR closes #273 by deriving all input for `wrap-newexp` from the input obslist/exposures file.

### DIFF
--- a/bin/pixsim_mpi
+++ b/bin/pixsim_mpi
@@ -1,22 +1,18 @@
 #!/usr/bin/env python
 
 """
-MPI Parallel version of pixsim
+Simulate DESI spectrograph CCD images given input spectra and PSF
+
+Stephen Bailey, LBL
+January 2015
 """
-
-from desispec.parallel import use_mpi
-
-comm = None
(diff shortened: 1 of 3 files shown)


sbailey:
Forgot to mention: I accidentally added a `pixsim_mpi` script as part of this update.  That is unrelated to this PR, but also shouldn't get in the way.

PR #284, opened 2017-11-28
check whether output file exist or not

julienguy:
Single change to wrap-fastframe: checking for pre-existing output if option --overwrite not set


julienguy:
argh stephen had already done this, forget about it ...

PR #285, opened 2017-11-29
notebook to construct BGS reference spectra

moustakas:
The notebook `bgs-reference-spectra.ipynb` in this PR constructs a set of reference spectra for the BGS exposure-time model (see https://github.com/desihub/desisurvey/issues/77).

The spectra themselves are at NERSC in 
  `/global/project/projectdirs/desi/spectro/sim/bgs-refspec/bgs-refspec-v1.0.fits`


sbailey:
Looks good to me; feel free to merge or wait for @dkirkby as you see fit.

moustakas:
@dkirkby I just realized that these spectra may not have enough wavelength coverage.  They're sampled from ~3500-1e4 A in the observed frame, and if you're going to be deredshifting and reredshifting them you probably need them to go bluer.  Right?

moustakas:
I went ahead and changed the observed-frame wavelength range to 2000-10000 A. Files at NERSC have been updated with the same version number.

dkirkby:
I need a slightly higher upper limit of 10990A to cover the DECam r-band filter curve limits (3330,  10990) at z=0.

moustakas:
Done.  I went ahead and went bluer, too, just in case we want to synthesize u-band photometry.  The wavelength range is now 0.15-1.3 micron and the NERSC file has been updated (same version number).

dkirkby:
This looks great @moustakas, thanks!

Should I worry that the redshift distribution has some spikes and, in detail, does not match the targeting n(z)?

I am confused by the flux normalization.  Typical fluxes seem to be O(1e17) when I expected O(1) since HDU[1] has:
```
BUNIT   = '1e-17 erg / (Angstrom cm2 s)'
```
I think this is just a normalization issue since I can reproduce your `FLUX_R` column if I rescale fluxes by 1e-17.

Have you already normalized each spectrum to represent a suitable "threshold" sample or should I normalize to some fixed value, e.g., r=19.5?

dkirkby:
Here are histograms of the redshift and r-mag distributions for the refspec sample, for reference:

refspec

I am worried that the tail of bright objects are not really "threshold" and will skew the median SNR.

moustakas:
Sorry for the normalization issue -- I had forgotten that `desisim.templates` already scales by 1e17 and so I did it again right before writing out.  I'm fixing this now.

Regarding n(z) -- the spikes are just the large-scale structure in the BGS mock, so in a sense this is a feature not a bug.  Although these overdensities are present---and we basically have to live with them---the advantage from using the mock is that we inherit the appropriate correlations between absolute magnitude and rest-frame color.

I just saw the histograms you posted -- which I should have included at the end of the notebook.

As you've seen I have *not* rescaled the spectra to a threshold r=19.5, but you *should*.  The main point of what I've done is to get spectra with the correct SED (e.g., 4000-A break strength) to lie in the green valley, where we think we'll have to work hardest to get redshifts.  I left the observed-frame spectra as-is so they could be used for multiple purposes (e.g., rescaling to r=19.5 and, separately, r=20). 

And ditto for the redshifts -- there's nothing crucial about these redshifts.  I would simply take this set of spectra, de-redshift and re-redshift after sampling from the smooth BGS n(z) and then renormalize to whatever faint magnitude cutoff you need.

I hope this makes sense!

moustakas:
Notebook and spectra on-disk have been updated (same filename / version number).  I'll be offline for a few hours but can fix any further issues later this afternoon but if everything looks good please feel free to merge.

dkirkby:
Thanks, this makes sense.  For a first pass, I will leave the redshifts as they are and rescale to DECam-r = 19.5.

If I end up resampling the redshifts later (e.g., to reproduce the bright/faint n(z) curves), I would need a higher wavemax since the current limits only allow me to reduce the redshift.

moustakas:
OK, let me write the spectra out over the full wavelength range of the templates (they go out to ~5 microns or so -- I don't remember exactly).

moustakas:
@dkirkby The reference spectra now extend to 2 microns, which should be sufficient for any redshiftting / deredshifting over the full BGS redshift range.  For example, a reference spectrum at z=0.8 could be deredshifted to z=0 and still cover the full r-band filter curve.

The file at NERSC has been updated (with the same version number -- with any subsequent changes I'll increment the version).

PR #286, opened 2017-11-30
BGS redshift efficiency under nominal (moon down!) conditions

moustakas:
The [notebook](https://github.com/desihub/desisim/blob/bgs-efficiency/doc/nb/bgs-redshift-efficiency.ipynb) in this PR builds on lots of previous work by me, @akremin, @sbailey, and @julienguy.

Here, I attempt to quantify the redshift efficiency of BGS targets under "nominal" conditions (i.e., moon-down, 5 minute exposures).  The results are just based on 1000 simulated spectra so far, so there's small-number statistics to contend with, but the results are nevertheless interesting.

This work is closely related to https://github.com/desihub/desisurvey/issues/77, so input from @dkirkby would be welcome.


dkirkby:
This looks good!  I haven't gone through in detail, but it looks like you have simulated an ~unbiased sample of 1K BGS bright+faint targets and run them through redrock (which is taking ~3mins per spectrum on 4 cores?)

Is the ~linear drop in efficiency with r-mag > 19 expected?  Efficiencies < 50% for the BGS-faint program are sobering.

I didn't expect the efficiency to be flat in redshift, due to the improvement in fiberloss with DA(z), but perhaps you aren't including this effect or there is a compensating effect?

Can you plot the efficiency vs median(SNR) in each camera, since that's closest to the "redshift success" proxy I am using for the exposure time model.

There is a canned plotting routine `desiutil.plots.plot_slices` that you could use for your `qa_dz()` function:  http://desiutil.readthedocs.io/en/latest/api.html#desiutil.plots.plot_slices

moustakas:
Thanks for the comments, @dkirkby. It's definitely a work in progress -- I'm building more QA that will help us understand what's going on.  For example, regarding the significant drop in efficiency starting around r~19.  It looks like nearly all (92%!) of the "lost" redshifts are at the correct redshift but have ZWARN==4, which is the "small_delta_chi2" flag. So I'm optimistic that we'll be able to recover these by staring at the chi2 vs redshift curves (volunteers are welcome).

Regarding the fiberloss:  I'm using whatever is implemented in `quickspectra`, which looks like it's the `fastsim` option but with fixed half-light radii and Sersic indices (i.e., no redshift dependence). It would be good to get the redshift dependence you're incorporating into this piece of code (`desisim.simexp.simulate_spectra`).  Or perhaps @julienguy has another suggestion?

As for the time -- there are 100 spectra per "chunk" of files so `redrock` is taking ~1.8 sec/spectrum with 4 cores (on my laptop).  That's fast!

moustakas:
Here's the current redshift efficiency curve alluded to, for reference.
bgs_eff

julienguy:
Just for confirmation, am I understanding correctly that the difference of efficiency you show with what I simulated is due to the moon? 
My simulations had MOONALT=60 MOONFRAC=0.7 MOONSEP=50, and I got an efficiency dropping at lower magnitudes , with ~80% at rmag=19 for a 300s exposure.

moustakas:
Right, my simulations are moon down / no moon, which is what @dkirkby needs (in https://github.com/desihub/desisurvey/issues/77).  I also have the "right" input distribution of redshifts and magnitudes because I tie the galaxy templates to the MXXL/BGS mock whereas I think you used a uniform input distribution of magnitudes and redshifts?

Related:  In this notebook I should also derive the redshift efficiency with the set of BGS reference spectra generated in https://github.com/desihub/desisim/pull/285 and scaled to r=19.5 so we can tune the exposure time to achieve the redshift efficiency we need / want.

But suggestions for dealing with galaxies with `ZWARN==4` would be helpful.

moustakas:
This PR just adds a notebook, so I'd like to merge even though Travis tests have not yet passed.

PR #287, opened 2017-11-30
add option to directly write cframes

julienguy:
Enable fastframe and wrap_fastframe to directly write cframe files.

This is convenient to study the variation of redshift efficiency with or without the pipeline fiberflat, sky subtraction and flux calibration.


sbailey:
Looks good, but while checking this I noticed that fastframe is writing incorrect BUNIT keywords, e.g. 'Angstrom' for the 'FLUX' HDU.  This wasn't caused by this PR, but let's go ahead and fix it here before it is merged.  Make sure to include the 1e-17 factor in the flux units when using the `--cframe` option.

Issue #291, opened 2017-12-01
Lya forest is added twice in lya_spectra


andreufont:
desisim.lya_spectra is supposed to add lya forest fluctuations from a file, but it currently calls 
qso.make_templates() without specifying lyaforest=False, what means that it gets a quasar template that already has (random, 1D) fluctuations. 
Then it reads the lya file, and adds extra fluctuations. 
I think this is trivially fixed by calling instead
qso.make_templates(lyaforest=False)

moustakas:
Yes, nice catch. Would you mind fixing this?

andreufont:
sure, I'll give it a try :-)

andreufont:
Fixed in PR #292

PR #292, opened 2017-12-01
do not add random lya forest when also reading lya forest from file

andreufont:
Tiny pull request to change the default value of lyaforest in desisim.lya_forest. @moustakas, I guess it should be trivial to merge, right? It fixes issue #291


moustakas:
I was waiting for tests to complete but it's now been 40 minutes, so I'm going to go ahead and merge. We really need to speed up our unit tests here...

PR #297, opened 2017-12-08
Lya3d

julienguy:
Modifications to lya_spectra.py: two new routines
 - read_lya_skewers
 - apply_lyman_alpha_transmission

get_spectra is left untouched. get_spectra is not used with the new mock format.
the idea is that the mock simulation (in desitarget.mock.spectra) calls sequentially
- read_lya_skewers
- QSO.make_template (or any new version of this class)
-  apply_lyman_alpha_transmission

it is in desitarget.mock.spectra that we should add DLAs depending on a option keyword
in select-mock-targets.yaml (not implemented yet)

See the PR in desitarget lya3d for more informations.


sbailey:
I did some cleanup to add docstrings, remove unused imports, and standardize "lya" vs. "lyman_alpha".  OK to merge once tests pass.

Longer term we still need to cleanup get_spectra vs. calling the 3 functions, and how all that interacts with desitarget, but I'll leave that for a different update.

PR #303, opened 2018-01-02
various updates to templates.SIMQSO

moustakas:
This PR contains a variety of updates to the SIMQSO class, including the ability to (re)generate spectra from an input QsoSimObjects object (which can also optionally be rebuilt from a FITS HDU -- no pickling required) and some minor suggested updates from @sbailey.


sbailey:
Looks fine except that the flagship feature of this PR isn't included in the `SIMQSO.make_templates` docstring.  if `return_qsometa=True`, does it return QSO metadata in addition to the normal `meta` table or instead of it?  If it returns two metadata tables, does one recreate the results by passing both tables back in (via `input_meta` and `input_qsometa`?).

I'll go ahead and merge this now, but please make a docstring update to master with "[skip ci]" in the commit string so that we don't have to wait for another round of Travis tests before merging.

PR #304, opened 2018-01-04
fix bug in pixsim when simulating more than one spectro 

julienguy:
- Default option value nspec=500 was breaking simulations of more than one spectrograph. This bug was introduced recently.
- It's a two line change to the code but I submit a PR anyway in case this fix is posing other difficulties.


sbailey:
I'll go ahead and merge this because it re-enables the ability to simulate spectrograph >0, while noting that the option still doesn't work as described "Number of spectra to simulate per camera".  See issue #305 for tracking this.

PR #310, opened 2018-01-15
optionally do not wavelength resample the QSO templates 

moustakas:
This PR is tangentially related to #297.  It adds the option of not resampling the QSO templates in wavelength, which may be necessary when applying the Lya forest transmission *outside* (i.e., after) making the templates (also softly related to https://github.com/desihub/desitarget/issues/252).

This PR also adds the documentation requested by @sbailey in #303.  See also cells [32-38] of the [simqso-templates notebook](https://github.com/desihub/desisim/blob/master/doc/nb/simqso-templates.ipynb) for a demonstration of how the `qsometa` object can be used to regenerate simqso templates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ env:
         # - SCIPY_VERSION=0.16
         - ASTROPY_VERSION=1.3.3
         - SPHINX_VERSION=1.5
-        - DESIUTIL_VERSION=1.9.8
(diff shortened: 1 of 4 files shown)


moustakas:
The Travis test failure is something I haven't seen before.  Thoughts?

screen shot 2018-01-15 at 1 40 24 pm

weaverba137:
I think this is fixed by using the latest version of desiutil.

PR #312, opened 2018-01-19
changes to speed up MPI by reducing amount of broadcast data

lastephey:
Changed MPI bcast to Bcast to take advantage of faster numpy arrays vs slower python objects. Instead of preallocating as 'None', this required a little more work to correctly pre-allocate the empty numpy broadcast arrays. 

Speed testing is still in progress, but on a single node for a single processor (Haswell), the runtime for the MPI version is approximately 7 percent slower than the multiprocessing version. As the number of processes grow, MPI becomes increasingly slower than multiprocessing due to overhead. This may need to be improved in the future with functions like MPI scatter and gather, rather than broadcast.

### DIFF
--- a/py/desisim/io.py
+++ b/py/desisim/io.py
@@ -298,15 +298,19 @@ class SimSpec(object):
     """
     def __init__(self, flavor, wave, phot, flux=None, skyflux=None,
                  skyphot=None, metadata=None, fibermap=None, obs=None, header=None):
-        for channel in ('b', 'r', 'z'):
-            assert wave[channel].ndim == 1
-            assert phot[channel].ndim == 2
-            assert wave[channel].shape[0] == phot[channel].shape[1]
 
-        assert phot['b'].shape[0] == phot['r'].shape[0] == phot['z'].shape[0]
+        #change to a new testing procedure since channel is now an input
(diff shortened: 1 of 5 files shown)


py/desisim/io.py
```diff
@@ -387,24 +388,100 @@ def read_simspec_mpi(filename, comm, spectrographs=None):
     # Note: this is global scope within the function, so memmap file handle
     # will not close until the function returns (which is fine).
     hdus = None
+    hdus = fits.open(filename, memmap=True)
```
sbailey: Please update so that only rank 0 will open the fits file, broadcast the shapes to all the other ranks, and then proceed with broadcasting the data.  As currently written, all ranks will open the same file and scan its headers to get the shapes, potentially bogging down the I/O server.

Otherwise these changes look fine.  Thanks.

lastephey: Good catch. Changes made, some re-arranging was required.

sbailey:
For the record: this PR considerably helped with running with a single process, but didn't help much when running with all cores on a node.  We've identified why and are chasing that, so this PR is on hold.  We might supersede it or we might expand it.

PR #317, opened 2018-01-25
photons/bin -> photons/Angstrom and optional dwave with default = 0.8A

julienguy:
change of units for simulated frames

### DIFF
--- a/py/desisim/scripts/fastframe.py
+++ b/py/desisim/scripts/fastframe.py
@@ -118,19 +118,18 @@ def main(args=None):
             meta = simspec.header.copy()
             meta['CAMERA'] = camera
             if args.cframe :
-                units = '1e-17 erg/(s cm2 A)'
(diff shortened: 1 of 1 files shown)


sbailey:
Standardized "electrons/Angstrom" -> "electron/Angstrom" and fixed prior non-standard "1e-17 erg/(s cm2 A)" -> "1e-17 erg/(s cm2 Angstrom)" (Although it was clear to a human what we meant, A is for Ampere not Angstrom in SI...).

For some reason the previous code only set the BUNIT if it already existed as a keyword in the inherited header; I changed that to always set the output BUNIT.

Also noting that this PR changes the default output fastframe wavelength grid from 1.0 A bins to 0.8 A bins.  OK.  I probably need to increase my disk space estimates by 20%...

OK to merge after tests pass

julienguy:
allright so I have to do the modification  electrons/Angstrom -> electron/Angstrom on the desispec side as well  ...

PR #320, opened 2018-02-02
fix double PSF convolution in pixsim

### LINKED ISSUE #318: PSF is sharper in fast 1D simulations than in 2D spectroperf. extractions
The PSF in the fast 1D simulations are sharper than the ones obtained with the 2D spectroperf. extractions.
Example: sky line in blue camera. both 1D and 2D have a wavelength step of 0.8A.

figure_1

This has to be fixed for redshift efficiencies estimates and meaningful comparison with simulations starting from CCD images.

sbailey:
This PR does the minimal change to fix #318, where @julienguy noticed that the pixsims were getting double convolved by the PSF.  The fix is to use `specsim.simulator.Simulator(camera_output=False)` and write the unconvolved electrons to the simspec output files.

fastframe uses the FLUX (not PHOT) HDUs so isn't impacted.  This fix is fragile since it risks us making the opposite mistake later and thinking that the simspec PHOT HDUs are convolved when now they aren't.  A better long term solution would be to store just the FLUX in the simspec files, and re-run specsim on-the-fly for pixsim as is done for fastframe.  However, @lastephey is in the middle of refactoring the MPI pixsims to improve it's performance and I want to let that settle before making another change on top of it.

Comparison plot showing that the new simspec CCD photons are not convolved, how much better the sky lines appear in the 2D images, and how the fastframe outputs are not impacted:
convolve_comparison

For the record:
```
#- Using master:
newexp-random --program elg --nspec 5 --seed 1
#- Then using pixsim-no-convolve branch:
newexp-random --program elg --nspec 5 --seed 1

fastframe --simspec simspec-00000000.fits --outdir .
fastframe --simspec simspec-00000001.fits --outdir .

pixsim --psf /Users/sbailey/desi/git/desimodel/svn/trunk/data/specpsf/psf-r.fits --simspec simspec-00000000.fits --rawfile desi-0.fits.fz --cameras r0 --nspec 5 --wavemin 6280 --wavemax 6320
pixsim --psf /Users/sbailey/desi/git/desimodel/svn/trunk/data/specpsf/psf-r.fits --simspec simspec-00000001.fits --rawfile desi-1.fits.fz --cameras r0 --nspec 5 --wavemin 6280 --wavemax 6320

#- from ipython --pylab
from astropy.io import fits
import desisim.io
import desispec.io

s0 = desisim.io.read_simspec('simspec-00000000.fits')
s1 = desisim.io.read_simspec('simspec-00000001.fits')
r0 = fits.getdata('desi-0.fits.fz')
r1 = fits.getdata('desi-1.fits.fz')
f0 = desispec.io.read_frame('frame-r0-00000000.fits')
f1 = desispec.io.read_frame('frame-r0-00000001.fits')

clf()
subplot(311)
plot(s0.wave['r'], s0.skyphot['r'][0], label='master')
plot(s1.wave['r'], s1.skyphot['r'][0], label='new')
xlim(6280, 6320)
text(6282, 20000, 'simspec PHOT')
legend()

subplot(323)
imshow(r0[1290:1340, 170:220])
text(5, 40, 'master', color='y')
xlabel('master')

subplot(324)
imshow(r1[1290:1340, 170:220])
text(5, 40, 'new', color='y')
xlabel('new')

subplot(313)
plot(f0.wave, f0.flux[0], label='master')
plot(f1.wave, f1.flux[0], '--', label='new')
xlim(6280, 6320)
text(6282, 20000, 'fastframe output')
legend()
```

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,7 @@ desisim change log
 0.24.1 (unreleased)
 -------------------
 
-* No changes yet.
+* Fix double PSF convolution in pixsims.
 
 0.24.0 (2018-01-30)
 -------------------
--- a/py/desisim/io.py
+++ b/py/desisim/io.py
@@ -184,9 +184,9 @@ def write_simspec(sim, truth, fibermap, obs, expid, night, outdir=None, filename
     hx.append(hdu_skyflux)
 
     #- DEPRECATE?  per-camera photons (derivable from flux and throughput)
-    for i, camera in enumerate(sorted(sim.camera_names)):
-        wavemin = sim.camera_output[i]['wavelength'][0]
-        wavemax = sim.camera_output[i]['wavelength'][-1]
+    for i, camera in enumerate(sim.camera_names):
+        wavemin = sim.instrument.cameras[i].wavelength_min.to('Angstrom').value
+        wavemax = sim.instrument.cameras[i].wavelength_max.to('Angstrom').value
         ii = (wavemin <= wave) & (wave <= wavemax)
         hx.append(fits.ImageHDU(wave[ii], name='WAVE_'+camera.upper()))
 
--- a/py/desisim/obs.py
+++ b/py/desisim/obs.py
@@ -141,7 +141,8 @@ def new_exposure(program, nspec=5000, night=None, expid=None, tileid=None,
 
         if exptime is None:
             exptime = 10
-        sim, fibermap = desisim.simexp.simflat(infile, nspec=nspec, exptime=exptime, testslit=testslit)
+        sim, fibermap = desisim.simexp.simflat(infile, nspec=nspec,
+            exptime=exptime, testslit=testslit, camera_output=False)
 
         header['EXPTIME'] = exptime
         header['FLAVOR'] = 'flat'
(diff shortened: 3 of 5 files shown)


sbailey:
Hmm.  Hold off on merging this... now it appears that the pixsims are too good compared to the specsims:
pix-extract
Let's study this a bit more [tomorrow... after review pre-homework...]

julienguy:
The spectroperf math is a convolution only if the pixels have the same weight. In the case of bright emission lines, the Poisson noise in the lines results in a sharper line profile for the "reconvolved" flux = R.f
The estimator of the original "unconvolved" flux f is still linear so there is nothing fundamentally worrying here.
Example with a 1D implementation of spectroperf.
 1) Same weight (same inverse variance) for all pixels. The profile of R.f is the same as the instrument resolution (blue: true flux density , red dashed:  flux f , red solid: "reconvolved" flux R.f  )

figure_1-2

 2) pixel variance dominated by Poisson noise. The profile of R.f appears sharper than the instrument resolution ( but again, R is just a convenient transformation of the measurement f, so it's not a problem as long as data are always analyzed with with the R resolution matrix).

figure_1-3

julienguy:
Using the example shown by Stephen comparing fastframe sims and the (new) pixel sims,
I have fit an emission line on both set spectra **using their own resolution matrix**
(I have also averaged the results of 5 fibers to reduce the noise which)
I obtain
fastframe sim. flux = 58029.619895 +- 120.891546 , s/n = 480.013880
pixel sim. flux = 56134.002547 +- 118.836793 , s/n = 472.362144
A difference of S/N of 1.7% which is perfectly fine,

so, I think we can merge this.

PR #321, opened 2018-02-02
added BAL QSO utilities

paulmartini:
This will optionally add BAL features to simulated QSOs. The main TBD is to propagate the BAL index into meta. I'd appreciate input about how best to do that. For example, should there be a new column?


sbailey:
Import syntax needs to be updated for 3.x; "from bal import ..." doesn't work and tests are failing.  I think you need "from .bal import ..."?

paulmartini:
I've updated how the bal import. Please see if this now works.

py/desisim/bal.py
```diff
@@ -0,0 +1,108 @@
...
+
+"""
+from __future__ import division, print_function
+
+import numpy as np
+from astropy.io import fits 
+import random 
+
+def readbaltemplates(file):
+    """ Read in the BAL templates fits file
```
moustakas: `file` used to be a built-in function (in python 2.x), although it isn't anymore (in 3.x).  Nevertheless, my EMACS buffer thinks it's reserved(!), so to avoid confusion you might consider using a different name (e.g., `balfile`).

py/desisim/bal.py
```diff
@@ -0,0 +1,108 @@
...
+	booleal: True if a BAL, otherwise False
+    """ 
+
+    # Make sure probability is on the interval [0.,1.]
+    if balprob > 1.:
+        balprob = 1. 
+    if balprob < 0.: 
+        balprob = 0.
+
+    if np.random.random() < balprob:
```
moustakas: Here and in `getbaltemplate` you should use an input seed and a `numpy.random.RandomState` object to ensure the generated templates are reproducible.  See, for example, [these lines](https://github.com/desihub/desisim/blob/master/py/desisim/templates.py#L543-L544) in templates.py.

You should also write a unit test to ensure the reproducability works as advertised (see, e.g., [this test](https://github.com/desihub/desisim/blob/master/py/desisim/test/test_templates.py#L55-L66), as an example).

py/desisim/bal.py
```diff
@@ -0,0 +1,108 @@
...
+        return True
+    return False
+
+
+def getbaltemplate(qsowave, qsoredshift, balwave, baltemplates):
+    """ Get BAL template for a single QSO.
```
moustakas: I wonder if a lot of the functionality here can be done *once* by instantiating a class, so that when we generate many templates we don't have to redo many of these calculations (e.g., extending the template in wavelength and adding in the CIV template).

For example, the [SUPERSTAR](https://github.com/desihub/desisim/blob/master/py/desisim/templates.py#L1014) class may be a useful model.

py/desisim/templates.py
```diff
@@ -1811,6 +1816,10 @@ def make_templates(self, nmodel=100, zrange=(0.5, 4.0), rmagrange=(20.0, 22.5),
 
             if mag is None:
                 mag = rand.uniform(rmagrange[0], rmagrange[1], nmodel).astype('f4')
+	# Read in the BAL templates if needed. 
+        if bal: 
+            baltemplatefile = "/global/project/projectdirs/desi/spectro/templates/basis_templates/basis_templates_svn/trunk/BAL-templates-v0.1.fits" 
+            balwave, baltemplates = bal.readbaltemplates(baltemplatefile)
```
moustakas: The location of the file definitely shouldn't be hard-coded.  Please use an updated version of `desisim.io.read_basis_templates` which makes use of the `DESI_BASIS_TEMPLATES` environment variable in order to find the underlying basis spectra.

py/desisim/templates.py
```diff
@@ -1747,6 +1747,10 @@ def make_templates(self, nmodel=100, zrange=(0.5, 4.0), rmagrange=(20.0, 22.5),
             template that also satisfies the color-cuts (default 20).
           uniform (bool, optional): Draw uniformly from the PCA coefficients
             (default False).
+          bal (bool, optional): Include broad absorption line (BAL) features
+            (default False).
+          balprob (float, optional): Probability than a QSO is a BAL 
+            (default 0.12).
```
moustakas: Documentation typo: "than" --> "that"

py/desisim/templates.py
```diff
@@ -1901,6 +1916,10 @@ def make_templates(self, nmodel=100, zrange=(0.5, 4.0), rmagrange=(20.0, 22.5),
                     flux[kk, :] = np.dot(self.eigenflux.T, PCA_rand[:, kk]).flatten()
                     if redshift[ii] > 2.39:
                          flux[kk, :pix912] *= np.exp(-phys_dist.value / mfp[ii])
+                    if bal: 
+                        if bal.isbal(balprob): 
+                            baltemplate, balindex[kk] = bal.getbaltemplate(zwave[:,ii], redshift[ii], balwave, baltemplates)
+                            flux[kk, :] *= baltemplate
```
moustakas: Do the BAL templates need to be computed one-by-one, in the loop, or can they be pre-computed in a way that's vectorized?  If all you need is the redshift and a probability that a given QSO has a BAL then I *think* you should be able to pre-generate these spectra, which should speed up this step measurably.

py/desisim/bal.py
paulmartini: Done

py/desisim/bal.py
paulmartini: Done

py/desisim/bal.py
paulmartini: I've changed the templates so they no longer need to be extended in wavelength each time.

py/desisim/templates.py
paulmartini: I've fixed this.

py/desisim/templates.py
paulmartini: The complete templates are now read in from the input file, which should speed this up.

paulmartini:
Hi @moustakas, thanks for your comments. I think I've addressed them all, with the exception of the one about SIMQSO(). I'd like to get the current bal branch merged with master first, and then I (or maybe someone else) takes on the implementation in SIMQSO().

julienguy:
I am going make the changes necessary for the Travis tests to pass (an import issue).

moustakas:
I'm reviewing this now but note that it's not yet connected to `SIMQSO` and therefore won't work with `quickquasars` as-is.

moustakas:
I'm reviewing this now but note that it's not yet connected to `SIMQSO` and therefore won't work with `quickquasars` as-is.

Also, I need to add the BAL templates to the basis templates and cut a new tag.

py/desisim/templates.py
```diff
@@ -1812,6 +1817,10 @@ def make_templates(self, nmodel=100, zrange=(0.5, 4.0), rmagrange=(20.0, 22.5),
 
             if mag is None:
                 mag = rand.uniform(rmagrange[0], rmagrange[1], nmodel).astype('f4')
+	# Read in the BAL templates if needed. 
+        if balqso: 
+            baltemplatefile = os.environ['DESI_BASIS_TEMPLATES'] + "/../../trunk/BAL-templates-v0.2.fits"
+            balwave, baltemplates = bal.readbaltemplates(baltemplatefile)
```
moustakas: The location of the templates can't be hard-coded like this -- I will add them to $DESI_BASIS_TEMPLATES, desisim-testdata, and bump the version number.

Also, the templates should be read in *once* in `QSO.__init__` and cached.

moustakas:
[I had written the comment below a couple days ago but never submitted it.  Blarg.  Sending it now for the record.]

I'm reviewing this now but note that it's not yet connected to `SIMQSO` and therefore won't work with `quickquasars` as-is.

Also, I need to add the BAL templates to the basis templates and cut a new tag.

py/desisim/templates.py
```diff
@@ -1827,8 +1836,13 @@ def make_templates(self, nmodel=100, zrange=(0.5, 4.0), rmagrange=(20.0, 22.5),
...
                                (redshift, mag, templateseed)):
             meta[key] = value
-        if lyaforest:
+        if lyaforest: 
             meta['SUBTYPE'] = 'LYA'
+        if balqso: 
+            if lyaforest: 
+                meta['SUBTYPE'] = 'LYA+BAL'
+            else: 
+                meta['SUBTYPE'] = 'BAL'
```
moustakas: @paulmartini Shouldn't the SUBTYPE be set to LYA+BAL only if the spectrum contains a BAL, i.e., after `balprob` (the BAL probability) has been applied?

julienguy: I suggest we merge as soon as the template file issue is solved. This PR has been opened more than 3 month ago; the basic functionalities are in place and the Lya WG would need this to move on: this code is needed to test the BAL identifiers, the impact on redshift efficiency, correlation function ... etc.

moustakas: This code is not ready to merge yet and--again--it is not holding up any of the Lya QSO activities because this PR is not integrated with SIMQSO.  I'll merge as soon as it's ready but right now the output spectra are not reproducible even with an input seed, which needs to be resolved.

julienguy: BALs are integrated in quickquasars in the branch qqso_dla_bal which has been branched from this one. We are now discussing about adding metals. We have some sort of a branch traffic jam here.

paulmartini: I opted to set SUBTYPE to 'LYA+BAL' if balqso=True for the simulated quasars, even though not all quasars will be BALs (unless balprob=1). My rationale was that someone may want to know if the balqso flag were True when a set of simulated quasars were generated. I intended to eventually create a separate index (to be stored in META) that would either have the template ID (if a BAL) or be empty, but I have not figured out how to do that yet.

paulmartini: @moustakas With regard to @julienguy 's comment, there is another group (not me) in the Ly-a WG who will add BALs to SIMQSO. I think their work will be easier if the issues with this implementation are resolved first.

alxogm: Julien, I can comment out the pieces for the BALs and move to another branch that does not depend on this one until this is completed... I branched from the BAL's branch only because I was adding the BALS, but the other pieces don't need this...Also, maybe for quickquasars we don't need to add the BALs at the level of SIMQSO, could it be added to the transmission as we did with the DLAs?

moustakas:
I just tagged a new version of the basis templates which includes the BAL templates.  To get the latest set (e.g., on your laptop) you need to do:
```
svn export svn+ssh://{yourusername}@dtn01.nersc.gov/global/project/projectdirs/desi/spectro/calib/svn/basis_templates/tags/v2.6
```
On NERSC you'll either need to wait until desisim/master gets updated or do
```export DESI_BASIS_TEMPLATES=/global/projecta/projectdirs/desi/spectro/templates/basis_templates/basis_templates_svn/tags/v2.6```
yourself.

I made significant simplifications and speed improvements to the code which should make it trivial to pull out the relevant pieces and put them into `quickquasars` or any other script.  Please take a look.

I also added a brief notebook which demonstrates how to add BALs to QSO spectra --
  https://github.com/desihub/desisim/blob/bal/doc/nb/bal-templates.ipynb
However, there's a reproducibility issue (see the last few cells of the notebook) that I don't understand.  Others working on BALs can perhaps take a look.

There are other small desired improvements to the code (including the need for unit tests) but they should not block this PR further.  I'll open separate issues and merge this when Travis tests pass.

alxogm:
@moustakas this link is not longer working
https://github.com/desihub/desisim/blob/bal/doc/nb/bal-templates.ipynb 
can you point me to the location?

sbailey:
@alxogm see https://github.com/desihub/desisim/blob/master/doc/nb/bal-templates.ipynb

i.e. "bal" branch -> "master" branch in the URL now that it has been merged into master.

Issue #289, opened 2017-11-30
fiber acceptance fraction for BGS targets should depend on redshift


moustakas:
The "fast" fiber acceptance (or fiberloss) fraction for BGS targets coded up in `simexp.simulate_spectra` is currently a constant / fixed, but the redshift dependence is non-negligible (more than a factor of two) due to the cosmological dependence of the angular diameter distance.  We should fix this.

The relevant portion of the code is roughly
https://github.com/desihub/desisim/blob/master/py/desisim/simexp.py#L473-L505

Also note that the correct link to the notebook from whence some of these numbers were derived is 
https://github.com/desihub/desitarget/blob/master/doc/nb/bgs-morphology-properties.ipynb
(i.e., part of master not the long-gone branch).

moustakas:
@julienguy with the Director's review over would you be willing to fix this bug so I can revisit the BGS redshift efficiency calculations?

The relevant bit of code is in cell [16] of [this notebook](https://github.com/desihub/surveysim/blob/twilight/doc/nb/BGS_ETC_Study.ipynb) by @dkirkby, although this uses `astropy.cosmology` and you mentioned (over email, off-list) that you wanted a standalone cosmology module in `desisim` (I'm ambivalent).

dkirkby:
It looks like the minimal fix required here is to replace these lines in simexp.py:
```
        source_half_light_radius[bgss,0]=4.7  # disk comp in BGS, arcsec
        source_half_light_radius[bgss,1]=1.3  # bulge comp in BGS, arcsec
```
with:
```
        LCDM = astropy.cosmology.Planck15
        angscale = (
            LCDM.angular_diameter_distance(0.1) /
            LCDM.angular_diameter_distance(source_redshift[bgss]).to(1).value
        source_half_light_radius[bgss,0]=4.7  * angscale
        source_half_light_radius[bgss,1]=1.3  * angscale
```
I made up `source_redshift` here, but presumably that array is available somewhere.

If someone wants to rewrite `astropy.cosmology.angular_diameter_distance`, then please include a unit test that compares against the astropy values.

akremin:
Completed with the Bgsfibaccept pull request merger in desisim

PR #326, opened 2018-02-22
Bgsfibaccept

akremin:
Addresses [desisim issue 289](https://github.com/desihub/desisim/issues/289)

Added realism to spectral sims by including varying fiber acceptance for BGS targets based on angular diameter distance (via redshift). Required the propagation of redshifts first to _quick_spectra_ and then to _simulate_spectra_ in _simexp.py_. Accomplished with optional argument in _quick_spectra_ that appends another column to the _frame_fibermap_ table that is passed to _simulate_spectra._ _None_ value is used as default to indicate no provided redshifts. _None_ forces all fiber acceptances to default to a fixed value (just as before).


sbailey:
Thanks for implementing this.  I would rather pass the `redshift` array directly into `simexp.simulate_spectra` as a new option (just like you added it as a new option to `quickspectra.sim_spectra`) instead of adding it as a new column to`frame_fibermap` just to pass it in so that it can be re-extracted.  i.e. let's keep things called "fibermap" matching what the real data fibermap will have, which won't include a REDSHIFT column.

py/desisim/simexp.py
```diff
@@ -480,9 +485,11 @@ def simulate_spectra(wave, flux, fibermap=None, obsconditions=None,
...
-            log.warning("the half light radii are fixed here (and not magnitude or redshift dependent)")
-
+        if np.sum(lrgs)>0 or np.sum(elgs)>0:
+            log.warning("the half light radii are fixed here for LRGs and ELGs (and not magnitude or redshift dependent)")
+        if np.sum(bgss)>0 and source_redshift is None:
+            log.warning("the half light radii are fixed here for BGS (as redshifts weren't supplied)")
+            
         # BGS parameters based on SDSS main sample, in g-band
         # see analysis from J. Moustakas in
         # https://github.com/desihub/desitarget/blob/bgs-properties/doc/nb/bgs-morphology-properties.ipynb
```
moustakas: Can you update this notebook link to https://github.com/desihub/desitarget/blob/master/doc/nb/bgs-morphology-properties.ipynb please?

py/desisim/simexp.py
```diff
@@ -505,18 +512,32 @@ def simulate_spectra(wave, flux, fibermap=None, obsconditions=None,
         source_fraction[elgs,0]=1.   # ELG are disk only
         source_fraction[lrgs,1]=1.   # LRG are bulge only
         source_fraction[bgss,0]=0.52 # disk comp in BGS
-        source_fraction[bgss,1]=0.48 # bulge comp in BGS
-        
+        source_fraction[bgss,1]=0.48 # bulge comp in BGS
```
moustakas: This is a general comment that shouldn't impact this review, but we should really be drawing these values from a distribution of morphological properties.

py/desisim/simexp.py
```diff
@@ -505,18 +512,32 @@ def simulate_spectra(wave, flux, fibermap=None, obsconditions=None,
...
-        source_half_light_radius[bgss,1]=1.3  # bulge comp in BGS, arcsec
-       
+
+        # 4.7 is angular size of z=0.1 disk, and 1.3 is angular size of z=0.1 bulge
+        bgs_disk_z01 = 4.7
+        bgs_bulge_z01 = 1.3
+        
+        # Input cosmology to calculate the angular diameter distance of the galaxy's redshift
+        from astropy.cosmology import Planck15 as LCDM
+        ang_diam_dist = LCDM.angular_diameter_distance
```
moustakas: Two comments: (1) I would put this import at the top of the `if` statement.  And (2) I don't feel strongly about this but I would use a much simpler concordance cosmology of H0=70, Omega0=0.3, and OmegaL=0.7, rather than the Planck+15 cosmology.  My $0.02.

py/desisim/simexp.py
```diff
@@ -505,18 +512,32 @@ def simulate_spectra(wave, flux, fibermap=None, obsconditions=None,
...
+        
+        # Convert to angular size of the objects in this sample with given redshifts
+        if source_redshift is None:
+            angscales = np.ones(np.sum(bgss))
+        else:
+            angscales = ( ang_diam_dist(0.1) / ang_diam_dist(source_redshift[bgss]) ).to(1).value
+
```
moustakas: Shouldn't this be ```( ang_diam_dist(source_redshift[bgss]) / ang_diam_dist(0.1) )```?  Also, can you independently confirm that this code gives you the behavior you expect, i.e., that the angular extent increases with redshift and then levels off?!

py/desisim/simexp.py
```diff
@@ -505,18 +512,32 @@ def simulate_spectra(wave, flux, fibermap=None, obsconditions=None,
...
+        
+        # Convert to angular size of the objects in this sample with given redshifts
+        if source_redshift is None:
+            angscales = np.ones(np.sum(bgss))
+        else:
+            angscales = ( ang_diam_dist(0.1) / ang_diam_dist(source_redshift[bgss]) ).to(1).value
+            
+        source_half_light_radius[bgss,0]= bgs_disk_z01 * angscales # disk comp in BGS, arcsec
+        source_half_light_radius[bgss,1]= bgs_bulge_z01 * angscales  # bulge comp in BGS, arcsec
+
```
moustakas: Presumably `bgs_disk_z01` and `bgs_bulge_z01` are expected to be in arcsec?

akremin:
Thank you for the responses. My thinking was to keep all the observing
conditions together in obsconds and target properties together in fibermap
(and to reduce the changes to function params), but I would actually prefer
the redshifts to be explicitly passed so I'm glad you both vote for it that
way.

I'll make those changes and do John's proposed sanity checks.

Best,
Anthony

akremin:
All comments addressed in the most recent commit.

py/desisim/simexp.py
```diff
@@ -505,18 +511,31 @@ def simulate_spectra(wave, flux, fibermap=None, obsconditions=None,
...
+            angscales = np.ones(np.sum(bgss))
+        else:
+            bgs_redshifts = redshift[bgss]
+            # Avoid infinities
+            if np.any(bgs_redshifts == 0.):
+                bgs_redshifts[bgs_redshifts==0.] = 0.0001
+            angscales = ( ang_diam_dist(0.1) / ang_diam_dist(bgs_redshifts) ).value
+        source_half_light_radius[bgss,0]= bgs_disk_z01 * angscales # disk comp in BGS, arcsec
+        source_half_light_radius[bgss,1]= bgs_bulge_z01 * angscales  # bulge comp in BGS, arcsec
+
```
moustakas: What am I misunderstanding?  If `bgs_disk_z01` is the half-light radius at z=0.1 shouldn't you multiply by `(ang_diam_dist(bgs_redshifts) /  ang_diam_dist(0.1) ).value` rather than `( ang_diam_dist(0.1) / ang_diam_dist(bgs_redshifts) ).value` as you have here?

py/desisim/simexp.py
```diff
@@ -505,18 +511,31 @@ def simulate_spectra(wave, flux, fibermap=None, obsconditions=None,
...
+        
+        # Convert to angular size of the objects in this sample with given redshifts
+        if redshift is None:
+            angscales = np.ones(np.sum(bgss))
+        else:
+            bgs_redshifts = redshift[bgss]
+            # Avoid infinities
+            if np.any(bgs_redshifts == 0.):
+                bgs_redshifts[bgs_redshifts==0.] = 0.0001
+            angscales = ( ang_diam_dist(0.1) / ang_diam_dist(bgs_redshifts) ).value
```
moustakas: I wonder if these BGS redshifts could ever be negative in the mocks, e.g., for nearby galaxies / dwarfs whose peculiar velocities produce blueshifts.  What if you made this conditional
`bgs_redshifts <= 0.0` ?

py/desisim/simexp.py
akremin: The relevant formula is:   angular_size = physical_size / ang_diam_dist
bgs_disk_z01 is the angular size at z=0.1

Therefore:
 physical_size = bgs_disk_z01 * dA(z=0.1)

And finally:
 angular_size_truez = physical_size / dA(z=truez)
                                =  bgs_disk_z01 * ( dA(z=0.1) / dA(z=truez) )

py/desisim/simexp.py
akremin: Good point, I'll make that switch right now while I'm still looking at it

py/desisim/simexp.py
akremin: Sorry, I don't use markdown so the above looks terrible. The confusion is that bgs_disk_z01 is an _angularsize_ not the actual extent of the object physically. Otherwise you would be correct

PR #329, opened 2018-03-01
implement scatter gather, fix broken cosmics for mpi

lastephey:


lastephey:
This branch was the most successful of several ways to implement MPI communication in pixsim. Scatter/gather was both the fastest and most human-readable method. 

The other methods we tried:

MPI scatterv and gatherv, git hash: 0.23.0-40-g6fa4224
MPI send/receive, git hash: 0.23.0-43-g980f3be

sbailey:
Looks good, thanks.  I tried to compare timing to desisim/0.26.0, but that crashes with a segfault (!).  Using this branch pixsim succeeds, though it isn't clear to me what change fixed that or whether we're just lucky.  Merging now.

sbailey:
Post-merge update: @lastephey clarified to me that the previous `pixsim_mpi` code had been incorrectly calling`comm_group.bcast(cosmics, root=0)` only for rank 0.  Moving that to correctly call that for all ranks was the fix for the segfault.  It is still unclear to me why MPI would segfault in that case instead of just hanging waiting for the other ranks, but at least I feel better that this isn't just being lucky with the fix working now.

PR #331, opened 2018-03-03
fix psf convolution for newexp_mock

sbailey:
Previous PR #320 fixed the PSF double convolution for newexp_random but not newexp_mock (Mea culpa!).  This PR fixes it for newexp_mock too.  I plan to do a few more tests before being ready to merge but I'm opening this PR now to get the travis tests running.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -2,10 +2,12 @@
 desisim change log
 ==================
 
-0.26.0 (unreleased)
+0.26.1 (unreleased)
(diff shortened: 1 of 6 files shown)


sbailey:
I verified that this PR fixes the double PSF convolution with `newexp_mock`, while not messing up the `fastframe` output which was already convolved once and should remain so.  I'm ready to merge once Travis finishes.
pr331-skyphot

PR #334, opened 2018-03-09
Astropy 2 compatibility

### LINKED ISSUE #309: Update desisim to astropy 2.x
Update desisim to be astropy 2.x compatible.  Maintain astropy 1.3.x compatibility if possible.

weaverba137:
This PR fixes #309.  I am still working out the bug related to `Table()` units.  However, I have determined the exact cause of this bug, and it is **NOT** due to `Table()` somehow requiring that all columns have units.  Please see astropy/astropy#7279 for the exact cause, but in brief, some tables produced by desisim have columns with `dex` as the unit.  There is no problem with a `Table()` having a column with this unit.  The problem is converting that unit *into a FITS-compliant unit*.

I will consider some simple workarounds, like not using 'dex' as a unit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,12 @@ os:
 
 # Travis tech support recommended trying sudo true for more stable builds.
 # Seems odd, but maybe they isolate us from other users more if we need sudo.
-sudo: true
+# sudo: true
+# Setting sudo to false opts in to Travis-CI container-based builds.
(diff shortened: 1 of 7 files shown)


weaverba137:
See also #335.

weaverba137:
Well, that was easy...  Ready for review.

PR #337, opened 2018-03-16
fix pixsim for new pix file output location

sbailey:
desihub/desispec#523 moved the output location of preprocessed pix files, which broke pixsim tests because they didn't pre-create the necessary output directory.  This PR fixes that.

A better long-term solution might be to add a `mkdir` option to `desispec.io.findfile` similar to what is already in `desisim.io.findfile`.  Maybe.  For now this PR just makes the minimal change in desisim to restore its tests when running against desispec master.

### DIFF
--- a/py/desisim/scripts/pixsim.py
+++ b/py/desisim/scripts/pixsim.py
@@ -525,6 +525,7 @@ def main(args, comm=None):
                     camera = args.cameras[c]
                     pixfile = desispec.io.findfile('pix', night=args.night,
                         expid=args.expid, camera=camera)
+                    os.makedirs(os.path.dirname(pixfile), exist_ok=True)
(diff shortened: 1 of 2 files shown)


tskisner:
This all looks good- sorry for the unintended breakage....

sbailey:
I hit some intermittent Travis pixsim test problems on this for py2.7, including one case that ran to almost completion, printing
```
Ran 60 tests in 599.998s

OK (skipped=2, expected failures=1)
```
and then hanging for 10 minutes before Travis killed off the job.  When it reran successfully it took only half the time so perhaps there was some problem on the Travis end, e.g. memory contention with another job on the same node.  I'll merge now, but I'm adding these notes in case this crops up again.

PR #344, opened 2018-03-28
One month pixsim

lastephey:
This pull request fixes a previous bug in which the cameras were not properly divided between nodes in pixsim. This pull request also significantly cuts down on memory overhead in read_simspec_mpi, allowing the processes to stay under the cori haswell node memory limit.


sbailey:
Looks good.  I tested on an exposure with 30 frames on 30 nodes in <5 min.  Fixed a bug when running without cosmics.  Will merge when Travis tests pass.

PR #345, opened 2018-04-02
Healpix in spectra header

julienguy:
Solving issue https://github.com/desihub/desispec/issues/567
Writes HPXNSIDE HPXPIXEL HPXNEST keywords in spectra files generated with quickquasars .

### DIFF
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -186,10 +186,11 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
 
     log.info("Simulate DESI observation and write output file")
     pixdir = os.path.dirname(ofilename)
-    if not os.path.isdir(pixdir) :
-        log.info("Creating dir {}".format(pixdir))
-        os.makedirs(pixdir)
-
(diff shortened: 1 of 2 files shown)


py/desisim/scripts/quickquasars.py
```diff
@@ -200,7 +201,10 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
         log.warning("No TARGETID")
         targetid=None
 
-    sim_spectra(qso_wave,qso_flux, args.program, obsconditions=obsconditions,spectra_filename=ofilename,seed=args.seed,sourcetype="qso", skyerr=args.skyerr,ra=metadata["RA"],dec=metadata["DEC"],targetid=targetid)
+    log.warning("Assuming the healpix scheme is 'NESTED'")
```
sbailey: I don't think we need this warning.  By definition DESI uses the nested scheme everywhere; it isn't an option that requires an assumption about which one to use.

ngbusca:
By the way, I've asked Thomas/JM to add those keywords to their absorption files. At some point we might want to read them directly from the input files.

Issue #346, opened 2018-04-16
QA: algorithmic access to desi_qa_zfind metrics


sbailey:
The plots output by `desi_qa_zfind` include metrics like efficiency, catastrophic failure rate, dv bias, dv RMS, but I don't see a way to algorithmically obtain those numbers.

* Provide a function to calculate those numbers given a redshift catalog and a truth table, without making plots.
* When running `desi_qa_zfind`, cache those numbers in a json (or equivalent) file in the qa directory along with the plots (e.g. for reading by other code to compare two productions and highlight changes).

profxj:
Regarding algorithmically,  it is rather straightforward to generate
the stats from a list of fibermap_files.  This generates a Table that
merges the measured redshifts and the truth.  

If you wish to input a catalog and truth table, we'll need to be strict
on formatting, etc.

sbailey:
Example inputs in `/project/projectdirs/desi/datachallenge/reference_runs/18.3`:
* redshift catalog: spectro/redux/mini/zcatalog-mini.fits
* superset truth table: targets/truth.fits

Note: `zcatalog-mini.fits` is a concatenation of individual zbest files (is that what you meant rather than fibermap files?) which was then joined with the input `targets/targets.fits` file (not the truth file, but the target catalog file like real data will have).  This is a standard pipeline output file.

I was thinking that you would define the minimal subset of columns that you need to be able to do the calculations using the same names that are in those files, e.g. something like
* redshift catalog must have columns `TARGETID`, `Z`, `ZWARN`, `SPECTYPE` (any others?)
* truth catalog must have columns `TARGETID`, `TRUEZ`, `TRUESPECTYPE` (any others?)

Or optionally provide a pre-joined single table that has all of those columns.

e.g. I'd be happy if the end-user code looked something like:
```
zcat = fitsio.read('spectro/redux/mini/zcatalog-mini.fits')
truth = fitsio.read('targets/truth.fits')
zqa_stats = desispec.qa.calc_zqa_stats(zcat, truth)
print(zqa_stats['QSO']['dv_rms'])
print(zqa_stats['QSO']['dv_bias'])
print(zqa_stats['QSO']['zeff'])
#- etc.
```

profxj:
ok, I nearly have this.  Only column missing is DESI_TARGET.
I can add that, but inform me how to determine which of the 
STAR and WD TEMPLATETYPE's are MWS objects.  It must
be only a subset.

profxj:
[...]

PR #349, opened 2018-04-17
Redshift QA metrics from summary tables

profxj:
Calculate the summary QA statistics for redshift performance
provided an input truth and measured redshift tables.  Tested
to work with truth.fits and zcatalog-mini.fits from reference run 18.3

Here is the code snippet

```
from desisim.spec_qa import redshifts as dsqa_z
from astropy.table import Table   # The underlying code uses MaskedColumn and other astropy functionality

zcat = Table.read('spectro/redux/mini/zcatalog-mini.fits')
truth = Table.read('targets/truth.fits')

# Sync the two tables
dsqa_z.match_truth_z(truth, zcat, mini_read=True)

# Calculate
metric_dict = dsqa_z.summ_stats(truth)
```

Addresses Issue #346 

Have also tested that --yaml_file= option in desi_qa_zfind works
with reference run 18.3.  Will add it to the minitest Notebook soon.

### DIFF
--- a/py/desisim/spec_qa/redshifts.py
+++ b/py/desisim/spec_qa/redshifts.py
@@ -253,14 +253,11 @@ def load_z(fibermap_files, zbest_files=None, outfil=None):
 
 def match_truth_z(simz_tab, zb_tab, mini_read=False, outfil=None):
     """ Match truth and zbest tables
-    :param simz_tab: astropy.Table
-      Either generated from load_z() or read from disk via 'truth.fits'
-    :param zb_tab: astropy.Table
-      Either generated from load_z() or read from disk via 'zcatalog-mini.fits'
-    :param mini_read: bool, optional
-      Tables were read from the summary tables written to disk
+    :param simz_tab: astropy.Table; Either generated from load_z() or read from disk via 'truth.fits'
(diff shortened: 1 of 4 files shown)


py/desisim/spec_qa/redshifts.py
```diff
@@ -116,39 +117,82 @@ def calc_stats(simz_tab, objtype, flux_lim=True):
...
+    import pdb; pdb.set_trace()
+    for brick in brick_names:
+        zbest_path=desispec.io.findfile('zbest', groupname=brick, specprod_dir=args.reduxdir)
+        if os.path.exists(zbest_path):
+            log.debug('Found {}'.format(os.path.basename(zbest_path)))
+            zbest_files.append(zbest_path)
+        else:
+            log.warn('Missing {}'.format(os.path.basename(zbest_path)))
+            #pdb.set_trace()
+'''
```
sbailey: Minor: Let's drop this "commented with quotes" code that searches for zbest files with the old brickname organization that we don't use anymore.

profxj: removed.

Have added a legend to the summary dict that
should appear towards the top of the YAML file.

I fear the hooks to astropy Table's are deeper than
simply Masked Column.  Would be a rather major
refactor to move to numpy arrays alone.

PR #350, opened 2018-04-18
Quickquasars random seed

julienguy:
Fix issues with random seeds in quickquasars


julienguy:
merging. travis test failure has nothing to do with quickquasars.

PR #351, opened 2018-04-18
Add ztests method to redshift QA

profxj:
Integrates the ztests method from desitest mini-Notebook within
the existing redshift QA.  Tested against the summary tables from
18.3 with code like this:

```
from desisim.spec_qa import redshifts as dsqa_z

truth = Table.read('truth.fits')
zcat = Table.read('zcatalog-mini.fits')

dsqa_z.match_truth_z(truth, zcat, mini_read=True)
good, fail, miss, lost = dsqa_z.zstats(truth, objtype='STAR')
```

Also refactored scripts to use qaprod_dir 
and cleaned up several of the summary 
redshift stats.

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -5,7 +5,11 @@ desisim change log
 0.27.1 (unreleased)
 -------------------
 
-* Enable redshift QA using input summary catalogs of truth and redshifts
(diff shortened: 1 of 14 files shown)


profxj:
Addresses Issue #351

PR #352, opened 2018-04-18
Quickquasars mags

julienguy:
Optionally compute magnitudes in quickquasars and save them in hdu SCORES of the spectra files.


julienguy:
After discussion with Stephen, broad band fluxes should be written to the fibermap not the scores. I will change this in this branch.

PR #358, opened 2018-05-02
working pixsim_nights

lastephey:
This pull request makes use of the newly implemented simulate_exposure function which handles all the MPI I/O for pixsim. Pixsim_nights enables the user to specify the nights they would like to simulate and optionally specific cameras and cosmics for each exposure. Several utility functions allow the exposures and frames to be divided among the nodes in a sensible way. The job will fail with a helpful error message if the nodes cannot be filled appropriately.


sbailey:
I tested that this works with single and multiple exposures and various packings of nodes per exposure.  OK to merge once unit tests are fixed.  Thanks.

PR #366, opened 2018-05-24
Add dla

alxogm:
modified to read DLA's info from transmission file, construct the dla model and uptadate transmission file after downsampling but before target selection

### DIFF
--- a/py/desisim/lya_spectra.py
+++ b/py/desisim/lya_spectra.py
@@ -11,7 +11,7 @@ import numpy as np
 from desisim.dla import insert_dlas
 from desiutil.log import get_logger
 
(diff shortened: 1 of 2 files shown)


andreufont:
Thanks @alxogm! 

Yes, we want to store metadata about the DLAs that were added. I believe this was already done for the DLAs added randomly, this could be a good starting point. 

It would also be nice to have the option to add random DLAs to the London/Saclay mocks.

alxogm:
I added the option to also add random DLAs, and a temporary line to set the transmission to 1 for z>zqso... adding now the  the lines to store the DLAs metadata and a few lines about the BALs, this is in qqso_dla_bal branch ... will coordinate with @forero about this

PR #368, opened 2018-06-06
add transmission of metals from LYA transmission

londumas:
Add transmission from metals given the transmission from Lya.
`--metals SiII(1260) SiIII(1207) SiII(1193) SiII(1190) `?
To get the parameter for SiII(1260) SiIII(1207) SiII(1193) SiII(1190) I compared to the data.
Do we want to get the parameter for all the other lines CIV..., do we want to make a deeper study of the values of the parameters?
Such a study could take more time.

Here is a comparison of the 1d-correlation for: data, "noiseless+Transmission+Lya" and "noiseless+Transmission+Lya+Metals"
x1d

The unique small issue is that if setting something like SiII(1260) (wavelength>1215.67), because of the wavelength difference between Lya and this line, a part of the transmission at small wavelength is exactly 1.
@ngbusca did you notice that when doing the mockExpander? What solution did you use.

Here is an example of the metal transmission, you can see that it is exactly 1 for small wavelength.
spec_siii1260

### DIFF
--- a/py/desisim/lya_spectra.py
+++ b/py/desisim/lya_spectra.py
@@ -163,16 +163,11 @@ def apply_metals_transmission(qso_wave,qso_flux,trans_wave,trans,metals) :
 
     output_flux = qso_flux.copy()
     for q in range(qso_flux.shape[0]) :
-
-        metTrans = np.ones(trans_wave.size)
-
         for m in metals:
             mtau = absorber_IGM[m]['COEF']*tau[q,:]
             mtrans_wave = (zPix[q,:]+1.)*absorber_IGM[m]['LRF']
-            mtau = np.interp(trans_wave,mtrans_wave,mtau,left=0.,right=0.)
-            metTrans *= np.exp(-mtau)
+            output_flux[q, :] *= np.interp(qso_wave,mtrans_wave,np.exp(-mtau),left=1.,right=1.)
 
-        output_flux[q, :] *= np.interp(qso_wave,trans_wave,metTrans,left=0,right=1)
     return output_flux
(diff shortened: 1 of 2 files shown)


londumas:
@julienguy and @andreufont, I just found the way to have quickquasar running. Don't bother.

julienguy:
I am testing this. Thanks.

londumas:
@julienguy, could you expand the london/v1.1/ mocks for ~50,000 quasars? Or tell me how to do it with your `quickquasar_slurm_script.sh`? If it is good, we will test the cross and the auto.
I would like to see how it looks like with noise+continuum.

julienguy:
Two very little requests on the code.
- please add some logging messages log.INFO("adding metals X,Y,Z")
- dump list of possible metals in error message when wrong metal transition is given

londumas:
@julienguy, are the changes ok for you?

julienguy:
yes

andreufont:
@londumas - That was fast! Looks great. 

A couple of comments:
 - should we have all metal-related functions in a different file? Even if now they are only a couple of lines, it will potentially grow to more. 
 - you could also add Lyman beta absorption using exactly the same trick, it would be cool :-)
 - regarding the problem with F=1 in some range: in the long term, we will add these using the density field, what will solve this issue. However, meanwhile we have the option to do a hack: we can write the transmission files also at z < 2 and at z > z_qso, and make sure that quickquasars also uses the transmission at z > z_qso to add metals, and not to add Lya. It is very hacky, but it is probably either this or weird metals for now...

PR #373, opened 2018-06-07
add parameter for metal lines

londumas:
Add parameter for other lines, including Lyb.
The values have been tuned roughly to the data.
More work would be needed to get something very precise, but it is enough for a lot of
tests of the metals.


julienguy:
waiting for travis tests before merging ...

londumas:
@julienguy, I did some tests on transmission files it works very well:
The following plot gives you for example the 1d correlation between the Lyb forest and the Lya forests.

capture d ecran de 2018-06-07 15-57-07

londumas:
@julienguy, Maybe I should add the option to do all metals, for example if `args.metals = all`.
What do you think?

londumas:
@julienguy, I just add the fact that if the user set `--metals all`, then all the metals are used.

julienguy:
ok (and I tested it). waiting for travis tests before merging ...

londumas:
@julienguy, travis failed for reasons that don't seem related to my PR...

julienguy:
There is apparently a real bug in desisim.test.test_quickcat.TestQuickCat that has nothing to do with this PR. I am trying to fix this now...

londumas:
@julienguy, thanks

PR #374, opened 2018-06-08
Qqsuasar dlas

alxogm:
@julienguy New try to merge with master. Includes dla and bal features

### DIFF
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -299,7 +299,7 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
     if (args.balprob):
        if(args.balprob<=1. and args.balprob >0):
           log.info("Adding BALs with probability {}".format(args.balprob))
-          tmp_qso_flux,meta_bal=bal.insert_bals(tmp_qso_wave,tmp_qso_flux, metadata['Z'], balprob=args.balprob)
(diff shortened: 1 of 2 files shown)


py/desisim/scripts/quickquasars.py
```diff
@@ -245,7 +283,12 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
         nmodel=nqso, redshift=metadata['Z'], 
         lyaforest=False, nocolorcuts=True, noresample=True, seed = seed)
 
+    ##To add BALs to be checked by Luz and Jaime
+    if args.balprob:
+       bal=BAL()
+       tmp_qso_flux,meta_bal=bal.insert_bals(tmp_qso_wave,tmp_qso_flux, metadata['Z'], balprob=args.balprob)
```
moustakas: You should pass `seed=seed` to `bal.insert_bals()` on line 289 to ensure reproducibility (i.e., that the same QSOs become BAL QSOs when the seed is the same).

Also consider passing the `BAL()` object as an argument to `simulate_one_healpix` so you don't read the BAL basis templates over and over.

alxogm:
@moustakas I added the seed, but passing the BAL() as argument, maybe I should leave it for @forero who is going to take over the BALs simulations in general, also note I'm not saving the BAL metadata yet.

alxogm:
@julienguy I've checked that not dlas are being added if --dla is not present... in line 154 I only read the skewers from transmission file the there are no dlas or they are random...

alxogm:
@julienguy  I just fixed to add BALs only when --balprob is present, and since I was checking this piece the  I included the suggestion from @moustakas  to pass BAL() as argument to simulate_one_pixel... 

Could you check if there are more requests?

Issue #324, opened 2018-02-14
desisim.templates.SIMQSO.make_templates() has no option `restframe`


londumas:
`desisim.templates.SIMQSO.make_templates()` has no option `restframe`.
This option is useful to build PCA because without it leads to empty data.

londumas:
same for desisim.templates.QSO.make_templates()``

moustakas:
The only reason this is a non-trivial change is that `wave` would have to be a `flux`-shaped vector, one for each template.  (This situation is slightly different than for the other target classes because the QSO templates are all rendered in the observed frame by construction.)

Would this be OK @londumas?

londumas:
@moustakas, There is no rush to correct this. I can use the current to build PCAs.
I have no strong opinions on weather it is ok to have wave be a flux shaped vector or not.

londumas:
@moustakas, the main reason to get the option `restframe` is to get the full spectrum in order to use it to build PCAs and achetypes. Would that be possible?

moustakas:
Addressed in #377.

PR #377, opened 2018-06-25
add restframe option to SIMQSO

moustakas:
Simple PR to address request #324.  @londumas I haven't tested this extensively, so please make sure Lyman-alpha and other lines are where they should be!

Note that for issues of speed (i.e., not having to resample) and because of the way `desisim.templates` uniquely interfaces with `simqso` (i.e., uniquely, compared to the other templates which we build in-house) you have to specify the `restframe` option in the class constructor itself, e.g.,

```
from desisim.templates import SIMQSO
qso = SIMQSO(restframe=True)
flux, wave, meta = qso.make_templates(5, seed=1)
```

Finally note that the output (rest-frame) wavelength vector is hard-coded to 900-60000 A at a resolution of 8000, but this can be changed easily.

### DIFF
--- a/py/desisim/templates.py
+++ b/py/desisim/templates.py
@@ -2007,7 +2007,8 @@ class SIMQSO():
     """Generate Monte Carlo spectra of quasars (QSOs) using simqso."""
 
     def __init__(self, minwave=3600.0, maxwave=10000.0, cdelt=0.2, wave=None,
-                 nproc=1, normfilter='decam2014-r', colorcuts_function=None):
+                 nproc=1, normfilter='decam2014-r', colorcuts_function=None,
+                 restframe=False):
         """Read the QSO basis continuum templates, filter profiles and initialize the
            output wavelength array.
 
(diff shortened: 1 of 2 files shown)


py/desisim/templates.py
```diff
@@ -2231,15 +2239,17 @@ def _make_simqso_templates(self, redshift=None, rmagrange=None, seed=None,
                 meta['REDSHIFT'][these] = input_qsometa.z
             else:
                 meta['SEED'][these] = seed
-                meta['REDSHIFT'][these] = redshift[these]
-                
+                if self.restframe:
+                    meta['REDSHIFT'][these] = 0
```
londumas: even if the output is in restframe a qso at z=3 and at z=1 does not look the same, mainly because of the Lyman alpha forest. Maybe we should keep the redshift information.

moustakas: So you're thinking that you would specify an input redshift distribution?  I can implement this, but if you're building archetypes I worry that -- given how strongly the position of the Lyman break / the forest varies with redshift -- you'll end up with many more archetypes that you'd want.

What if you generated a generic set of Lya forest spectra -- over a range of redshifts (or even on-the-fly?) -- separately from the QSO continuum spectra?

londumas: ok, no sorry. I though it was the output redshift. Not the input one.

Issue #357, opened 2018-05-01
chi2 and amplitude computation is slow and chi2 is wrong


londumas:
https://github.com/desihub/desisim/blob/36b78ee41ad5d0f7e2bb4bc4b2058e761eeb3260/py/desisim/archetypes.py#L52

This line is very slow, a simple fix is using something like what follows.
We need to see how to add the error contribution.
This uses Bolton et al. 2012 (https://arxiv.org/pdf/1207.7326.pdf), see section 4.3.2.
This gives the same result for the amplitude but a factor 2 difference for the chi2.
It seems that my chi2 is good and that the one of `mathutils.quick_amplitude` is not.
```
def my_compute_chi2(flux, ferr=None):  

    nspec, npix = flux.shape

    chi2 = np.zeros((nspec, nspec)).astype('f4')
    amp = np.zeros((nspec, nspec)).astype('f4')

    rescale = np.sqrt(npix/np.sum(flux**2,axis=1))
    flux *= rescale[:,None]
    if ferr is None:
        ferr = np.ones(flux.shape)
    else:
        ferr *= rescale[:,None]

    for ii in range(nspec):
        if ii % 500 == 0 or ii == 0:
            log.info('Computing chi2 matrix for spectra {}-{} out of {}.'.format(
            (ii % 500) * 500, np.min(((ii+1) * (ii % 500), nspec-1)), nspec))
        amp1 = np.sum(flux[ii]*flux,axis=1)/npix
        chi21 = npix*(1.-amp1**2)
        chi21[np.isclose(amp1,1.)] = 0.
        chi2[ii, :] = chi21
        amp[ii, :] = amp1

    return chi2, amp
```

Here is a simple code to compare both results
```
import scipy as sp
import matplotlib.pyplot as plt

from desisim.archetypes import ArcheTypes, compute_chi2

def my_compute_chi2(flux):  

    nspec, npix = flux.shape
    ferr = sp.ones_like(flux)

    chi2 = sp.zeros((nspec, nspec)).astype('f4')
    amp = sp.zeros((nspec, nspec)).astype('f4')

    for ii in range(nspec):
        if ii % 500 == 0 or ii == 0:
            print('Computing chi2 matrix for spectra {}-{} out of {}.'.format((ii % 500) * 500, min(((ii+1) * (ii % 500), nspec-1)), nspec))
        amp1 = sp.sum(flux[ii]*flux,axis=1)/npix
        chi21 = npix*(1.-amp1**2)
[...]

PR #382, opened 2018-06-27
correct computation of chi2 and amplitude between archetypes

londumas:
This PR fix issue https://github.com/desihub/desisim/issues/357 by computing the chi2 between normalized archetypes. Thanks to that the chi2 matrix is symmetric. The amplitude matrix is however given for the input spectra, not for the normalized one. If no error vector is given, this PR is way faster than the previous code.

### DIFF
--- a/py/desisim/archetypes.py
+++ b/py/desisim/archetypes.py
@@ -53,6 +53,7 @@ def compute_chi2(flux, ferr=None):
             amp[ii,:] = amp1
     else:
         from SetCoverPy import mathutils
+        ferr = ferr.copy()
         ferr *= rescale[:,None]
         for ii in range(nspec):
             if ii % 500 == 0 or ii == 0:
--- a/py/desisim/archetypes.py
+++ b/py/desisim/archetypes.py
@@ -29,29 +29,43 @@ def compute_chi2(flux, ferr=None):
     -------
     Tuple of (chi2, amp) where:
         chi2 : numpy.ndarray
-            Chi^2 matrix [Nspec, Nspec] between all combinations of spectra.
+            Chi^2 matrix [Nspec, Nspec] between all combinations of normalized spectra.
         amp : numpy.ndarray
             Amplitude matrix [Nspec, Nspec] between all combinations of spectra.
 
     """
-    from SetCoverPy import mathutils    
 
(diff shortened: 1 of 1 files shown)


py/desisim/archetypes.py
```diff
@@ -29,29 +29,42 @@ def compute_chi2(flux, ferr=None):
...
+        from SetCoverPy import mathutils
+        ferr *= rescale[:,None]
+        for ii in range(nspec):
+            if ii % 500 == 0 or ii == 0:
+                log.info('Computing chi2 matrix for spectra {}-{} out of {}.'.format(
+                    (ii % 500) * 500, np.min(((ii+1) * (ii % 500), nspec-1)), nspec))
+            xx = flux[ii, :].reshape(1, npix)
+            xxerr = ferr[ii, :].reshape(1, npix)
+            amp[ii,:], chi2[ii,:] = mathutils.quick_amplitude(xx, flux, xxerr, ferr, niter=1)
+
```
moustakas: In #357 you found that the chi^2 value returned by `SetCoverPy.mathutils.quick_amplitude` was wrong, so why are we using it at all?  

In case it's useful, here's another calculation of chi2 that takes into account the inverse variance, but I haven't tested how it compares to your algorithm or to `quick_amplitude`.

```
ivar = 1 / ferr**2
amplitude = np.sum(ivar[:, :, np.newaxis] * flux[:, :, np.newaxis] * \
  flux[:, np.newaxis, :], axis=0) ) / np.sum(ivar[:, :, np.newaxis] * \
  flux[:, np.newaxis, :]**2, axis=0)

chi2 = np.sum(ivar[:, :, np.newaxis] * ( flux[:, :, np.newaxis] - \
  amplitude[np.newaxis, :, :] * flux[:, np.newaxis, :])**2, axis=0)
```

moustakas:
And it looks like one Travis test failed due to the randomly occurring "LRG spectrum with too-low flux"; restarting the test should be good enough.

py/desisim/archetypes.py
londumas: Actually it is correct once the spectra are normalized. So both piece of code give the same chi2 matrix.

PR #389, opened 2018-07-11
fix newexp-mock and wrap-fastframe NIGHT/EXPID/ vs. NIGHT/ parsing

sbailey:
This PR fixes a few leftovers from the switch to raw data in `basedir/NIGHT/EXPID/` directories instead of all exposures in a single `basedir/NIGHT/` directory.  In particular:
  * `wrap-fastframe` looks for input files with the correct glob (one level deeper)
  * `wrap-fastframe` parses the night from the path correctly
  * `newexp-mock` passes `expid` to `desisim.io.simdir` to get correct output location
  * `desisim.io.simdir` no longer has a default `expid=0` which was previously leading to science exposures silently being written to the incorrect directory.

@weaverba137 these changes are required for the minitest notebook to work.  I have tested them by taking the minitest notebook up to these stages and failing, and then debugging the `newexp-mock` and `wrap-fastframe` commands separately using this branch.

### DIFF
--- a/bin/wrap-fastframe
+++ b/bin/wrap-fastframe
@@ -54,7 +54,7 @@ else:
 
 #- Assemble full list of simspec files; group by flavor to help MPI balancing
 if rank == 0:
-    allfiles = sorted(glob.glob('{}/*/simspec*.fits'.format(desisim.io.simdir())))
+    allfiles = sorted(glob.glob('{}/*/*/simspec*.fits'.format(desisim.io.simdir())))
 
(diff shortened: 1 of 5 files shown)


weaverba137:
Cori is down today.  I should be able to do a full run through the notebook Thursday morning.

sbailey:
After chatting with @weaverba137, merging this into master so that we can update the master copy at NERSC and continue with the mini-test notebook integration testing.

PR #390, opened 2018-07-16
faster emission line sims for MKL>=2018.0.2

sbailey:
This PR fixes desihub/desitarget#323 where `select_mock_targets` became mysteriously slower for BGS and ELG using the latest desiconda.  I traced this to a single line in `desisim.templates.EMSpectrum`:
```python
emspec += amp * np.exp(-0.5 * (self.log10wave-thislinewave)**2 / log10sigma**2) \
            / (np.sqrt(2.0 * np.pi) * log10sigma)
```
The Gaussian is evaluated at all wavelengths `self.log10wave` instead of just a few sigma around `thislinewave`.  For earlier desiconda installations using the Intel Math Kernel Library (MKL) 2018.0.1, this wasn't a problem.  But newer desiconda installations use MKL 2018.0.2, which is 10x slower for this particular case of evaluating exponentials of large arrays of large negative numbers that will mostly result in 0.

I fixed this by pre-filtering the wavelengths to +/- 5 sigma around the line to add.  This restores the performance, and even makes it a bit faster than before.

### DIFF
--- a/py/desisim/templates.py
+++ b/py/desisim/templates.py
@@ -222,14 +222,16 @@ class EMSpectrum(object):
         # Finally build the emission-line spectrum
         log10sigma = linesigma/LIGHT/np.log(10) # line-width [log-10 Angstrom]
         emspec = np.zeros(len(self.log10wave))
+
         for ii in range(len(line)):
             amp = line['flux'][ii] / line['wave'][ii] / np.log(10) # line-amplitude [erg/s/cm2/A]
             thislinewave = np.log10(line['wave'][ii] * (1.0+zshift))
             line['amp'][ii] = amp / (np.sqrt(2.0 * np.pi) * log10sigma)  # [erg/s/A]
 
             # Construct the spectrum [erg/s/cm2/A, rest]
-            emspec += amp * np.exp(-0.5 * (self.log10wave-thislinewave)**2 / log10sigma**2) \
-                      / (np.sqrt(2.0 * np.pi) * log10sigma)
+            jj = np.abs(self.log10wave-thislinewave) < 5*log10sigma
+            emspec[jj] += amp * np.exp(-0.5 * (self.log10wave[jj]-thislinewave)**2 / log10sigma**2) \
+                          / (np.sqrt(2.0 * np.pi) * log10sigma)
 
         return emspec, 10**self.log10wave, line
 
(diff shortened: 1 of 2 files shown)


moustakas:
Yowsers!

py/desisim/templates.py
```diff
@@ -222,14 +222,16 @@ def spectrum(self, oiiihbeta=None, oiihbeta=None, niihbeta=None,
...
             amp = line['flux'][ii] / line['wave'][ii] / np.log(10) # line-amplitude [erg/s/cm2/A]
             thislinewave = np.log10(line['wave'][ii] * (1.0+zshift))
             line['amp'][ii] = amp / (np.sqrt(2.0 * np.pi) * log10sigma)  # [erg/s/A]
 
             # Construct the spectrum [erg/s/cm2/A, rest]
-            emspec += amp * np.exp(-0.5 * (self.log10wave-thislinewave)**2 / log10sigma**2) \
-                      / (np.sqrt(2.0 * np.pi) * log10sigma)
+            jj = np.abs(self.log10wave-thislinewave) < 5*log10sigma
+            emspec[jj] += amp * np.exp(-0.5 * (self.log10wave[jj]-thislinewave)**2 / log10sigma**2) \
+                          / (np.sqrt(2.0 * np.pi) * log10sigma)
```
moustakas: I can't check right now if this would ever happen in the code, but you may want to make sure `jj` isn't sometimes false.  Also, can you make allowances for my idiosyncrasies and make it 6-sigma, please?

sbailey: I thought I was already being excessive in going to 5 sigma instead of 4 sigma, but ok, I changed it to 6.

Note that `jj` is an array of booleans, not a single boolean.  It is ok even if they are all `False`.  numpy will do the right thing and just not update any wavelengths in that case (i.e. not crash).  Does that address your concern, or am I missing something about your "make sure `jj` isn't sometimes false" suggestion?

```python
x = np.arange(10.0)
y = x**2

#- Some True, some False
ii = x<5
y[ii] += np.exp(-x[ii])

#- All False
jj  = x > 100
y[jj] += np.exp(-x[jj])
```

moustakas: Yes, of course.  All good, thanks.

Issue #386, opened 2018-07-04
quickquasars simulated spectra can not be reproduced


andreufont:
I can not reproduce the simulated spectra, even when passing a random seed. For instance, when I run the simple code below.

``
#!/bin/bash

downsampling=0.40
outdir=/project/projectdirs/desi/mocks/lya_forest/london/v2.0/test_quick/v0
infile=/project/projectdirs/desi/mocks/lya_forest/london/v2.0/0/0/transmission-16-0.fits

quickquasars --exptime 4000. -i $infile --zmin 1.8 --outdir $outdir/spectra-16 --downsampling 
$downsampling --zbest --mags --desi-footprint --seed 123
``

Moreover, we need to make sure that we are able to run two simulations that are identical in any way, except adding a new artefact. This means that we should be able to add DLAs or not, without this changing the random numers for the quasar continua or for the instrumental noise.

I belive we can easily accomplish this by resetting the random number generator, using the seed that has been passed, every time we need to generate numbers at a different part of the code. We don't care if the random numbers in continuum and in noise are correlated, as lonf as they are different in different quasars. Right?

andreufont:
@alxogm @julienguy @londumas @forero

alxogm:
From what Ive seen quasars are reproduced if no features present. No BAL quasar are reproduced when run with BAL probability different from zero, i.e. only BALs look different. 
Differences  still happening with DLAs, dig in into it...

alxogm:
So I've seen that spectra are reproduced if no noise is added. in line https://github.com/desihub/desisim/blob/9df332d970b2e32ffb26f4c7a4203df745896bb9/py/desisim/scripts/quickspectra.py#L197

As well as if I increase the exposure time,   the differences gets much smaller as can be seen in the current version of the notebook  https://github.com/desihub/desisim/blob/fix_dlabugs/etc/quick_quasar.ipynb

Also checked that the status of the random generator status is the same for all runs  up to line 
https://github.com/desihub/desisim/blob/9df332d970b2e32ffb26f4c7a4203df745896bb9/py/desisim/scripts/quickspectra.py#L181

However funny things happen  in
https://github.com/desihub/desisim/blob/9df332d970b2e32ffb26f4c7a4203df745896bb9/py/desisim/scripts/quickspectra.py#L182

like if I pass a seed number explicitly here, the difference between the dla random/no_dla was smaller, but not for any number I pass :S 

[...]

PR #393, opened 2018-07-20
No poisson

julienguy:
numpy.random.poisson has the drawback of using a varying number of random numbers depending on the mean value (the argument of the function). this is problematic for some applications like testing the effect of DLA on the Lya correlation function where one would like to preserve the random numbers used to simulate the instrumental noise when adding DLAs or not. This PR fixes this by allowing the use of numpy.random.normal instead of numpy.random.poisson in specsim.simulator.Simulator.generate_random_noise.

Solves issue #386.
(Need to rerun tests when the equivalent PR in specsim is merged.)

### DIFF
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -208,9 +208,9 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
 ##ALMA:added to set transmission to 1 for z>zqso, this can be removed when transmission is corrected.        
     for ii in range(len(metadata)):       
         transmission[ii][trans_wave>1215.67*(metadata[ii]['Z']+1)]=1.0
-              
+
     if(args.dla=='file'):
-        log.info('Adding DLAs from transmision file')
+        log.info('Adding DLAs from transmission file')
         min_trans_wave=np.min(trans_wave/1215.67 - 1)
         for ii in range(len(metadata)):
             if min_trans_wave < metadata[ii]['Z']: 
@@ -231,17 +231,19 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
 
     elif(args.dla=='random'):
         log.info('Adding DLAs randomly')
+        saved_rnd_state = np.random.get_state()
         min_trans_wave=np.min(trans_wave/1215.67 - 1)
         for ii in range(len(metadata)):
             if min_trans_wave < metadata[ii]['Z']: 
-               idd=metadata['MOCKID'][ii]
-               dlass, dla_model = insert_dlas(trans_wave, metadata[ii]['Z'])
-               if len(dlass)>0:
-                  transmission[ii]=dla_model*transmission[ii]
-                  dla_z+=[idla['z'] for idla in dlass]
-                  dla_NHI+=[idla['N'] for idla in dlass]
-                  dla_id+=[idd]*len(dlass)    
+                idd=metadata['MOCKID'][ii]
+                dlass, dla_model = insert_dlas(trans_wave, metadata[ii]['Z'], seed=np.random.randint(2**32)) # need to pass a random seed otherwise insert_dla breaks the random sequence
+                if len(dlass)>0:
+                    transmission[ii]=dla_model*transmission[ii]
+                    dla_z+=[idla['z'] for idla in dlass]
+                    dla_NHI+=[idla['N'] for idla in dlass]
+                    dla_id+=[idd]*len(dlass)
         log.info('Added {} DLAs'.format(len(dla_id)))
+        np.random.set_state(saved_rnd_state) # restore random state to get the same random numbers later as when we don't insert DLAs
 
     if args.dla is not None :
        dla_meta=Table()
@@ -297,7 +299,9 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
     if (args.balprob):
        if(args.balprob<=1. and args.balprob >0):
           log.info("Adding BALs with probability {}".format(args.balprob))
+          rnd_state = np.random.get_state() # save current random state
           tmp_qso_flux,meta_bal=bal.insert_bals(tmp_qso_wave,tmp_qso_flux, metadata['Z'], balprob=args.balprob,seed=seed)
+          np.random.set_state(rnd_state) # restore random state to get the same random numbers later as when we don't insert BALs
        else:
           log.error("Probability to add BALs is not between 0 and 1")
           sys.exit(1)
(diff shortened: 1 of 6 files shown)


py/desisim/scripts/quickspectra.py
```diff
@@ -50,7 +50,8 @@ def sim_spectra(wave, flux, program, spectra_filename, obsconditions=None,
         meta : dictionnary, saved in primary fits header of the spectra file 
         fibermap_columns : add these columns to the fibermap
         fullsim : if True, write full simulation data in extra file per camera
-    
+        use_poisson : if False, do not use numpy.random.poisson to simulate the Poisson noise. 
+                      This is useful to get reproducible random realizations.
```
dkirkby: -> "This is useful to decouple the (normalized) noise contributions from the predicted mean signal."

py/desisim/scripts/quickspectra.py
```diff
@@ -178,10 +179,10 @@ def sim_spectra(wave, flux, program, spectra_filename, obsconditions=None,
     sim = desisim.simexp.simulate_spectra(wave, flux, fibermap=frame_fibermap,
         obsconditions=obsconditions, redshift=redshift, seed=seed,
         psfconvolve=True)
-
+
```
dkirkby: invisible white space (https://www.python.org/dev/peps/pep-0008/?#id28)

py/desisim/scripts/quickquasars.py
```diff
@@ -231,17 +231,19 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
 
     elif(args.dla=='random'):
         log.info('Adding DLAs randomly')
+        saved_rnd_state = np.random.get_state()
```
dkirkby: Its much easier to manage different parts of a program consuming random numbers independently if they each have their own private `np.random.RandomState`.  A corollary is that you should avoid using the global RandomState with direct calls to `np.random....`

py/desisim/scripts/quickquasars.py
```diff
@@ -231,17 +231,19 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
...
             if min_trans_wave < metadata[ii]['Z']: 
-               idd=metadata['MOCKID'][ii]
-               dlass, dla_model = insert_dlas(trans_wave, metadata[ii]['Z'])
-               if len(dlass)>0:
-                  transmission[ii]=dla_model*transmission[ii]
-                  dla_z+=[idla['z'] for idla in dlass]
-                  dla_NHI+=[idla['N'] for idla in dlass]
-                  dla_id+=[idd]*len(dlass)    
+                idd=metadata['MOCKID'][ii]
+                dlass, dla_model = insert_dlas(trans_wave, metadata[ii]['Z'], seed=np.random.randint(2**32)) # need to pass a random seed otherwise insert_dla breaks the random sequence
```
dkirkby: Pass in a `dla_random_state` here that was created during initialization (instead of a seed).

py/desisim/scripts/quickquasars.py
```diff
@@ -231,17 +231,19 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
...
+                if len(dlass)>0:
+                    transmission[ii]=dla_model*transmission[ii]
+                    dla_z+=[idla['z'] for idla in dlass]
+                    dla_NHI+=[idla['N'] for idla in dlass]
+                    dla_id+=[idd]*len(dlass)
         log.info('Added {} DLAs'.format(len(dla_id)))
+        np.random.set_state(saved_rnd_state) # restore random state to get the same random numbers later as when we don't insert DLAs
```
dkirkby: This is not necessary if the DLA code has its own RandomState object.

py/desisim/scripts/quickquasars.py
```diff
@@ -297,7 +299,9 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
     if (args.balprob):
        if(args.balprob<=1. and args.balprob >0):
           log.info("Adding BALs with probability {}".format(args.balprob))
+          rnd_state = np.random.get_state() # save current random state
```
dkirkby: (Similar comments here)

py/desisim/scripts/quickspectra.py
julienguy: seriously?

py/desisim/scripts/quickquasars.py
julienguy: will do

py/desisim/scripts/quickquasars.py
julienguy: yes, now that I understand where the change of noise is coming from.

dkirkby:
I am restarting the travis tests after merging desihub/specsim#92.  If those pass, this looks ready to merge.

dkirkby:
It looks like travis is still stuck on a docstring formatting error:
```
Warning, treated as error:
/home/travis/build/desihub/desisim/py/desisim/scripts/quickspectra.py:
  docstring of desisim.scripts.quickspectra.sim_spectra:25:Unexpected indentation.
```

dkirkby:
Taking a closer look at this package's travis config, tests are hardcoded to use specsim v0.11 so are not using the update to specsim master.  I will tag a new version of specsim soon (with desimodel updates) so I suggest we wait for that before updating the travis config here, but this PR could be merged now if its blocking someone.

andreufont:
@dkirkby - It is not critical, but it would be great to have this solved soonish, since it will allow several tests in the Lyman alpha working group.

dkirkby:
I just tagged specsim v0.12 and updated this package's travis tests to use this new version. If they pass this is ready to merge & tag, then both tags can be included in the 18.7 release.

andreufont:
That's great! Fingers crossed :-)

julienguy:
desisim docstring issue ... I am having a look at this

julienguy:
a curious test failure that should never have succeeded: in quickspectra, when the seed is changed the flux in spectra  is expected to change but not the ivar. however the test is requiring a different ivar (??). I am changing this.

julienguy:
quickgen test failing (nothing to do with quickspectra or quickquasars ...)

julienguy:
- test don't pass because quickgen doesn't give the same result when called twice with the same random seed.
- specsim.instrument.initialize(config) has a np.random.RandomState() if config.instrument.offset has no attribute 'seed'. this is what breaks the random number sequence. Setting a seed value in a config file wouldn't make sense, so one would need to alter the config object before it is used in the constructor of instrument ... I am not sure where one should do that and I don't think we should have a random seed in a configuration object anyways.
- a short term fix would be load the config, modify it, then create the simulator object in desisim.specsim.get_simulator(with a new argument seed) ... but that looks like a band-aid solution.

sbailey:
I spun the specsim part to desihub/specsim#94 .  This quickgen test was using a seed for generating the instrument offsets random state, but since it was using the same Simulator object multiple times in a row, its results depended upon this history of what else was simulated first.  i.e. simulating the same spectra twice in a row using the same simulator doesn't give the same answer.

For now I've "fixed" this in the desisim tests by spawning quickgen using `subprocess.call` instead of `quickgen.main()` so that I won't re-use the same `specsim.Simulator` on the second call.

sbailey:
Arg.  My `subprocess.call` fix works when desisim/bin is in $PATH, but not for the bare-bones testing environment pre-installation and without setting $PATH.  Working on it...

PR #395, opened 2018-07-26
clobber -> overwrite

sbailey:
This PR converts `astropy.io.fits` argument `clobber` to `overwrite` to avoid deprecation warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,9 +45,9 @@ env:
         # to repeat them for all configurations.
         # - NUMPY_VERSION=1.10
         # - SCIPY_VERSION=0.16
-        - ASTROPY_VERSION=2.0.4
+        - ASTROPY_VERSION=2.0.7
         # - SPHINX_VERSION=1.6.6
-        - DESIUTIL_VERSION=1.9.9
+        - DESIUTIL_VERSION=1.9.11
         - SPECLITE_VERSION=0.7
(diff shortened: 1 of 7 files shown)


sbailey:
In general we've been keeping the travis dependencies as the oldest known versions that work, and let the nightly cronjob tests at NERSC test the latest version of each package against each other.  It's fine to update the travis dependencies when we know that we're making a change that requires a newer version, but as long as the tests pass I consider it somewhat of a feature when travis is checking old versions given that NERSC cronjobs are checking new versions.  Note that the recent minitest notebook integration tests problems wouldn't have been caught by the current unit tests at all, with old or new dependencies.

Travis had a few hiccups when running tests yesterday; restarting those jobs now and I'll fix any remaining tests if things fail again.

weaverba137:
@sbailey, in general I agree, but weren't there some significant changes in desimodel in 18.6?  Also, it looks like desisim is testing against the full desimodel download, instead of the testing branch.

sbailey:
Good point.  I'll update desimodel tests to use the latest testing branch (that also explains why desisim tests have become so slow again on Travis...)

sbailey:
Updated to desimodel test-0.9.6 and confirmed that the test logs no longer have the clobber deprecation warning.  @weaverba137 any other blocking issues or ok to merge?

PR #397, opened 2018-07-31
Update to use fiberassign script in place of fiberassign_exec from c++ code

rncahn:
This resurrects quicksurvey to enable simulations of survey with multiple epochs.  It also removes references to the template (features file).  The primary change is in the subprocess call, which now uses the fiberassign script rather than going directly to the c++ executable fiberassign_exec.


sbailey:
The changes here look fine, but Travis tests are failing for other reasons.  I'm going to rebase against current master and then re-evaluate.  For the moment please don't make any updates to this branch.

PR #400, opened 2018-08-07
update templates to DR7 standard stars

moustakas:
Update templates code to accommodate the changes to `desitarget` in https://github.com/desihub/desitarget/pull/338, largely related to changing the standard-star designation from `FSTD` to `STD`.


sbailey:
Looks good; thanks.  Travis failures appear due to internet hiccups on the Travis side; rerunning now.

sbailey:
Note that we might have another name change coming soon to split best vs. regular standards, but merging this now to restore current test integration.

PR #401, opened 2018-08-08
SIMQSO and QSO classes in quickquasar

forero:
I've added the option to use either `desisim.templates.SIMQSO` or `desisim.templates.QSO` in `quickquasars`.

Below some comparison plots for the same rows in the `spectra_*.fits` file:

image

image

image

The data was produced using:

* SIMQSO templates
```
quickquasars -i $DESI_MOCKS/lya_forest/london/v2.0/17/1712/transmission-16-1712.fits --zmin 1.6 --nproc 10 --downsampling 1. --seed 42 --zbest --outdir ./SIMQSO --simqso --overwrite --exptime 1E6
```

* QSO templates
```
quickquasars -i $DESI_MOCKS/lya_forest/london/v2.0/17/1712/transmission-16-1712.fits --zmin 1.6 --nproc 10 --downsampling 1. --seed 42 --zbest --outdir ./QSO --overwrite --exptime 1E6
```

Any tests that I should be running on these data?

### DIFF
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -14,6 +14,7 @@ from desispec.io.util import write_bintable
 from desispec.io.fibermap import read_fibermap
 from desisim.simexp import reference_conditions
 from desisim.templates import SIMQSO
+from desisim.templates import QSO
 from desisim.scripts.quickspectra import sim_spectra
 from desisim.lya_spectra import read_lya_skewers , apply_lya_transmission, apply_metals_transmission
 from desisim.dla import dla_spec,insert_dlas
@@ -63,7 +64,8 @@ def parse(options=None):
     #- Optional arguments to include dla
 
     parser.add_argument('--dla',type=str,required=False, help="Add DLA to simulated spectra either randonmly (--dla random) or from transmision file (--dla file)")
-    parser.add_argument('--balprob',type=float,required=False, help="To add BAL features with the specified probability (e.g --balprob 0.5). Expect a number between 0 and 1 ")    
+    parser.add_argument('--balprob',type=float,required=False, help="To add BAL features with the specified probability (e.g --balprob 0.5). Expect a number between 0 and 1 ") 
+    parser.add_argument('--simqso',action = "store_true", help="Uses desisim.templates.SIMQSO to generate templates. desisim.templates.QSO is used otherwise.")
     if options is None:
         args = parser.parse_args()
     else:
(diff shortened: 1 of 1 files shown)


andreufont:
Hi @forero - I couldn't reproduce your plots because of bitmask-related errors (given Stephen email I probably need to re-synchronize my desihub packages), but I went through the code and I only have minor comments: 
 - I fixed the code so that it can run in a single processor (was giving an error)
 - I renamed the option --no-simqso, since it would be great to keep SIMQSO as default (as it was now).
 - SIMQSO.make_templates has an option noresample that is currently being used. I'm not sure how relevant this is, but I added a comment to remind us that we do not have this option in QSO at the moment.

andreufont:
I think the code crashes because this branch does not include the recent changes on master related to bitmask. I guess we'd need to rebase, but I'm always terrified of doing this, so I'll leave that to you :-)

andreufont:
Ok, I managed to get it running after just doing "git merge master" from this branch, and after looking at a few examples I noticed that QSO quasar seem to be on average brighter. 

I think the reason is that while SIMQSO draws random magnitudes following a QLF, QSO draws random magnitudes using a flat prior on a magnitude range. It would be great if we added an option for QSO to also draw from a QLF, but this should probably go in another issue. 

Happy to merge when the checks pass.

andreufont:
Should I push my branch after doing "git merge master"? This will add new 17 commits, that are already on master, so I'm not sure...

moustakas:
Yes, when you branch gets behind master doing 
```bash
git checkout master
git pull
git checkout BRANCH
git merge --no-ff master
git push
```
is the typical flow.

andreufont:
Done. Thanks!

About noresample, it would probably be good to add the option, so that both QSO / SIMQSO modules behave as similar as possible in anything else apart from the actual template generation.

forero:
Thanks for all the additional work and comments!

@andreufont please push your changes to the branch `optional_simqso` so that others (including myself) can test the code; @moustakas can you please add what is needed for the `noresample` option?

andreufont:
I already did!

moustakas:
> About noresample, it would probably be good to add the option, so that both QSO / SIMQSO modules behave as similar as possible in anything else apart from the actual template generation.

OK, I added a `noresample` argument to `templates.QSO` but please note that the returned (observed-frame) wavelength vector is now an `nmodel` by `npixel` length array (rather than a common, one-dimensional `npixel`-length array), which will likely require some handling in the downstream code.  

Additional details:  Because SIMQSO generates templates using a model, we can evaluate the model with any desired wavelength vector (in the case of `templates.SIMQSO` this wavelength vector is controlled by the `basewave_min`, `basewave_max`, and `basewave_R` arguments).  

`templates.QSO`, on the other hand, is tied to the (rest-frame) `eigenwave` wavelength vector which was generated as part of the PCA analysis.  We then compute the (outer) dot product with the redshift of each QSO to get the (individual) observed-frame wavelength vectors.  When `noresample=False` each wavelength vector is resampled to a default (DESI-like) wavelength range and resolution (by default, 3600-10000 A with 0.2 A pixels).  When `noresample=True` we simply return the full wavelength matrix.

I hope this behavior is what is desired but if not let's open a new issue.

andreufont:
Great!  Happy to merge, except for Travis tests failing. Probably not related to this?

PR #402, opened 2018-08-09
update to STD_FAINT

sbailey:
Updates for compatibility with desitarget/0.23.0 which renamed bit STD -> STD_FAINT.

When it wouldn't clutter the code too much, this also maintains backwards compatibility with previous names (STD, STD_FSTAR).  quickspectra and quickcat are not backwards compatible but simexp is.

Companion PR to desihub/desitarget#341 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ env:
         - SPECSIM_VERSION=v0.12
         # - DESISPEC_VERSION=0.18.0
         - DESISPEC_VERSION=master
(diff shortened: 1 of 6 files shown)


geordie666:
Looks good, at least for `STD` -> `STD_FAINT`. We'll need to refactor if we ultimately additionally implement `STD_FAINT_BEST` and `STD_BRIGHT_BEST` but let's cross that bridge if we come to it.

PR #406, opened 2018-08-13
different seeds for each jobs

londumas:
Ask for one seed per file, for example:
```
quickquasars --exptime 4000.
-i /project/projectdirs/desi/mocks/lya_forest/london/v2.0//9/931/transmission-16-931.fits
/project/projectdirs/desi/mocks/lya_forest/london/v2.0//9/932/transmission-16-932.fits
--zmin 1.8 --nproc 4
--outdir /project/projectdirs/desi/mocks/lya_forest/london/v2.0/quick-1.4//spectra-16
--downsampling 0.40 --zbest --mags --desi-footprint
--seed 937728865 159455235
--metals all --nmax 1
```

### DIFF
--- a/etc/quickquasar_slurm_script.sh
+++ b/etc/quickquasar_slurm_script.sh
@@ -45,16 +45,6 @@ nfilespernode=$((nfiles/nodes+1))
 echo "n files =" $nfiles
 echo "n files per node =" $nfilespernode
 
-# Random seeds
-get_seeded_random()
-{
-    thisseed="$1";
-    openssl enc -aes-256-ctr -pass pass:"$thisseed" -nosalt \
-    </dev/zero 2>/dev/null;
-}
-randoms=($(shuf --random-source=<(get_seeded_random $seed) -i 0-999999999 -n $nfiles))
-
-
 first=1
 last=$nfilespernode
 for node in `seq $nodes` ; do
@@ -62,17 +52,14 @@ for node in `seq $nodes` ; do
 
     # list of files to run
     if (( $node == $nodes )) ; then
-        seeds=(${randoms[@]:$first-1})
         last=""
-    else
-        seeds=(${randoms[@]:$first-1:$nfilespernode})
     fi
 
     echo ${first}-${last}
     tfiles=`echo $files | cut -d " " -f ${first}-${last}`
     first=$(( first + nfilespernode ))
     last=$(( last + nfilespernode ))
-    command="srun -N 1 -n 1 -c $nthreads  quickquasars --exptime 4000. -i $tfiles --zmin 1.8 --nproc $nthreads --outdir $outdir/spectra-16 --downsampling $downsampling --zbest --mags --desi-footprint --seed ${seeds[*]} --metals all"
+    command="srun -N 1 -n 1 -c $nthreads  quickquasars --exptime 4000. -i $tfiles --zmin 1.8 --nproc $nthreads --outdir $outdir/spectra-16 --downsampling $downsampling --zbest --mags --desi-footprint --seed seed"
     echo $command
     echo "log in $outdir/logs/node-$node.log"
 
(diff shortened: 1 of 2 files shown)


andreufont:
I would prefer the code suggested by Stephen, since it is (in my opinion) much simpler and hides the issue inside the python script. For instance, when I run quickquasars on our local UCL cluster I need to use QSUB instead of SLURM, and I wouldn't know how to reproduce the magic there.

londumas:
@andreufont, as you prefer. However, there are no magic in the code. It is simply a bash script that shuffles numbers between 0 and 999999999 and select a fraction to use as seeds.

andreufont:
@londumas - sorry, I didn't mean to imply that your code was bad. I'm just never been very good at bash / slurm / qsub scripts, and it is difficult for me to follow what's going on. 

But yes, if you don't mind, I'd prefer to implement the python code suggested by Stephen. I can take care of it.

londumas:
@andreufont, don't worry I was not taking your remark that way. Please feel free to update this PR.

etc/quickquasar_slurm_script.sh
```diff
@@ -45,34 +45,21 @@ nfilespernode=$((nfiles/nodes+1))
...
-    else
-        seeds=(${randoms[@]:$first-1:$nfilespernode})
     fi
 
     echo ${first}-${last}
     tfiles=`echo $files | cut -d " " -f ${first}-${last}`
     first=$(( first + nfilespernode ))
     last=$(( last + nfilespernode ))
-    command="srun -N 1 -n 1 -c $nthreads  quickquasars --exptime 4000. -i $tfiles --zmin 1.8 --nproc $nthreads --outdir $outdir/spectra-16 --downsampling $downsampling --zbest --mags --desi-footprint --seed ${seeds[*]} --metals all"
+    command="srun -N 1 -n 1 -c $nthreads  quickquasars --exptime 4000. -i $tfiles --zmin 1.8 --nproc $nthreads --outdir $outdir/spectra-16 --downsampling $downsampling --zbest --mags --desi-footprint --seed seed"
```
londumas: $seed

py/desisim/scripts/quickquasars.py
```diff
@@ -73,14 +74,20 @@ def parse(options=None):
...
     return args
 
-def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filters,footprint_healpix_weight,footprint_healpix_nside,seed,bal=None) :
+
+def get_pixel_seed(pixel, nside, global_seed):
+    if global_seed is None:
+        # return a random seed
+        return np.random.randint(2**32, size=1)[0]
+    np.random.seed(global_seed)
+    seeds = np.random.randint(2**32, size=healpy.nside2npix(nside))
```
londumas: You can have repeated seeds.
Better do:
`seeds = sp.unique(np.random.randint(2**32, size=10*healpy.nside2npix(nside)))[:healpy.nside2npix(nside)]`

andreufont:
@londumas - I pushed some changes implementing the python function. I tested on a small scale and it seems to work, let me know what you think. Thanks!

py/desisim/scripts/quickquasars.py
andreufont: Why would you have repeat seeds? I'm confused.

etc/quickquasar_slurm_script.sh
andreufont: I only tested quickquasars, not the slurm script (I don't use it on my local cluster)

py/desisim/scripts/quickquasars.py
londumas: `np.random.randint` can give you twice the same seed. The probability is low because `2**32` is a big number, but see the examples at the bottom: https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.randint.html

andreufont: I'm on my way home now, will implement these later

londumas:
@andreufont, thanks for the changes.

andreufont:
@londumas - I think it is ready now. @julienguy - the code has changed considerably, you might want to review it? I've checked that it runs on small batches.

julienguy:
OK for me. I haven't run the code, I trust Helion on this. I agree it's better this way. Thank you all for this, I'm merging.

Issue #411, opened 2018-09-10
fix quickquasars; write out truth metadata


moustakas:
#409 likely broke `quickquasars`.

One immediate issue is that the "normalization filter" argument needs to be explicitly passed as either `normfilter_south` or `normfilter_north` for quasars observed in the DECaLS vs BASS/MzLS filters.

When making these changes we should also write out the "truth" metadata tables as an additional output file as we do in `desitarget.build` (via `desitarget/bin/select_mock_targets`).

andreufont:
@moustakas - it would be great if quickquasars in the master branch was always working, since this is an important code used very often in the Lyman alpha working group. 

I guess this means we should setup a test in Travis for it? 

@alxogm - Tell your students that quickquasars is broken, but it looks like the fix should be easy.

alxogm:
Hi @moustakas, I made the test of not passing any argument, as I saw normfilter_north/south has defaults. But  the default  normfilter_north='BASS-r' sends the error 

 No such group 'BASS'. Choose one of ['sdss2010', 'decam2014', 'wise2010', 'hsc2017', 'lsst2016', 'bessell'].

(in templates.py:SIMQSO() but I think is the same for QSO() and others)

What should I choose?

moustakas:
You need the master branch of speclite until @dkirkby creates a new tag.  

Please see the thread here:
https://github.com/desihub/redrock/issues/150

moustakas:
And just for the record: I'm fixing what's broken with `quickquasars` (independent of the issue with `speclite`) in the `fix-quickquasars` branch and should have that ready for review soon.

alxogm:
OK. Now it works. Thanks! @moustakas 

The quickquasars flag '--norm-filter' was set as default for "decam2014-g", that corresponds to normfilter_south.   Shall I add both north and south to the quickquasar options?  or just leave it as it is, and pass normfilter_south=args.norm-filter to SIMQSO( ) and leave north as the default?

moustakas:
I've made these changes in my branch already.  I'm working on passing forward and writing out the metadata table, so you can add to that.

alxogm:
@moustakas ok! thanks.  So I'll wait for your changes.

moustakas:
Done in #413.

Issue #404, opened 2018-08-13
QSO.make_templates should be able to draw magnitudes from Luminosity Function


andreufont:
When generating correlated Lyman alpha mock spectra, we read quasar redshifts and angular positions from the "transmission files", but we do not read magnitudes. 

When using SIMQSO.make_templates we can pass a magnitude range and a quasar redshift, and we will get a random magnitude following a given QSO LF. It would be great to have the same feature in QSO.make_template (either from the same QSO LF or from another one).

Ideally, we would like to have two different behaviors: 
 - when generating mocks that will go through target selection, we would use a realistic luminosity function.
 - when generating mocks that will not go through target selection, we would use the expected distribution of magnitudes observed by DESI.

This double functionality should also be implemented in SIMQSO.

moustakas:
Sounds good, any contributions would be more than welcome.

Note that this issue is related to #106.

andreufont:
True, I didn't realize that there was a similar ticket open. Hopefully we can close both soon :-)

moustakas:
Some additional notes:

In `templates.SIMQSO` we sample from a specified luminosity function here-- 
https://github.com/imcgreer/simqso/blob/master/simqso/sqgrids.py#L1223-L1264

The default luminosity function in `SIMQSO` is from [BOSS/DR9](https://arxiv.org/abs/1210.6389), as written down here (although there are other choices)--
https://github.com/imcgreer/simqso/blob/master/simqso/sqmodels.py#L222-L235

Finally, @dkirkby developed some code for sampling from the [Palanque et al. 2012](https://arxiv.org/abs/1209.3968), [Palanque et al. 2015](https://arxiv.org/abs/1509.05607) luminosity functions--
  https://github.com/dkirkby/ArgonneLymanAlpha/blob/master/notebooks/SampleProperties.ipynb

I can't take the lead on this but hopefully someone from the Lya WG can implement the requisite changes.

andreufont:
@alxogm - could you think of someone that might be able to implement these? If we wanted to use templates.QSO this would be important, although I guess we could decide to continue using SIMQSO.

andreufont:
@moustakas @alxogm , fixing this should be a priority for the Lyman alpha working group.

Ideally, one would decouple the question of drawing random magnitudes from the question of generating quasar templates. So I would like to have two functions, outside of either SIMQSO or QSO:

[...]

PR #413, opened 2018-09-12
restore quickquasars to functioning state after templates refactor

moustakas:
This PR addresses #411 by restoring `quickquasars` to a functioning state and:
* Self-consistently handles "north" (BASS/MzLS) vs "south" (DECaLS) photometry and target selection;
* Properly handles the output of `templates.SIMQSO` and `templates.QSO` when `noresample=True`, thereby removing one interpolation step.

@andreufont @alxogm @julienguy please test and let me know if you find any problems.

Note that I did not change the code to write out any additional metadata---or the *truth* spectra---to the *zbest.fits* file.  I leave this up to you to implement so you can build the data model you need.  However, in case it's helpful, [here](https://github.com/desihub/desitarget/blob/master/py/desitarget/mock/build.py#L875-L897) is where we write out the *truth.fits* files in `desitarget/bin/select_mock_targets`.

Finally, I agree that some unit tests for `quickquasars` would be valuable, but I have not implemented any here.

### DIFF
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -355,11 +355,11 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
     # photometric systems.
     south = np.where( is_south(metadata['DEC']) )[0]
     north = np.where( ~is_south(metadata['DEC']) )[0]
-    meta, qsometa = empty_metatable(nqso, objtype='QSO', simqso=~args.no_simqso)
+    meta, qsometa = empty_metatable(nqso, objtype='QSO', simqso=not args.no_simqso)
     if args.no_simqso:
         log.info("Simulate {} QSOs with QSO templates".format(nqso))
-        tmp_qso_flux = np.zeros([nobj, len(model.eigenwave)], dtype='f4')
-        tmp_qso_wave = np.zeros_like(qso_flux)
+        tmp_qso_flux = np.zeros([nqso, len(model.eigenwave)], dtype='f4')
+        tmp_qso_wave = np.zeros_like(tmp_qso_flux)
     else:
         log.info("Simulate {} QSOs with SIMQSO templates".format(nqso))
         tmp_qso_flux = np.zeros([nqso, len(model.basewave)], dtype='f4')
@@ -368,7 +368,7 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
     for these, issouth in zip( (north, south), (False, True) ):
         if len(these) > 0:
             _tmp_qso_flux, _tmp_qso_wave, _meta, _qsometa = model.make_templates(
-                nmodel=nqso, redshift=metadata['Z'], lyaforest=False, nocolorcuts=True,
+                nmodel=len(these), redshift=metadata['Z'][these], lyaforest=False, nocolorcuts=True,
                 noresample=True, seed=seed, south=issouth)
             meta[these] = _meta
             qsometa[these] = _qsometa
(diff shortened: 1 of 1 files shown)


andreufont:
I said that the code run on my desktop, but the luminosity of the quasars seems to have changed significantly... they are all much brighter now. To be honest, this is something that we have not tested much in the past, so it is possible that the code is doing the right thing now and the wrong in the past, but I would like someone to cross-check the differences in normalization. Some examples (note that I generated these with very low noise to better comparison).

[test-quick-0.pdf](https://github.com/desihub/desisim/files/2375686/test-quick-0.pdf)
[test-quick-1.pdf](https://github.com/desihub/desisim/files/2375687/test-quick-1.pdf)
[test-quick-2.pdf](https://github.com/desihub/desisim/files/2375688/test-quick-2.pdf)
[test-quick-3.pdf](https://github.com/desihub/desisim/files/2375689/test-quick-3.pdf)

moustakas:
Nice find, @andreufont.  Can you send me the input arguments to make these plots?

andreufont:
Hi @moustakas - I use something like this (extracted from desisim/etc/quickquasar_slurm_script.sh):

#!/bin/bash

# change that to a folder containing London mock skewers
idir=/data/london-skewers/v2.0/
outdir=/data/london-skewers/v2.0/test-quick/

if [ ! -d $outdir ] ; then
    mkdir -p $outdir
fi    
if [ ! -d $outdir/spectra-16 ] ; then
    mkdir -p $outdir/spectra-16
fi

file="$idir/0/0/transmission-16-0.fits $idir/0/2/transmission-16-2.fits"

quickquasars --exptime 1000000. -i $file --zmin 1.8 --outdir $outdir/spectra-16 --zbest --mags --desi-footprint --nmax 10 --seed 123 --overwrite

moustakas:
@andreufont I don't think I'll have time to really investigate these differences in luminosity / scaling. I *did* change the normalization filters in my previous PR (to accommodate the north/south split in photometric systems), so maybe this is related.  I suggest you (or @alxogm) investigates this by generating a bunch of spectra with, e.g.,
```
flux, wave, meta, objmeta = templates.SIMQSO(100, nocolorcuts=True)
```
 and looking at `objmeta['MABS_1450']` to see if the distribution (and the correlation with redshift and apparent magnitude) is sensible (unfortunately the previous version of `SIMQSO` didn't return this quantity -- sorry).  

You might also make sure you're simulating spectra below 32 deg declination, so that `decam2014-r` gets used as the normalization filter.   (Or should we be using the g-band???)

Finally, it would be great if one of you could add a very basic unit test for `quickquasars` before merging, although this would require adding a very small Lyman-alpha skewer/transmission file and `master.fits` file to the repository (in, e.g., `test/data`).

I leave it up to you whether to merge now and implement these issues separately, or to keep working on this branch.  Thanks (and sorry I can't do more right now!)

andreufont:
@moustakas - well, the current version in the master branch is broken, so anything would be better than that...

But this is an important issue, we should really check that the generated luminosity function is consistent with the one used to draw random luminosities and with the effect of target selection. 

I would suggest to merge the branch as is, and to put as a priority for the working group to measure the luminosity function of the simulated quasars, with and without applying target selection cuts, and compare it to the input model used in SIMQSO. @alxogm, do you think you could find a volunteer to do this?

alxogm:
@andreufont @moustakas I can check with Rodrigo to make this tests and report what is being get. I'll make my pending pull request and can take a look too. I agree we should merge this one to have a working version. thanks

alxogm:
@moustakas @andreufont Rodrigo and Gustavo are looking into this. Can you point us to a sensible distribution of 'MABS_1450'?

moustakas:
Hopefully it matches the BOSS/DR9 QSOLF!  I added some notes here recently---
https://github.com/desihub/desisim/issues/404#issuecomment-421385114

alxogm:
> Finally, it would be great if one of you could add a very basic unit test for `quickquasars` before merging, although this would require adding a very small Lyman-alpha skewer/transmission file and `master.fits` file to the repository (in, e.g., `test/data`).
> 

I am about to make a PR for writing the zbest_file (with error added) and the truth file from quickquasars. I can try to add this unit test. If I understand right this involves writing a script like https://github.com/desihub/desisim/blob/master/py/desisim/test/test_quickspectra.py
?

julienguy:
Exactly. That would be great.

PR #409, opened 2018-08-28
update template/simspec metadata data model

moustakas:
This PR addresses #290 with a major (and backwards-incompatible) change to how the template/simulated metadata are returned by the various `desisim.templates` classes, and written in the `simspec` files.   

This is still a work in progress (~one more day) but I wanted @sbailey to take an early look since he's simultaneously updating the fibermap data model, which will touch several of the same routines.  This change will also affect the Lya WG via changes to `quickquasars` (see below).

Some new features:
* The `TRUTH` metadata table is significantly simpler in that it only contains objtype-agnostic information, e.g.,
```
['TARGETID', 'OBJTYPE', 'SUBTYPE', 'TEMPLATEID', 'SEED', 'REDSHIFT', 'MAG', 'MAGFILTER',
 'FLUX_G', 'FLUX_R', 'FLUX_Z', 'FLUX_W1', 'FLUX_W2']
```
* We now write out objtype-specific metadata, depending on the mix of object types (linked via their unique `TARGETID`), e.g., for STDs in the `TRUTH_STD` extension the columns are simply:
```
['TARGETID', 'TEFF', 'LOGG', 'FEH']
```
* I have added a `MAGFILTER` column indicating the spectrum normalization filter (corresponding to `MAG`) (previously the filter was not discoverable from the data themselves), which is also needed to deal with the north/south split in our photometry.
* A metadata table for BALs in correctly constructed and populated in `templates.QSO` and returned.
* The `SIMQSO` metadata (not a simple astropy Table) are properly written out and read back in for reproducibility (however, [this bug](https://github.com/imcgreer/simqso/issues/26) will need to be fixed first.

Still to do:
- [ ] Add arguments to synthesize photometry and apply target selection cuts separately for DECaLS vs MzLS/BASS (i.e., south vs north) photometry.
- [ ] Update `scripts.quickspectra.sim_spectra` and the routines that call it (e.g., `quickquasars`).
- [ ] Restore support for (re)generating templates using the `input_meta` argument.
- [ ] Update code related to `desitarget/bin/select_mock_targets`, which will be done in a separate PR.

Finally, note that the simulated simspec files are comparable in size to previously generated files:  there are now additional HDUs and a replicated `TARGETID` but the change in file size is not significant.  In addition, `desisim.io.read_simspec` can still read in old-style simspec files without error.

PR #416, opened 2018-09-13
read and write select_mock_targets style simspec file

moustakas:
#409 missed a corner case whereby `desisim.io.read_simspec` could read the output of `desisim.obs.new_exposure` but *not* the output of `desisim/bin/newexp-mock` (via the output of `desitarget.build.write_targets_truth`).

@sbailey, the committed changes will allow you to proceed; however, they do not include the enhancement of writing out the object-specific metadata (e.g., `TRUTH_ELG`) to the `simspec` files using `newexp-mock`, which I think they should.


moustakas:
@sbailey This is not quite ready to be merged--I've identified at least one other issue on the `desitarget` side which will break what you're working on.  I'll submit a PR shortly.

PR #435, opened 2018-10-08
Fix qquasar

alxogm:
@andreufont @sbailey  
I added a try statement when trying to read the 'Z_noRSD' from the transmission file. And included a test to check quickquasar runs with the Saclay mocks. 

Some travis tests are failing but I think they are not due to this PR. Please let me know if any change is needed.


moustakas:
Yeah, I saw this and looked into it.  I think it's because we test against desisim-testdata/master, which is now using the not-yet-supported v3.0 templates.  But unfortunately the latest tag is from Feb 2017, before the v2.6 templates.

So one immediate solution would be to cut a desisim-testdata tag tied to a commit *preceding* https://github.com/desihub/desisim-testdata/pull/11 and add that version to the `.travis.yml` file, or wait until I push the `v3.0` basis templates through (today).

sbailey:
Thanks @moustakas .  I made a desisim-testdata/0.5.0 tag from the version prior to your v3.0 templates update, and updated this repo to use that tag.

Minor stylistic change: I also updated the `Z_noRSD` logic from try/except to if/else so that we don't accidentally catch and pass on real errors.

alxogm:
OK. Thanks @sbailey

PR #439, opened 2018-10-19
support new BGS target cuts

moustakas:
Addresses the updated BGS target cuts implemented in https://github.com/desihub/desitarget/pull/407.


moustakas:
Merging this now so we can restore `master` back to a functioning state.  Changes should not be controversial.

PR #446, opened 2018-11-06
Update to new fibermap format

sbailey:
This is a companion PR to desihub/desispec#717 to update to using the new fibermap format (see desihub/desispec#679).  The most significant changes are:
* switching from MAG[5] to FLUX_G, FLUX_R, FLUX_Z, FLUX_W1, FLUX_W2
* OBJTYPE values are now just TGT, SKY, and BAD, instead of ELG, LRG, QSO, etc. to avoid ambiguity when a given target passes more than one cut. Use DESI_TARGET and desitarget.targetmask.desi_mask instead.

A reference run is available at `/global/cscratch1/sd/sjbailey/desi/dev/end2end/`

Let's evaluate desihub/desispec#679 first, and then proceed with this.


julienguy:
test are failing because of the call to desispec.io.empty_fibermap with the master version of desispec.
we should merge the desispec branch and reload the travis tests in this repo.

sbailey:
Failed tests was a rst/sphinx typo fixed in a [ci skip] push so that I don't have to re-wait for all the code tests.  Merging now.

Issue #447, opened 2018-11-07
quickquasars should add FoG effect to the quasar redshift before generating templates


andreufont:
Right now we are reading quasar redshifts from the transmission files, and use these to generate the random quasar continua. 

Later on, we add a random Fingers of God (FoG) correction to the quasar redshift, but this shift is not applied to the quasar continua. 

This is probably not a large effect for things like BAO, but if some is trying to use the synthetic spectra to generate composite quasar spectra, the noise in the composite will  be large underestimated. 

Note that this should not apply to the "redshift fitting error", since this should not modify the "true quasar spectrum", only our fake measurement of its redshift. 

This is marginally related to issues #441 and #442. It is not a priority either, I'm just putting it here so that we do not forget about it.

alxogm:
Hi @andreufont so I moved the code to add the random FoG correction to quasar redshift before generating the template 
See [ #L399](https://github.com/desihub/desisim/blob/20ae01f46b94e424b4b86bb46a3574b2e31f01f3/py/desisim/scripts/quickquasars.py#L399)

This is also affecting the addition of BALs. I haven't tested if the effect is large, but probably is best to use the no_FOG redshift right for the addition of BALs? 

thanks 
alma

andreufont:
@alxogm - It looks good. Why do you say we should use the no_FOG for the BALs? I would consider that the true redshift of the quasar, the one we would have if we had perfect N-body simulations, would be z + dz_fog, so that is the one I would use.

alxogm:
@andreufont, I was confused on whether it should affect the inclusion of all DLAs, BALs, metals... and the shift we are adding in #442 (which we clarified)

If considered as the "true redshift" then I'll move this lines  to use it everywhere [#L250](https://github.com/desihub/desisim/blob/20ae01f46b94e424b4b86bb46a3574b2e31f01f3/py/desisim/scripts/quickquasars.py#L250)

do you agree?

andreufont:
Yes, I would add it right before that line.

londumas:
@alxogm, Seems to be fixed. Can we close?

Issue #441, opened 2018-10-24
quickquasars: add Lorentzian 


londumas:
Figure 4 of Zarrouk et al 2018 (https://arxiv.org/pdf/1801.03062.pdf) shows that the redshift error are Lorentzian.

andreufont:
Is the argument that FoG is Lorentzian, or that the redshift errors are? 

The modelling of FoG can be informed from Nbody simulations, I guess. 

If we decided that either redshift errors or FoG are better modelled by a Lorentzian, this should be trivial to change.

londumas:
@alxogm and @andreufont, here is what I had in mind:
```
import scipy as sp
import scipy.stats
import scipy.constants
def lorentzian(nb, mean, var):
    return sp.stats.cauchy.rvs(loc=mean,scale=var,size=nb)
def gaussian(nb, mean, var):
    return sp.stats.norm.rvs(loc=mean,scale=var,size=nb)

mean = 0.
var = 200.
dv = lorentzian(nb, mean, var)
dv = gaussian(nb, mean, var)
```

In the following plot I compare these two functions to the difference between two estimator of redshift
in DR12Q.
I am planning to do the same plot with two redshift with the same estimator, but on two different epochs.

![deltav_distribution](https://user-images.githubusercontent.com/3166436/48079871-8c0a2d80-e1a9-11e8-8bde-b1e05167c60f.png)

![deltav_distribution_zoom](https://user-images.githubusercontent.com/3166436/48079873-8c0a2d80-e1a9-11e8-9033-67e4a4d57d24.png)

alxogm:
Hello @londumas, thank you for the plot. This is similar to what the plots I was doing when I thought of truncating the Cauchy distribution in the mod_cauchy function... I should've showed it instead of trying to explain it ...  I didn't have the DR12Q data to compare directly but by eye I thought just forcing the samples to be inside some limits (delta v<3000) could be sufficient.  Actually in the mock we are adding a gaussian with sigma=150(for the fingers of god effect),  and the Lorentzian. We can add another gaussian and choose the sigma to match the data.

Is that DR12Q data in NERSC? or is this plot in desi_QA?  I could compare how the mock looks like in comparison  and play with the different parameters, maybe more than one Lorentzian, with samples within different ranges? 

Thanks

andreufont:
[...]

PR #449, opened 2018-11-16
Qquasar addfeature

alxogm:
Hi @andreufont and @londumas  I think this is good to go now,  if you would like to take look and let me know if any other change is needed.  

@sbailey I see that the master branch has not passed some travis test, so it is likely this branch will not do it either, but is not related to this PR... do we wait until the master passes all tests? 

thanks
Alma


londumas:
Looks great, Thank you @alxogm.

py/desisim/scripts/quickquasars.py
```diff
@@ -574,14 +603,14 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
...
-        zbest["Z"]            = metadata['Z']+dz_fog
+        zbest["Z"]            = metadata['Z']
         zbest["ZERR"][:]      = 0.
 
         if args.sigma_kms_zfit:
            log.info("Added zfit error with sigma {} to zbest".format(args.sigma_kms_zfit))
-           sigma_zfit=(args.sigma_kms_zfit/299792.458)*(1.+metadata['Z'])
-           zbest["Z"]+=sigma_zfit*np.random.normal(0,1,nqso)
-           zbest["ZERR"]=sigma_zfit
+           sigma_zfit=mod_cauchy(loc=0,scale=args.sigma_kms_zfit,size=nqso,cut=3000)
```
andreufont: Hi @alxogm - Do I understand correctly that sigma_zfit here is a random variable with a shift in redshift? If so, we shouldn't call it sigma, since it suggests it has to do with a RMS, not an actual random variable. May be dz_fit? Just like we use dz_fog?

py/desisim/scripts/quickquasars.py
```diff
@@ -574,14 +603,14 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
...
+           zbest["Z"]+=sigma_zfit/c*(1.+metadata['Z'])
+           zbest["ZERR"]=args.sigma_kms_zfit/c*(1.+metadata['Z'])
```
andreufont: So sigma_kms_zfit used to be the Gaussian RMS of the redshift error. Now that it is a variable of the Lorentzian, is it still the case that the redshift error is described by this? It could be, I am just curious.

py/desisim/scripts/quickquasars.py
alxogm: Hello @andreufont, yes you are right. I'll change that.

py/desisim/scripts/quickquasars.py
alxogm: Hi @andreufont, it is a good question. Formally it is not, since for the Cauchy distribution the variance is not well defined.  I thought that args.sigma_kms_zfit is still defining the width of the distribution, and in that sense the error (difference) to the true redshift. But is true it is underestimated. Be could compute the RMS directly on the generated samples, and read it from them, if you think this will be important...

alxogm:
Summary of this PR: 
- Adds a random shift to the true redshift. This is to simulate as if we have run redrock to find the redshift from the spectra. This shift is now sampled from a modified Cauchy distribution instead of a Normal (as it was before), this is motivated by the distribution observed in Zarrouk et al 2018 (https://arxiv.org/pdf/1801.03062.pdf) See #441. [L609](https://github.com/desihub/desisim/blob/fa4b663ae11a29211a98227ddeb8ba406ab4f405/py/desisim/scripts/quickquasars.py#L609 )


- Adds a shift to redshift according to table 4 of du Mas des Bourboux et al 2017 (https://arxiv.org/pdf/1708.02225.pdf)   [L571](https://github.com/desihub/desisim/blob/fa4b663ae11a29211a98227ddeb8ba406ab4f405/py/desisim/scripts/quickquasars.py#L571)

- Change the location for the addition of the FoG shift to the beginning of the simulation, so that it happens before generating the templates, see #447. [L255](https://github.com/desihub/desisim/blob/fa4b663ae11a29211a98227ddeb8ba406ab4f405/py/desisim/scripts/quickquasars.py#L255)

py/desisim/scripts/quickquasars.py
andreufont: I don't know what is the best thing to do, but I wouldn't worry too much, since not many codes will actually use zbest["ZERROR"] unless they are actually using results from Redrock (in which case we would not run with this option).

So probably, the best thing to do is to either not write that column, or write ZERR=0 or ZERR=NAN, just to make it clear that none should really believe this number. 

It was cleaner when using the Gaussian distribution, but if it is not easy to get a number now, I would just forget about it.

londumas:
@alxogm, Thanks for all the work. You have a lot of merging conflicts `<<<<<<< HEAD`. Please have a look and fix them.

py/desisim/scripts/quickquasars.py
```diff
@@ -491,13 +484,9 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
         if args.balprob<=1. and args.balprob >0:
             log.info("Adding BALs with probability {}".format(args.balprob))
             # save current random state
-<<<<<<< HEAD
-            rnd_state = np.random.get_state() 
-            tmp_qso_flux,meta_bal=bal.insert_bals(tmp_qso_wave,tmp_qso_flux,metadata['Z'],
```
londumas: @alxogm, are you sure you want to remove this?

py/desisim/scripts/quickquasars.py
```diff
@@ -611,23 +600,13 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
...
                 sourcetype="qso", skyerr=args.skyerr,ra=metadata["RA"],dec=metadata["DEC"],targetid=targetid,
                 meta=specmeta,seed=seed,fibermap_columns=fibermap_columns,use_poisson=False) # use Poisson = False to get reproducible results.
 
-<<<<<<< HEAD
+
 ##Adedd to write the truth file, including metadata for DLA's and BALs      
-=======
 
-##Adedd to write the truth file, includen metadata for DLA's and BALs
-    log.info("Added FOG to redshift with sigma {} to zbest".format(args.sigma_kms_fog))
```
londumas: @alxogm, are you sure you want to remove this?

py/desisim/scripts/quickquasars.py
```diff
@@ -636,14 +615,10 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
         meta.add_column(Column(metadata['Z_noRSD'],name='TRUEZ_noRSD'))
     else:
         log.info('Z_noRSD field not present in transmission file. Z_noRSD not saved to truth file')
-<<<<<<< HEAD
     if args.shift_kms_los:
        metadata['Z']+=(args.shift_kms_los/c*(1.+metadata['Z']))
        log.info('Added a shift of {} km/s to the redshift'.format(str(args.shift_kms_los)))
-=======
 
-    meta.add_column(Column(metadata['Z']+dz_fog,name='TRUEZ'))
```
londumas: @alxogm, are you sure you want to remove this?

alxogm:
@londumas and @andreufont I don't know what happened when I merged the master branch :(! I'll probably make a fresh PR to make sure I'm not deleting anything I shouldn't. Thanks!

PR #450, opened 2018-11-19
Eboss quickquasars

londumas:
Add the option in `quickquasars.py` to produce eBOSS like data:
- magnitude distribution
- footprint: inhomogeneities of the density and of the n(z)
- n(z)

Tested in `/project/projectdirs/desi/mocks/lya_forest/london/v4.0.0/eboss-0.2/` and it works very well.

histo_z

footprint

### DIFF
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -14,8 +14,7 @@ from desiutil.log import get_logger
 from desispec.io.util import write_bintable
 from desispec.io.fibermap import read_fibermap
 from desisim.simexp import reference_conditions
-from desisim.templates import SIMQSO
-from desisim.templates import QSO
+from desisim.templates import SIMQSO, QSO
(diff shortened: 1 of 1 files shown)


londumas:
@andreufont and @moustakas, Doesn't pass the Travis tests, but it doesn't seem to be because of anything this PR does.

moustakas:
@londumas Travis tests are failing in `test_quickquasars`, and this PR changed `quickquasars`, so you should please try to track down the problem (or update the unit test).

@andreufont For these types of "support" data files there are no data model or documentation requirements that I'm aware of (or have conformed to!), as long as it's clear where they're being used in the code.

londumas:
@moustakas, I will have a look, thanks.
@andreufont, For what is worth, I have added in the header of the two files the description of the files used to produce them, and the options: nside, nest, ...

londumas:
@moustakas, The issue was that the code was setting the magnitude limits to `None` to reach the default. Fixed now.

PR #452, opened 2018-11-21
Add feature to quickquasars

alxogm:
This PR supersede #449 due to some confusing conflicts when merging with master branch. 

Summary of this PR:

Adds a random shift to the true redshift. This is to simulate as if we have run redrock to find the redshift from the spectra. This shift is now sampled from a modified Cauchy distribution instead of a Normal (as it was before), this is motivated by the distribution observed in Zarrouk et al 2018 (https://arxiv.org/pdf/1801.03062.pdf) See #441. L609
Right now we completely changed from normal to lorentzian distribution. If you think we should keep the option of a gaussian, we can add it as an option with two parameters.

Adds a shift to redshift according to table 4 of du Mas des Bourboux et al 2017 (https://arxiv.org/pdf/1708.02225.pdf) L571

Change the location for the addition of the FoG shift to the beginning of the simulation, so that it happens before generating the templates, see #447. L255

### DIFF
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -27,6 +27,8 @@ from desimodel.io import load_pixweight
 from desimodel import footprint
 from speclite import filters
 from desitarget.cuts import isQSO_colors
+c = speed_of_light/1000. #- km/s
+
 
 def parse(options=None):
     parser=argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter,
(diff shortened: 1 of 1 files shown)


py/desisim/scripts/quickquasars.py
```diff
@@ -61,7 +66,8 @@ def parse(options=None):
 
     parser.add_argument('--sigma_kms_fog',type=float,default=150, help="Adds a gaussian error to the quasar redshift that simulate the fingers of god effect")
 
-    parser.add_argument('--sigma_kms_zfit',nargs='?',type=float,const=400,help="Adds a gaussian error to the quasar redshift, to simulate the redshift fitting step. E.g. --sigma_kms_zfit 200 will use a sigma value of 200 km/s. If a number is not specified the default is used.")
+    parser.add_argument('--sigma_kms_zfit',nargs='?',type=float,const=200,help="Adds a Lorentzian distributed shift to the quasar redshift, to simulate the redshift fitting step. E.g. --sigma_kms_zfit 200 will use a sigma value of 200 km/s. If a number is not specified the default is used.")
```
andreufont: Hi @alxogm - Last minor request: should we change the name sigma_kms_zfit to something else? If I read sigma_kms_fog and then sigma_kms_zfit, I have the impression that both are talking about Gaussian RMS, but the latter is not a Gaussian rms. Why call it "sigma"? What is the usual term used to refer to the width of a Lorentzian? gamma?

alxogm:
@moustakas Hi, this has been approved and passes all tests. Can we merge?

thanks

moustakas:
Unless you'd specifically like me to review a PR please don't wait for my approval, particularly for feature additions like this, etc.  If you end up changing things like templates or other more fundamental parts of the 1D simulation code then it might be good to get other sets of eyes on the changes.  But otherwise please go ahead and merge yourself!

PR #453, opened 2018-11-29
precompute colors for star and galaxy templates

moustakas:
Simple PR which adds a command-line script to compute colors for ELG, BGS, and LRG templates on a uniform redshift grid from `z=0-2.5` and fixed (rest-frame) colors for STAR and WD templates.  The results are written to a dedicated HDU named `DESI-COLORS` which is only read by `io.read_basis_templates` if present (in other words, the changes here are backwards-compatible).

Note that for STAR templates we do not compute WISE colors because the templates do not extend red enough, but this issues is handled elsewhere (e.g., in `templates.STAR.make_templates` and `desitarget.mock.mockmaker.STARMaker`).

For the record, the following filters are used: `decam2014-*`, `BASS-*`, `MzLS-z`, and `sdss2010-*`.

The new set of templates (with this HDU included) will be added in a separate PR after some additional testing.

### DIFF
--- a/bin/precompute-template-colors
+++ b/bin/precompute-template-colors
@@ -18,10 +18,7 @@ def get_zgrid():
     """Minimum, maximum, and redshift spacing.
 
     """
-    if False: # for testing
-        zmin, zmax, dz = 0.0, 0.5, 0.1
-    else:
-        zmin, zmax, dz = 0.0, 2.5, 0.25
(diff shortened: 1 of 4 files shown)


moustakas:
@sbailey I think this is ready to be merged.  None of the changes here *require* the `v3.1` basis templates, plus this PR (indirectly) fixes https://github.com/desihub/desitarget/issues/397.

PR #455, opened 2018-12-03
fix error in size of Z_noFOG

londumas:
fix error in size of Z_noFOG, the selection in the eBOSS case was not applied.


alxogm:
@londumas Thanks for catching that. You can merge either before or after the other PR we have open, since I think they are not in conflict. thanks again.

londumas:
@alxogm, could you review and "approve"?

Issue #456, opened 2018-12-03
quickquasars: TARGETID is not correct in truth-16-*.fits


londumas:
I think that quickquasars doesn't save TARGETID with enough precision. Example:

import fitsio
h = fitsio.FITS('/global/projecta/projectdirs/desi/mocks/lya_forest/london/v4.0.0/eboss-0.3/spectra-16/1/160/spectra-16-160.fits')
print(h[1]['TARGETID'][:][0])

h = fitsio.FITS('/global/projecta/projectdirs/desi/mocks/lya_forest/london/v4.0.0/eboss-0.3/spectra-16/1/160/zbest-16-160.fits')
print(h[1]['TARGETID'][:][0])

h = fitsio.FITS('/global/projecta/projectdirs/desi/mocks/lya_forest/london/v4.0.0/eboss-0.3/spectra-16/1/160/truth-16-160.fits')
print(h[1]['TARGETID'][:][0])

100140083
100140083
0
```

```
import fitsio
h = fitsio.FITS('/global/projecta/projectdirs/desi/mocks/lya_forest/london/v4.0.0/eboss-0.3/spectra-16/1/137/spectra-16-137.fits')
print(h[1]['TARGETID'][:][0])

h = fitsio.FITS('/global/projecta/projectdirs/desi/mocks/lya_forest/london/v4.0.0/eboss-0.3/spectra-16/1/137/zbest-16-137.fits')
print(h[1]['TARGETID'][:][0])

h = fitsio.FITS('/global/projecta/projectdirs/desi/mocks/lya_forest/london/v4.0.0/eboss-0.3/spectra-16/1/137/truth-16-137.fits')
print(h[1]['TARGETID'][:][0])

80105552
80105552
0
```

PR #457, opened 2018-12-03
Fix quickquasars targetid truth

londumas:
Fix issue https://github.com/desihub/desisim/issues/456:
    - Correct the TARGETID of truth for all 4 HDUs
    - Change `z` to `Z` for DLA

### DIFF
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -621,7 +621,7 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
 
     ### Add a shift to the redshift, simulating the statistic imprecision of redrock
     if args.gamma_kms_zfit:
-        log.info("Added zfit error with sigma {} to zbest".format(args.gamma_kms_zfit))
+        log.info("Added zfit error with gamma {} to zbest".format(args.gamma_kms_zfit))
         DZ_stat_shift = mod_cauchy(loc=0,scale=args.gamma_kms_zfit,size=nqso,cut=3000)/c*(1.+metadata['Z'])
         meta['REDSHIFT'] += DZ_stat_shift
         metadata['Z'] += DZ_stat_shift
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -461,8 +461,8 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
                     lyaforest=False, nocolorcuts=True,
                     noresample=True, seed=seed, south=issouth)
 
-        _meta['TARGETID'][:] = metadata['MOCKID']
-        _qsometa['TARGETID'][:] = metadata['MOCKID']
+        _meta['TARGETID'] = metadata['MOCKID'][these]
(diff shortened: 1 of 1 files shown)


py/desisim/scripts/quickquasars.py
```diff
@@ -462,8 +460,8 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
                     lyaforest=False, nocolorcuts=True,
                     noresample=True, seed=seed, south=issouth)
 
-        _meta['TARGETID'] = metadata['MOCKID']
-        _qsometa['TARGETID'] = metadata['MOCKID']
+        _meta['TARGETID'][:] = metadata['MOCKID']
+        _qsometa['TARGETID'][:] = metadata['MOCKID']
```
alxogm: Those two lines are what fixes the original bug?

py/desisim/scripts/quickquasars.py
```diff
@@ -494,6 +492,9 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
             # restore random state to get the same random numbers later
             # as when we don't insert BALs
             np.random.set_state(rnd_state)
+            meta_bal['TARGETID'] = metadata['MOCKID']
+            w = meta_bal['TEMPLATEID']!=-1
+            meta_bal = meta_bal[:][w]
```
alxogm: Still there is more information that we probably want to the metadata, like the BAL index, which I'm not sure is the same as the balnicity, Do you know?  this can be added in another issue anyway.

py/desisim/scripts/quickquasars.py
londumas: Yes, it is. Plus some other for DLA and BAL.

py/desisim/scripts/quickquasars.py
londumas: @alxogm, Yes it can be added later.

londumas:
@alxogm, could you review again? I made some changes since the Lorentz smoothing was not stored in the truth file.
I decided to use `DZ_FOG`, `DZ_SYS` and `DZ_STAT`.
Thank for the help.

londumas:
@andreufont and @alxogm, here is a summary of this PR.
I fixes a bug that was giving the wrong TARGETID to the truth file.
I changed the output of the truth file to give:
- 'Z': final redshift
- 'Z_INPUT': input redshift with RSD
- 'DZ_FOG': delta z contribution of the Finger-Of-God
- 'DZ_SYS': delta z contribution of the systematic shift
- 'DZ_STAT': delta z contribution of the statistic shift (Lorentzian) 
- 'Z_NORSD': redshift without RSD

Do you both agree with these changes?

alxogm:
@londumas  I'm only worried that we have been using TRUEZ and we'll get confused by changing to Z_input.  But I don't really know what is the most informative name, so I leave it to you and @andreufont to decide

londumas:
@alxogm, is `TRUEZ` used by anyone for the moment?

londumas:
I have no issue with changing back `Z` to `TRUEZ`.
It is `TRUEZ_noFOG` that I don't like.

londumas:
@alxogm, ok I understand better. I made the corrections.
Could you do a final review and give me the green light?

Issue #468, opened 2019-02-16
Add dust-extinction in quickquasar


londumas:
Add dust-extinction in quickquasar by default, to simulated data coming from desi.
The desi data will not be corrected for dust-extinction.

moustakas:
In case it's helpful, we do this in `desitarget.mock.mockmaker` [here](https://github.com/desihub/desitarget/blob/master/py/desitarget/mock/mockmaker.py#L298-L312) and [here](https://github.com/desihub/desitarget/blob/master/py/desitarget/mock/mockmaker.py#L1038-L1039).

alxogm:
Thanks @moustakas. It is very helpful, I'm working in it...

alxogm:
Sorry for the delay, I've missed to comment the progress on this. I've tried to follow the way it is done in desitarget.mockmaker.  See https://github.com/desihub/desisim/compare/qquasar_extintion
I add here a couple of plots, of the simulated spectra with and without extinction, and the result on the 1D CF... @londumas  you might have a best feeling of whether the change is as expected...   should I do the 3DCF?  I'm not making a PR yet

![Screen Shot 2019-03-15 at 11 03 00 AM](https://user-images.githubusercontent.com/14896082/54449266-4f2fe380-4713-11e9-8538-176cb793d94a.png)
![Screen Shot 2019-03-15 at 11 13 16 AM](https://user-images.githubusercontent.com/14896082/54449294-5d7dff80-4713-11e9-9e5a-ffea8a2d1679.png)
![Screen Shot 2019-03-15 at 11 13 12 AM](https://user-images.githubusercontent.com/14896082/54449296-5f47c300-4713-11e9-937c-9b8932e6b424.png)

alxogm:
@londumas I also wanted to ask if this option should apply when --eboss is used ?

londumas:
@alxogm, thanks for doing that. Looks good. I don't think we need to look at the 3d correlation. And yes, we need this option for all type of surveys. Simply add something like `--dust-extinction`.

londumas:
@alxogm, can we close as fixed? I think you did all this job.

alxogm:
yes, lets close it. we'll rise a new issue if something comes up. Thanks

andreufont:
[...]

PR #472, opened 2019-03-19
Qquasar extintion

alxogm:
Adds dust-extinction to quickquasars .  It was added as default as said originally  in [#468](https://github.com/desihub/desisim/issues/468).

Two new flags were added. 

--no-extinction 
--Rv to modify the extinction factor (default:Rv=3.1)

@londumas let me know if this ok, or is better to have no extinction by default...


py/desisim/scripts/quickquasars.py
```diff
@@ -607,6 +611,16 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
     else :
         fibermap_columns=None
 
+# Attenuate the spectra for extinction
```
londumas: comment should be on same indent as code

py/desisim/scripts/quickquasars.py
```diff
@@ -80,6 +82,8 @@ def parse(options=None):
     parser.add_argument('--no-simqso',action = "store_true", help="Does not use desisim.templates.SIMQSO to generate templates, and uses desisim.templates.QSO instead.")
     parser.add_argument('--no-transmission',action = 'store_true', help='Do not multiply continuum by transmission, use F=1 everywhere')
     parser.add_argument('--eboss',action = 'store_true', help='Setup footprint, number density, redshift distribution, and exposure time to generate eBOSS-like mocks')
+    parser.add_argument('--Rv',type=float,default=3.1,help='Adds Galactic extintion, for the specified extintion factor. (default:Rv=3.1')
```
londumas: no need to write "(default:Rv=3.1')". `quickquasar --help` already give this information

py/desisim/scripts/quickquasars.py
```diff
@@ -607,6 +611,16 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
     else :
         fibermap_columns=None
 
+# Attenuate the spectra for extinction
+    if not sfdmap is None:
+       log.info("Dust extintion added with Rv={}".format(args.Rv))
+       indx=np.arange(len(metadata['RA']))
```
londumas: `np.arange(metadata['RA'].size)` is nicer, less parenthesis leads to less bugs

londumas:
@alxogm, Thanks for the work.
I think you will have to add importing the package in travis.
```
ImportError: Failed to import test module: desisim.test.test_quickquasars
  File "/home/travis/miniconda/envs/test/lib/python3.5/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/home/travis/miniconda/envs/test/lib/python3.5/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/travis/build/desihub/desisim/py/desisim/test/test_quickquasars.py", line 4, in <module>
    from desisim.scripts import quickquasars
  File "/home/travis/build/desihub/desisim/py/desisim/scripts/quickquasars.py", line 33, in <module>
    from desiutil.dust import ext_odonnell
ImportError: cannot import name 'ext_odonnell'
```

alxogm:
> @alxogm, Thanks for the work.
> I think you will have to add importing the package in travis.
> 

I thought so too, but I don't see why, other desiutil functions are being used from quickquasars without problem... still I'm trying it.

py/desisim/scripts/quickquasars.py
```diff
@@ -80,6 +82,8 @@ def parse(options=None):
     parser.add_argument('--no-simqso',action = "store_true", help="Does not use desisim.templates.SIMQSO to generate templates, and uses desisim.templates.QSO instead.")
     parser.add_argument('--no-transmission',action = 'store_true', help='Do not multiply continuum by transmission, use F=1 everywhere')
     parser.add_argument('--eboss',action = 'store_true', help='Setup footprint, number density, redshift distribution, and exposure time to generate eBOSS-like mocks')
+    parser.add_argument('--Rv',type=float,default=3.1,help='Adds Galactic extintion, for the specified extintion factor.)
```
londumas: @alxogm, you forgot the `'` at the end of the command line. That is the reason travis crashed.

alxogm:
@londumas I think the problem is not in the travis test file... because it does import the  SFDMap from the same desiutil.dust in the previous line...

@moustakas any idea of why I can not import ext_odonnell from desiutil.dust? I'm getting this error even though it works well in cori with the master  desi_enviroment  

ERROR: desisim.test.test_quickquasars (unittest.loader._FailedTest)
ImportError: Failed to import test module: desisim.test.test_quickquasars
  File "/home/travis/miniconda/envs/test/lib/python3.6/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/home/travis/miniconda/envs/test/lib/python3.6/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/travis/build/desihub/desisim/py/desisim/test/test_quickquasars.py", line 4, in <module>
    from desisim.scripts import quickquasars
  File "/home/travis/build/desihub/desisim/py/desisim/scripts/quickquasars.py", line 33, in <module>
    from desiutil.dust import ext_odonnell
ImportError: cannot import name 'ext_odonnell'

moustakas:
Indeed, Travis was using an older tag of `desiutil`, after we had added the SFD stuff but before the O'Donnell extinction curve.

moustakas:
The new failure is because we need a dust map within the Travis environment, unfortunately.  You should be able to copy the small maps that are used in `desiutil` [here](https://github.com/desihub/desiutil/tree/master/py/desiutil/test/t) and [here](https://github.com/desihub/desiutil/blob/master/py/desiutil/test/test_dust.py).

alxogm:
> The new failure is because we need a dust map within the Travis environment, unfortunately. You should be able to copy the small maps that are used in `desiutil` 

Thanks for the help @moustakas , I'm trying it...

alxogm:
Hi @moustakas a new error appeared, now it finds the maps in the test/data directory but then crashes when actually using it....   These maps you pointed to are supposed to be small versions of those in $DUST_DIR?(/global/project/projectdirs/cosmo/data/dust/v0_1/maps/)  Because I tried to set the later manually in the travis file  and it works fine...  but not with the small ones.... I'm testing this here to see if works as in my computer... 

If this is problematic, I think it is not really necessary to have it in the travis file, I could do quickquasars --no-extintion  

ERROR: test_quickquasars (desisim.test.test_quickquasars.Testquickquasars)
  File "/home/travis/build/desihub/desisim/py/desisim/test/test_quickquasars.py", line 51, in test_quickquasars
    quickquasars.main(opts)
  File "/home/travis/build/desihub/desisim/py/desisim/scripts/quickquasars.py", line 815, in main
    bal=bal,sfdmap=sfdmap,eboss=eboss)
  File "/home/travis/build/desihub/desisim/py/desisim/scripts/quickquasars.py", line 619, in simulate_one_healpix
    EBV = sfdmap.ebv(metadata['RA'],metadata['DEC'], scaling=1.0)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/desiutil/dust.py", line 394, in ebv
    interpolate)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/desiutil/dust.py", line 248, in ebv
    return _bilinear_interpolate(self.data, y, x)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/desiutil/dust.py", line 186, in _bilinear_interpolate
    (1.0-xw) * yw       * data[y1, x0] +
IndexError: index -614 is out of bounds for axis 1 with size 10
Ran 63 tests in 235.116s
FAILED (errors=1, expected failures=1)

andreufont:
Hi @alxogm @londumas - it would be great if we could quickly discuss this at the next telecon. What exactly are we talking about here? Is this about a extinction caused by our galaxy, a slowly varying function of angular position in the sky? Does it have any spectral feature, or it is just a power law kind of extinction (I guess so, right?).

So, BOSS/eBOSS corrected for that, but DESI will not correct for that? Do we know why?

I guess if we do not correct for that, it will mean that the "mean quasar continuum" will change with angular position. Or if the differences are just a power law type of extinction, it means that when we fit a power law on top of the continuum, the distribution of power law coefficients will vary accross the sky. Is that it? If so, it should not be a big issue. 

About having this option set by default, I have mixed feelings. I don't know whether the default setting for quickquasars should be "maximum realism" or a "simplified version". But if we decide we want the default to be "maximum realism", we should also by default add DLAs, BALs, metals, lorentzian redshift errors... 

I guess the main reason to add it by default is that otherwise the simulated quasars are too bright? the signal-to-noise in the simulated is too large? @moustakas , when we run quickquasars (or quickspectra) in different parts of the sky, with different extinction, is the exposure time adjusted to compensate for the difference in signal to noise? Because this is what will happen in the survey, right?

moustakas:
> Hi @moustakas a new error appeared, now it finds the maps in the test/data directory but then crashes when actually using it.... These maps you pointed to are supposed to be small versions of those in $DUST_DIR?(/global/project/projectdirs/cosmo/data/dust/v0_1/maps/) Because I tried to set the later manually in the travis file and it works fine... but not with the small ones.... I'm testing this here to see if works as in my computer...

@alxogm  The maps are most likely for a tiny bit of sky, so your test needs to use RA,Dec coordinates that overlap the map.  [Here](https://github.com/desihub/desiutil/blob/master/py/desiutil/test/test_dust.py#L32-L35) are the coordinates used in the `desiutil` which should work.

moustakas:
@andreufont 

> Hi @alxogm @londumas - it would be great if we could quickly discuss this at the next telecon. What exactly are we talking about here? Is this about a extinction caused by our galaxy, a slowly varying function of angular position in the sky? Does it have any spectral feature, or it is just a power law kind of extinction (I guess so, right?).

These code changes account for foreground Galactic extinction/reddening using a standard Milky Way extinction curve (from [O'Donnell 1994](http://adsabs.harvard.edu/abs/1994ApJ...422..158O)) with a nominal total-to-selective extinction ratio, `R_V=3.1`, appropriate for dust above and below the Galactic plane.  The curve is close to a power law (with more extinction in the blue, of course), but with some wavelength-dependent variations.  It's been included in `select_mock_targets` because all observed spectra will be reddened by Galactic dust, but it's up to the Lya WG whether it should be included in `quickquasars`; I like the optional argument that's been incorporated.

> So, BOSS/eBOSS corrected for that, but DESI will not correct for that? Do we know why?

When we perform redshift fitting we'll *definitely* correct for Galactic extinction (just as we do for the photometry, otherwise you'll introduce systematic effects that vary with Galactic latitude)---I just don't think we're doing so *now* (and I believe someone mentioned that eBOSS doesn't do this at all, but should).  What DESI *won't* do is release/write spectra that have been corrected for extinction since assumptions have to be made and maps of the MW dust improve over time; we'll keep the spectra written to disk "as observed", which is the right thing to do (again, just as we do for the imaging catalogs).

> I guess the main reason to add it by default is that otherwise the simulated quasars are too bright? the signal-to-noise in the simulated is too large? @moustakas , when we run quickquasars (or quickspectra) in different parts of the sky, with different extinction, is the exposure time adjusted to compensate for the difference in signal to noise? Because this is what will happen in the survey, right?

Yes, the observing strategy includes an exposure time boost in areas of higher Galactic extinction (among other factors like airmass), in order to achieve the ~same S/N at a nominal observed-frame wavelength.

andreufont:
Thanks @moustakas, this was very useful.

The only thing still unclear is whether quickquasars do change the exposure time or not. I know the survey strategy will, so the right thing would be to have quickquasars / quickspectra / specsim / simspec / others also rescale... but I'd like to have a confirmation. Because if specsim does take into account dust extinction when setting the actual exposure time, then somewhere in desisim we already have the dust maps, and we should not be allowed to change Rv in quickquasars without passing that information to specsim.

moustakas:
@dkirkby or @sbailey should confirm, but I believe the target exposure time (during survey operations) will come from the `Next Field Selector`, in other words, `specsim`, etc., doesn't know anything about dust.  

Similarly, the survey simulations (e.g., for determining the survey margin) *do* account for the increased exposure time with Galactic dust, but those simulations don't ever generate spectra.

dkirkby:
specsim only simulates the instrument response, using whatever input SEDs it is provided (which generally should have dust extinction already applied).

The NextFieldSelector schedules tiles under the assumption that exposures will be longer for pointings with more dust, as measured by the median E(B-V).

andreufont:
Hi @dkirkby, so what you are saying is that we shoud be calling specsim with different arguments for --exptime, for different parts of the survey. Correct?

I guess this is how it is done in the end-to-end simulations (like minitest), but I don't think this is how it is done in quickquasars. We are now passing a single value of --exptime to quickquasars, and it is then passed to desisim.quickspectra.sim_spectra, that passes it to desisim.simexp.simulate_spectra. So I'm quite sure that current quickquasars simulations use the same exposure time in all exposures, regardless of place in the sky. What is probably fine, since we are probably using "perfect" conditions in all pointings?

Sorry if I'm asking pretty basic questions, but is the target selection done on "extinction corrected magnitudes", or in "raw magnitudes"? In other words: do we expect to have the same density of targets (ignoring contaminants and cosmic variance) across the area, or do we expect less targets in region of higher extension, since fewer objects make the brightness cut?

moustakas:
Target selection cuts are applied on MW extinction-corrected magnitudes.

andreufont:
Good, so then exposure time *should* be longer in areas of higher extinction, to get the same SNR in spectra of two objects of same *extinction-corrected* magnitudes. I guess that's what David was saying.

Knowing that, it looks to me that quickquasars *should not* add variable extinction in the spectra, unless we are also taking this into account when choosing the exposure time in the different calls to specsim. Otherwise we'd end up having variable observed SNR for spectra with the same extinction-corrected magnitude.

alxogm:
> Good, so then exposure time _should_ be longer in areas of higher extinction, to get the same SNR in spectra of two objects of same _extinction-corrected_ magnitudes. I guess that's what David was saying.
> 
> Knowing that, it looks to me that quickquasars _should not_ add variable extinction in the spectra, unless we are also taking this into account when choosing the exposure time in the different calls to specsim. Otherwise we'd end up having variable observed SNR for spectra with the same extinction-corrected magnitude.

Sorry, I'm just trying to catch with the discussion here...  the current conclusion is that this addition don't go as default (if it goes at all)?. @andreufont although this don't change your argument, I think we don't actually apply target selection in current runs, the option is there, but I think we don't use it...  if we ever going to use it, then probably it is there were will need the extinction somehow applied even though we don't add the extinction to the spectra?

londumas:
> Good, so then exposure time _should_ be longer in areas of higher extinction, to get the same SNR in spectra of two objects of same _extinction-corrected_ magnitudes. I guess that's what David was saying.
> 
> Knowing that, it looks to me that quickquasars _should not_ add variable extinction in the spectra, unless we are also taking this into account when choosing the exposure time in the different calls to specsim. Otherwise we'd end up having variable observed SNR for spectra with the same extinction-corrected magnitude.

@andreufont, I disagree. There are two questions at stake here:
 - what is the effect of extinction on SNR?
 - what is the effect of extinction on the 1d power spectrum and 3d correlation.?

To the first question, if I understand well the discussion above, it is none, since the exposure is changed to correct for extinction.
To the second question, we don't know exactly the answer. When moddeling the quasar continuum by a slope, then this effect should disapear, but when modeled by a constant, like in 1d power spectrum studies, this effect should have effect similar to the effect of the slope of the continuum in Solene's paper.
So it is very valuable to have at least an option to add extinction, to see the effect on 1d power spectrum.

alxogm:
@andreufont may be is not very complicated to adjust the exposure time if we want to do it for quickquasars.  I think this is how it is done in what @dkirkby mentioned about the NextFieldSelector 

https://github.com/desihub/desisurvey/blob/c112f038e31817af1b2fbb140178522d47f90085/py/desisurvey/etc.py#L82

we could set a fixed EBV0 value...

So, maybe we could leave the optional argument to add the extinction including an adjustment to the exposure time using the median EBV for the pixel

andreufont:
> > Good, so then exposure time _should_ be longer in areas of higher extinction, to get the same SNR in spectra of two objects of same _extinction-corrected_ magnitudes. I guess that's what David was saying.
> > Knowing that, it looks to me that quickquasars _should not_ add variable extinction in the spectra, unless we are also taking this into account when choosing the exposure time in the different calls to specsim. Otherwise we'd end up having variable observed SNR for spectra with the same extinction-corrected magnitude.
> 
> @andreufont, I disagree. There are two questions at stake here:
> 
>     * what is the effect of extinction on SNR?
> 
>     * what is the effect of extinction on the 1d power spectrum and 3d correlation.?
> 
> 
> To the first question, if I understand well the discussion above, it is none, since the exposure is changed to correct for extinction.

As you say, it is none if exptime is changed. But we don't have the capability right now to do so, so I would not add this by default.

> To the second question, we don't know exactly the answer. When moddeling the quasar continuum by a slope, then this effect should disapear, but when modeled by a constant, like in 1d power spectrum studies, this effect should have effect similar to the effect of the slope of the continuum in Solene's paper.
> So it is very valuable to have at least an option to add extinction, to see the effect on 1d power spectrum.

Sure, it is good to have an option, but for now I would not have it on *by default*. If we figure out how to change the exposure times, then I'd be happy to set it by default.

alxogm:
Hi @andreufont @londumas I've included the update of the exposure times following https://github.com/desihub/desisurvey/blob/c112f038e31817af1b2fbb140178522d47f90085/py/desisurvey/etc.py#L82.

However I'm not quite sure of the reference parameter EBV0 to use here... for now what I have in mind is to define a threshold for which the exptime will be scaled, but I don't have a good sense of what this value would be... If the point is to have the same SNR as areas that have non-or very few extinction then I could use the minimum EBV value found in the footprint, but in that case wouldn't I be removing all the effect we want to see?
Then I thought of using de median value in the footprint, or some other percentile, but seems arbitrary... 


@dkirkby what EBV0 is to be used in NextFieldSelector for instance? I could not find it.

thanks 
Alma

moustakas:
`E(B-V)=0` is a good reference number!

londumas:
@alxogm, tell us when you think it is ready to do a last review and to be merged. Thanks for the work.

andreufont:
Hi @alxogm , sorry I have been silent. While I think it is good to have extra options in quickquasars, I'm worried of increasing too much the complexity of the code. Changing the exposure time is getting close to what I would consider the limit, I think anything fancier than this should just be done in the context of the proper survey simulations.

alxogm:
@londumas  I finalized testing it runs without stoping.  
@andreufont ok, will consider more carefully new additions. Thanks for the clarifying comments. So I think is done and ready for a last review.   I left this addition as optional, until we start using it and make sure the change in the exposure times has the desired effect, and then decide if it goes as default... So I think it could be merged if you agree...

py/desisim/scripts/quickquasars.py
```diff
@@ -29,6 +29,8 @@
 from desimodel import footprint
 from speclite import filters
 from desitarget.cuts import isQSO_colors
+from desiutil.dust import SFDMap
```
londumas: `from desiutil.dust import SFDMap, ext_odonnell` is cleaner

py/desisim/scripts/quickquasars.py
```diff
@@ -80,6 +82,7 @@ def parse(options=None):
     parser.add_argument('--no-simqso',action = "store_true", help="Does not use desisim.templates.SIMQSO to generate templates, and uses desisim.templates.QSO instead.")
     parser.add_argument('--no-transmission',action = 'store_true', help='Do not multiply continuum by transmission, use F=1 everywhere')
     parser.add_argument('--eboss',action = 'store_true', help='Setup footprint, number density, redshift distribution, and exposure time to generate eBOSS-like mocks')
+    parser.add_argument('--extintion-Rv',nargs='?',type=float,const=3.1,help='Adds Galactic extintion, for the specified extintion factor. E.g. --extintion-Rv 3.5 will use Rv of 3.5. If a number is not specified, a value of 3.1 is used.')
```
londumas: Please split comments on multiple lines so it can be read without moving cursor.

londumas:
@alxogm, looks good. I simply have two esthetic comments.

py/desisim/scripts/quickquasars.py
```diff
@@ -82,7 +81,9 @@ def parse(options=None):
     parser.add_argument('--no-simqso',action = "store_true", help="Does not use desisim.templates.SIMQSO to generate templates, and uses desisim.templates.QSO instead.")
     parser.add_argument('--no-transmission',action = 'store_true', help='Do not multiply continuum by transmission, use F=1 everywhere')
     parser.add_argument('--eboss',action = 'store_true', help='Setup footprint, number density, redshift distribution, and exposure time to generate eBOSS-like mocks')
-    parser.add_argument('--extintion-Rv',nargs='?',type=float,const=3.1,help='Adds Galactic extintion, for the specified extintion factor. E.g. --extintion-Rv 3.5 will use Rv of 3.5. If a number is not specified, a value of 3.1 is used.')
+    parser.add_argument('--extintion-Rv',nargs='?',type=float,const=3.1,help='Adds Galactic extintion, for the specified extintion factor. \
+E.g. --extintion-Rv 3.5 will use Rv of 3.5. \
+If a number is not specified, a value of 3.1 is used.')
```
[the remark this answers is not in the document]
alxogm: My text editor shows me  everything right, I hope the change I made is what you wanted...

londumas:
@alxogm, it is perfect. Thank you for your time.

alxogm:
> Hi @alxogm - Sorry for the silence.
> 
> I always get confused with nargs='?'. Let's see if I got it. If someone runs with:
> quickquasars --extintion-Rv (use extinction_Rv=3.1)
> quickquasars --extintion-Rv 3.5 (use extinction_Rv=3.5)
> quickquasars (use extinction_Rv=None)
> 
> If this is how it works, then I'm happy. I haven't checked that the timeexp scaling is correct, but we can worry about it when we actually use the extinction tests.

Exactly this is how it works...and yes, we'll check the scaling is correct when using it.

moustakas:
I missed this earlier in the thread, but `R(V)=A(V)/E(B-V)` is the total-to-selective extinction ratio, and is a property of the dust along the line-of-sight.  Above the Galactic plane it's largely constant, ~3.1.  The variable you should be varying is the reddening, `E(B-V)` (which then yields `A(V)` through an assumed `R(V)`).  

You can look up how much `E(B-V)` varies throughout the DESI footprint from the [tile file](https://github.com/desihub/desimodel/blob/master/py/desimodel/io.py#L150).

alxogm:
> I missed this earlier in the thread, but `R(V)=A(V)/E(B-V)` is the total-to-selective extinction ratio, and is a property of the dust along the line-of-sight. Above the Galactic plane it's largely constant, ~3.1. The variable you should be varying is the reddening, `E(B-V)` (which then yields `A(V)` through an assumed `R(V)`).
> 
> You can look up how much `E(B-V)` varies throughout the DESI footprint from the [tile file](https://github.com/desihub/desimodel/blob/master/py/desimodel/io.py#L150).

@moustakas I don't get it...I use R(V)   to get the extinction curve from O'donell
l 
https://github.com/desihub/desiutil/blob/e55f88186381490fe4736e72e0bd7fd3dc28bbcb/py/desiutil/dust.py#L23

Then this is used together with the E(B-V) which varies with the RA/DEC and is got from the SFDMaps, to attenuate the spectra  ... that is how I understood it works here ...

https://github.com/desihub/desitarget/blob/e03f809a70badf121f2848368c50e612dcb7298e/py/desitarget/mock/mockmaker.py#L1041

You say it is not necessary to leave this option to vary R(V)?

moustakas:
The `R_V` argument in `dust.ext_odonnell` is a convenience because the O'Donnell (1994) parameterization includes a dependence on `R_V`, but we shouldn't generally change it from the default, 3.1.  

In other words, I would change `--extintion-Rv` to `--extinction` (or something similar, but note the correct spelling) and make it a Boolean, using `E(B-V)` from the dust maps according to each object's Galactic coordinates.

alxogm:
@moustakas Ok. Thanks! will make the change in a moment. I think it is the last thing before merge.

alxogm:
Hi,

if there are no more requests, I'll merge it. Thanks

PR #474, opened 2019-03-20
Modify simqsotemplate

alxogm:
This PR is a proposal to improve on the comparison between quasar continuum in quickquasar mocks and data. The original issue rised in https://github.com/imcgreer/simqso/issues/25 and has been studied in the LyaWG by Rodrigo see: https://github.com/igmhub/desi_QA/issues/8

I think is easier to modify the simqso emission lines from within desisim, than create a PR in simqso, specially if we need to modify it again in the future (notice this changes to the emission lines might not be final).  Also, up to now we (mostly Rodrigo) have been working this in a local branch of simqso. 

For this to work I need to:
-  locate somewhere the content of the file desi_simqso_model.py, currently in y/desisim/scripts. And clean it to the minimum content needed. 
- Add one parameter to SIMQSO.make_templates (and probably to QSO.make_templates for consistency on how it is called from quickquasars) to pass the option to use the modified simqso_model from quickquasars, in order to not affect other uses of simqso... 

The changes in this PR so far only modifies the Lya region. See plot made by Rodrigo (we think differences in amplitude are related to difference in magnitude range). 
lya_continuum_alma

 In order to modify the Lyb region (which also needs improvement), I need to add one more file: a modification to  
https://github.com/imcgreer/simqso/blob/master/simqso/data/emlinetrends_v7.fits to include three lines not present... or make a PR in simqso. 

Before going any further I'd like to see if  @moustakas @andreufont @londumas you agree with the idea of the proposal. The PR will still need some fixing.


py/desisim/scripts/quickquasars.py
```diff
@@ -79,6 +79,7 @@ def parse(options=None):
     parser.add_argument('--balprob',type=float,required=False, help="To add BAL features with the specified probability (e.g --balprob 0.5). Expect a number between 0 and 1 ")
     parser.add_argument('--no-simqso',action = "store_true", help="Does not use desisim.templates.SIMQSO to generate templates, and uses desisim.templates.QSO instead.")
     parser.add_argument('--no-transmission',action = 'store_true', help='Do not multiply continuum by transmission, use F=1 everywhere')
+    parser.add_argument('--modified-emline',action = 'store_true', help='Do not multiply continuum by transmission, use F=1 everywhere')
```
londumas: @alxogm, the comment does not correspond. But anyway, I don't think we need an option. Let's use the new model always, there are no obvious reasons to use sometime the old one. Is there?

andreufont:
Hi @alxogm - I don't think we should have an option about emission lines in QSO, it doesn't make much sense. 

May be you can specify the emission line model at the same type that you setup the model object, i.e., in line 739 of quickquasars.py. I believe that should be cleaner, and then you call make_templates without worrying about the emission line model.

alxogm:
Hi @andreufont  that is why I stoped at this point, because I wanted to discuss it... For the option in SIMQSO called mod_emline I see the name was unfortunate... Let me call it sqmodel, because instead of looking for the model defined in https://github.com/imcgreer/simqso/blob/master/simqso/sqmodels.py#L222-L235,
as it does in lines
   2470          from simqso.sqmodels import get_BossDr9_model_vars

 I want it looks for the one defined by us, currently defined in file https://github.com/desihub/desisim/blob/modify_simqsotemplate/py/desisim/scripts/desi_simqso_model.py (this location should probably change) where the only difference (so far) is the emission lines, but in principle it allows other modifications...

This is so that SIMQSO keeps its current model for other uses of SIMQSO... If the model is going to be updated for all aplications, maybe it is enough to point directly from SIMQSO to the one we are defining...if is only for us, the we need a way to pass the model and overwrite the SIMQSO default one.... right? 

So I think the question is if it makes sense that SIMQSO can have an option called sqmodel that let us change the model... If so, yes I see now a better place to define it for quickquasar  is in line 739 of quickquasars.py...

py/desisim/scripts/quickquasars.py
alxogm: Yes, I'll fix it (remove it)

alxogm:
@andreufont The other option is to make a PR in simqso and define there a desimodel, but if we change it again in future, will have to go there and make the PR and so on... 

Maybe there is another option, but I need to explore it a bit more ... that is to use a file like https://github.com/imcgreer/simqso/blob/master/examples/bossqsos.py to define the model, but we need to see how to call this from simqso... Maybe @moustakas has seen this before?

alxogm:
The result of modifications for the Lyb region is shown in the plot. Rodrigo is still working on this...  the comparison here only shows that this modification of quickquasars is working ok. 
image1

I still don't know where to locate the new files, the emlinetrends.fits file I think it is ok in desist/data... but  I don't know about the sqmodels.py currently set on desisim/scripts/ . Any suggestions?

londumas:
@alxogm, Thank you very much to you and Rodrigo for the work.
What is preventing us from getting the more different complexity of the Lya and Lyb?
For example in Lyb forest it seems that still an emission line is missing around 975A
lyb_surounding
The same for the Lya complex emission lines of the emission at l~1060A and l~1125A.
lya

alxogm:
Hi Helion,  yes I need to change some lines because for example right now you can see that CIII977 is actually set at 985 (this is how it was done in some simqso example) so I will change and use that of table 4 in reference
 https://iopscience.iop.org/article/10.3847/0004-6256/151/6/155/pdf, which locates it at 977A

unless there is a more recent reference to guide us in which lines parameters to use...

alxogm:
@londumas So, I changed some of the lines to what appears in Table 4 of https://iopscience.iop.org/article/10.3847/0004-6256/151/6/155/pdf and I think it looks better in some lines but factors need to be re-adjusted for others... 

Screen Shot 2019-04-01 at 12 31 40 PM
Screen Shot 2019-04-01 at 12 31 46 PM

@moustakas I requested your approval since I touched templates.py, it was a minimal change at the end but still, let me know if you think this can be merged once we are happy with the final model...
 
@andreufont I can only assign two reviewers... but let me know if you have comments. Thanks

londumas:
@alxogm, Indeed it looks way better! Thanks for the work.

alxogm:
OK. We've changed the emission line table to match that of https://iopscience.iop.org/article/10.3847/0004-6256/151/6/155/pdf. I only left Lya from the original table in SIMQSO as it was looking more like the data... 

After that, some adjustment to a few lines were needed to get this plots. Although I've only look at the Lya/Lyb region... Rodrigo will look at the full wavelength range. How precise we want this to be?  So I think I could clean the PR to be ready for a final revision... 

Screen Shot 2019-04-04 at 3 25 09 AM
Screen Shot 2019-04-04 at 3 24 07 AM
Screen Shot 2019-04-04 at 3 24 03 AM

londumas:
@alxogm, Looks very good. I think it is already a very good version to merge, so tell me when I should do a last review. Thanks a lot.

alxogm:
@londumas @moustakas @andreufont I think I finish with this.

But I've been experiencing that sometimes my test runs very slowly on Cori. The tests are quickquasars runs with a downsampling of 0.1, in 10 nodes, in about 13 minutes. But the last tests took ~30 minutes. I tested running in individual transmission files and it takes the same time to run with the master branch and with this one. Therefore I think there is no problem with this changes. 

I was wondering if you can test it @londumas ?

Thanks in advance,

londumas:
@alxogm, I will not be able to test it, I leave you to do it, and we will the see effect on the next set of simulations. Thanks again for the work.

alxogm:
> @alxogm, I will not be able to test it, I leave you to do it, and we will the see effect on the next set of simulations. Thanks again for the work.

No worries, I made a full run in principle the same as  London v6.0.0/eboss0.0... It took about 13 minutes, which I think is ok. So I think, this PR is ready for a final review. Thanks

alxogm:
Hi @moustakas. We briefly discussed this at the Lya telecon today. It seems is a good enough change for lya, so I was wondering if we could merge this PR that only affects quickquasar mocks ( It would be useful to have the new templates for the next round of mocks that will be produced),  or  are you considering to make a more general change of the templates in desisim? ....   @londumas mentioned this could be useful also for the redshift efficiency. Is there anything else I could do for this?   

Thanks

moustakas:
Yes, please don't let me block this!  I'll open a new issue if anything arises when I get back to templates (eventually!).

PR #475, opened 2019-03-20
Update eBOSS magnitude to be closer to data

londumas:
Update eBOSS magnitude to be closer to data.
This is not exact science, but let's start with that and we will have a better picture after 10 realizations.


londumas:
old:
mag_distrib

new:
new

londumas:
@andreufont, does that work for you?

Issue #476, opened 2019-03-22
Lya should be `1215.67` everywhere


londumas:
In Lyman-alpha studies we use `lambda_lya = 1215.67 A`, a difference in this definition introduces a shift of the QSO x Lya cross-correlation. I don't know to what use are these three lines, however, they are the only one I found with a value different than `1215.67`. It would be nice to have a common definition across all DESI. Can we simply update these three lines?

https://github.com/desihub/desisim/blob/e45f93915f648c678e490a1ded5757124e7631f7/py/desisim/qso_template/qso_pca.py#L103

https://github.com/desihub/desisim/blob/e45f93915f648c678e490a1ded5757124e7631f7/py/desisim/qso_template/fit_boss_qsos.py#L109

https://github.com/desihub/desisim/blob/e45f93915f648c678e490a1ded5757124e7631f7/py/desisim/qso_template/fit_boss_qsos.py#L201

londumas:
I simply did a `grep '1215' -r * --include \*.py` to look at all the desi code I have on my Nersc home.

sbailey:
@profxj any objections (or implications) to changing these to 1215.67?  I think all of them reference deprecated code since we aren't using BOSS PCA QSO templates anymore (though we might revive the code for updating templates with DESI SV QSOs, so it is worth maintaining it).

moustakas:
This code is indeed deprecated now, but may be useful later.  In case @profxj misses this or is otherwise indisposed, I don't see any negative consequences to updating these wavelengths with a PR.

londumas:
Fixed in PR https://github.com/desihub/desisim/pull/478

PR #478, opened 2019-03-23
Update the value of the Lya absorption line

londumas:
Fix issue https://github.com/desihub/desisim/issues/476 by updating all found lya lines to `1215.67`.


londumas:
Searching for `1215`, it seems that these are the only remaining ones: https://github.com/search?q=org%3Adesihub+1215&type=Code

Issue #477, opened 2019-03-22
definition of speed of light is not consistent within DESI


londumas:
Doing the following simple grep on all the DESI packages, maybe there are other I have not thought of:
 - `grep '2.99' -r * --include \*.py`
 - `grep '299' -r * --include \*.py`
 - `grep '3e5' -r * --include \*.py`
 - `grep '3e8' -r * --include \*.py`
 - `grep '299792458' -r * --include \*.py`
 - `grep '2.99792458' -r * --include \*.py`

It shows that different definition of the speed of light is used in the overall desi packages.
It would be very useful to have every package use the same `from scipy.constants import speed_of_light` or `c = 299792458.`. This would allow consistency comparison and better definite results.
For Lyman-alpha this also has the possible impact of shifting the Lyman-alpha x QSO cross-correlation along the line-of-sight.

weaverba137:
[astropy.constants](http://docs.astropy.org/en/stable/constants/index.html) would also work, but we should pick one or the other and stick with it.

Also, be sure to check for `3E5`, or use `grep -i`!

moustakas:
Nice catch.

In [`desitarget.mock.mockmaker`](https://github.com/desihub/desitarget/blob/master/py/desitarget/mock/mockmaker.py#L29-L33) we do
```
try:
    from scipy import constants
    C_LIGHT = constants.c/1000.0
except TypeError: # This can happen during documentation builds.
    C_LIGHT = 299792458.0/1000.0
```

londumas:
The remaining, I could find, and I need to understand before changing are:
`desi/code/desispec/py/desispec/fluxcalibration.py:    ZP_fnu = ZP_flambda * wave**2 / (2.9979e18)  # c in A/s
`
and
`desi/code/specter/py/specter/throughput.py:        self._hc = 6.62606957e-27 * 2.99792458e8 * 1e10`

weaverba137:
It's 2025. Maybe it's time to fix this.

PR #479, opened 2019-03-23
import speed of light from scipy

londumas:
Move to a common definition of speed of light. Step toward fixing issue https://github.com/desihub/desisim/issues/477.


londumas:
@sbailey, could you have a review?

Issue #482, opened 2019-04-01
MOCKID is TARGETID in other hdu except DLA


londumas:
https://github.com/desihub/desisim/blob/9459dda8ab2bc7c9218f0bce64bc992849b9b4e7/py/desisim/scripts/quickquasars.py#L395

In other HDU we don't use MOCKID but TARGETID, it is very nice to have a common name everywhere. See for example:
https://github.com/desihub/desisim/blob/9459dda8ab2bc7c9218f0bce64bc992849b9b4e7/py/desisim/scripts/quickquasars.py#L499

londumas:
Fixed in https://github.com/desihub/desisim/pull/483

PR #483, opened 2019-04-01
Use TARGETID for DLA hdu

londumas:
Fix issue https://github.com/desihub/desisim/issues/482 by using `TARGETID` everywhere, before only DLA hdu was using `MOCKID`.


londumas:
@alxogm, I simply simply changed `MOCKID` to `TARGETID` and clean a bit the arguments, without any changes.

PR #488, opened 2019-05-02
Fix issue using the boost

londumas:
Fix issue using the boost. boost is float, not list

### DIFF
--- a/py/desisim/lya_spectra.py
+++ b/py/desisim/lya_spectra.py
@@ -197,7 +197,7 @@ def apply_metals_transmission(qso_wave,qso_flux,trans_wave,trans,metals,boost) :
     try:
         #for index in range(len(metals)):    FIND A WAY TO ASSOCIATE THE CORRESPONDING METAL TO ITS BOOST....
         
-        mtrans = { m:np.exp(-absorber_IGM[m]['COEF']*boost[idx]*tau) for idx,m in enumerate(metals) }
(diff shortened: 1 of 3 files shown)


alxogm:
@londumas the way we did it is to add one boost per metal line, so that we can move them independently... now that you have fixed the correct numbers for the lines, you don't need to use the boost option. 

If there will be more lines to change, I think is best to move each one independently

londumas:
@alxogm, so what you say is that either we get rid of the boost entirely, or we correct the way it is implemented so that master can work when not used? I let you @alxogm and @Andrea-MG what you prefer doing.

alxogm:
> @alxogm, so what you say is that either we get rid of the boost entirely, or we correct the way it is implemented so that master can work when not used? I let you @alxogm and @Andrea-MG what you prefer doing.


@londumas Yes, that is what I was going to say next. I don't think we need the boost option in the master, only the updates to the absorber_IGM COEFF right? I think that will be best, to avoid having an unused/confusing option... The boost option can be only in Andrea local branch.

londumas:
@alxogm, ok works for me. I am going to remove it.

alxogm:
@londumas So we are a bit confused : will you remove the boost option in this branch? or want  @Andrea-MG or me  take care of that?

londumas:
@alxogm, can you make a review? I updated the value of SiII(1260), since the power of ten was incorrect and I removed boost and some trailing spaces and one tab.

Issue #493, opened 2019-05-29
overwrite issues with desisim/master/bin/wrap-fastframe


michaelJwilson:
Lines 83/84 ignore the relevance of args.clobber.  If "to clobber", all simspec files should be appended, not just those that do not exist. 

Similarly, logfiles should be placed in the dir. referred to by --outdir rather than the dir. of original simspec.  Otherwise, e.g. there can be file permission issues when running on a preproduced set of simspec files.

michaelJwilson:
Ok, I think these changes are done:

i)  add args.clobber to generation of simspec list.  
ii)  place logfiles in outdir if provided, otherwise in original simspec as previously.  Logfiles inherit /night/exposure/ structure of simspec files in both cases. 
iii)  --cframe run off the sv branch to use appropriate bits.    

Largely goes through fine.  Output is here:

```/global/cscratch1/sd/mjwilson/svdc-summer2019/spectro/fframe/v1/``` 

Not sure how worrying 
```WARNING: Tried to get polar motions for times after IERS data is valid. Defaulting to polar motion from the 50-yr mean for those. This may affect precision at the 10s of arcsec lev\
el [astropy.coordinates.builtin_frames.utils]```

should be.  

My problem is that I already have a fork of the desisim repo. with changes to master for BEAST calculations.  So I can't submit a pull request, given I can't change the remote directory for this new sv fork to point to my github.  Any suggestions?  

The code itself is here:
```/project/projectdirs/desi/users/mjwilson/repos/desisim/bin/wrap-fastframe```

**Comment by michaelJwilson** [comment]
caveat:  --clobber is passed to fastframe by the wrapper, which doesn't parse --clobber as an argument.  Ultimately, desispec.io.write_frame assumes overwrite=True, and uses os.rename to move a .tmp to the destination, both irrespective of args.clobber.

**Comment by sbailey** [comment]
Short version: thanks, I'll incorporate your changes into my sv branch in PR #494 .

Long version:

Lesson learned on always working in a branch instead of directly in master.  From the DESI wiki UsingGit page:

**If you made changes to master and you want to retroactively turn that into a branch for further development and/or sending a pull request**
```
#- Go ahead and commit whatever you have
git commit -am "comment about what changed"

#- Create a new branch starting from this point
git branch newbranchname

#- Reset your local master to whatever github has
git reset --hard origin/master

#- checkout your branch and keep working
git checkout newbranchname
```

[...]

PR #494, opened 2019-05-29
support SV, especially SV1_DESI_TARGET

sbailey:
Updates to support the simulation chain with SV1_DESI_TARGET in prep for Survey Validation.

@michaelJwilson please try this "sv" branch with fastframe and report any problems to this PR and we'll get them fixed as part of this update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ env:
         - DESISIM_TESTDATA_VERSION=0.6.1
         - SPECSIM_VERSION=v0.12
         - DESISPEC_VERSION=0.26.0
-        - DESITARGET_VERSION=0.25.0
+        - DESITARGET_VERSION=0.29.1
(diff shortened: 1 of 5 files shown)


michaelJwilson:
Changes implemented so far look good.  I didn't look at the complete code, so I can't testify further changes aren't required.  I assume you want to deal with #493 on the sv branch first before merging?

sbailey:
Let's punt the remaining fast frame clobber question to a future issue / PR / decision and merge this now.  I think it's ok to have fastframe overwrite outputs when called.  The original reason for wrap-fastframe *not* doing that was for "pick up where you left off" -- if a job timed out you could just rerun the exact same wrap-fastframe command and it would just continue with any remaining pixels, without crashing on pre-existing pixels nor wastefully regenerating them, and without requiring the user to update the list of requested healpix.

PR #495, opened 2019-06-10
Qquasar readmetal

alxogm:
This PR allows to run quickquasars on the newest version of London mocks (>v7.3), which includes metal transmission in separate HDU's which is to be used for individual tunning of the metals. It is compatible with old format and with the final format once the metals are tunned individually. Optional argument names didn't change, only allows for a list of arguments instead of being a boolean argument. Works as follows:

--metals-from-file  all (default)   : Looks for "F_METALS" or "METALS" HDU in transmission file, to be used for old format or final format. 
--metals-from-file all-dev : Adds all the current metals added in the transmission files['F_SI1260','F_SI1207','F_SI1193','F_SI1190']    

--metals-from-file  SI1260 SI1207 : Adds only the specified metals if present in transmission file, rise an error otherwise... ( A comment here is that it would be easier to remember the sintaxis if we use the same notation  as in the --metals option... which for this example would be like --metals SII(1260), 'SII(1207') ) 

@andreufont I think this works and is not too messy. Maybe I can make it more compact in a function so that code is cleaner...  I've tested it only on small runs from the notebook. If you think is ok, then I can test it in a full run to make sure it don't crashes... or I can explore the config file option, but I don't think is needed...

### DIFF
--- a/py/desisim/lya_spectra.py
+++ b/py/desisim/lya_spectra.py
@@ -118,7 +118,7 @@ def read_lya_skewers(lyafile,indices=None,read_dlas=False,add_metals=False,add_l
         else:
             nolyb="No HDU with EXTNAME='F_LYB' in transmission file {}".format(lyafile)
             log.error(nolyb)
-            raise KeyError(nlyb) 
+            raise KeyError(nolyb) 
     
 
     if add_metals:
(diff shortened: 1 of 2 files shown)


py/desisim/lya_spectra.py
```diff
@@ -107,14 +111,48 @@ def read_lya_skewers(lyafile,indices=None,read_dlas=False,add_metals=False) :
...
-            log.error(nom)
-            raise KeyError(nom)
+    if (add_lyb):
+        if ("F_LYB" in h) :
+            lyb = h["F_LYB"].read()
+            trans*=lyb
+        else:
+            nolyb="No HDU with EXTNAME='F_LYB' in transmission file {}".format(lyafile)
+            log.error(nolyb)
+            raise KeyError(nlyb)
```
andreufont: this should probably be nolyb, isn't it?

andreufont:
Hi @alxogm - should we merge this? It would be useful to get people testing v7.3.

alxogm:
@andreufont Yes, I think is ready for a final review and merge. I've tested it and seems to work ok.

PR #502, opened 2019-08-09
pixsim updates for SV and consistency with real data

sbailey:
This PR makes updates to the pixsim header keywords for consistency with real data, in preparation for Survey Validation.

  * Switches from CCD amp names 1-4 to amp names A-D (like real data)
    * Requires deishub/desispec#807 to be able to preproc these
  * Drops GAIN* keywords that ICS isn't planning to write into real data
    * $DESI_SPECTRO_CALIB tracks our gains instead in a more flexible way
  * Adds additional keywords to match real data.  Some are useful (e.g. "ST" = local sidereal time); others are amusingly mysterious (e.g. "VESSEL").
  * MJD -> MJD-OBS

Also, if pixsims MPI subcommunicators don't pack evenly into the nodes, it now prints a warning instead of raising an error.  This makes it easier to simulate a prime number of exposures without being super constrained on how many nodes you have to ask for.

Example pixsims generated with these are at NERSC in /global/cscratch1/sd/sjbailey/desi/svdc2019d/spectro/data, and have been processed with the desispec "sv" branch in PR desihub/desispec#807.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ env:
         - ASTROPY_VERSION=2.0.8
         # - SPHINX_VERSION=1.6.6
         - DESIUTIL_VERSION=1.9.15
-        - SPECLITE_VERSION=master
+        - SPECLITE_VERSION=0.8
         - SPECTER_VERSION=0.8.3
         - DESIMODEL_VERSION=0.9.6
         - DESIMODEL_DATA=branches/test-0.9.6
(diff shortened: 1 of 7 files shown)


sbailey:
Forgot to mention: the original motivation of this branch was to make the CCD amplifier offsets consistent from one exposure to another.  Previously they were a new random number for each exposure, which generated unphysical jumps that triggered QA warnings.

PR #503, opened 2019-09-14
Fixes bug in quickcat

forero:
(Open for discussion, not ready to merge yet)

This fixes the bug with wrong `NUMOBS` counts after repeated observations.
The issue was that for these lines to work
```python
        #- targets that are in both zcat and newzcat
        repeats = np.in1d(zcat['TARGETID'], newzcat['TARGETID'])

        #- update numobs in both zcat and newzcat
        ii = np.in1d(newzcat['TARGETID'], zcat['TARGETID'][repeats])
        orig_numobs = zcat['NUMOBS'][repeats].copy()
        new_numobs = newzcat['NUMOBS'][ii].copy()
        zcat['NUMOBS'][repeats] += new_numobs
        newzcat['NUMOBS'][ii] += orig_numobs
```
the `TARGETID` in `zcat` and `newzcat` should have the same ordering. 

The modification in `quickcat` fixes that bug. But it gives problems in the tests where the `TARGETID`s are expected somehow to have a different `TARGETID` ordering.

Do we have any requirement about the `TARGETID` ordering in `zcat` files?

### DIFF
--- a/py/desisim/quickcat.py
+++ b/py/desisim/quickcat.py
@@ -676,6 +676,10 @@ def quickcat(tilefiles, targets, truth, zcat=None, obsconditions=None, perfect=F
         #- efficient while still letting us modify a column if needed
         zcat = zcat.copy()
 
+        # needed to have the same ordering as newzcat
+        # and have consistent use of masks from np.in1d()
+        zcat.sort(keys='TARGETID') 
+        
         #- targets that are in both zcat and newzcat
         repeats = np.in1d(zcat['TARGETID'], newzcat['TARGETID'])
 
--- a/py/desisim/quickcat.py
+++ b/py/desisim/quickcat.py
@@ -714,7 +714,7 @@ def quickcat(tilefiles, targets, truth, zcat=None, obsconditions=None, perfect=F
     #- Metadata for header
     newzcat.meta['EXTNAME'] = 'ZCATALOG'
     
-    newzcat.sort(keys='TARGETID')
+    #newzcat.sort(keys='TARGETID')
 
     log.info('{} QC done'.format(asctime()))
     return newzcat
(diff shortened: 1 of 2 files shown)


sbailey:
We do not (or at least should not) have a requirement on the ordering of the TARGETIDs in zcat.  This test failure was making the incorrect assumption that the order is always preserved:
```
FAIL: test_quickcat (desisim.test.test_quickcat.TestQuickCat)
  File "/home/travis/build/desihub/desisim/py/desisim/test/test_quickcat.py", line 123, in test_quickcat
    self.assertTrue(np.all(zcat1['TARGETID'] == self.truth['TARGETID'][0:2*nx]))
AssertionError: False is not true
```
Please try replacing it with this (or similar):
```
self.assertEqual(set(zcat1['TARGETID']), set(self.truth['TARGETID'][0:2*nx]))
```

Please also look into this test failure, which might be genuine or it may be another mistaken assumption of the test itself:
```
FAIL: test_multiobs (desisim.test.test_quickcat.TestQuickCat)
  File "/home/travis/build/desihub/desisim/py/desisim/test/test_quickcat.py", line 169, in test_multiobs
    self.assertGreater(n1, n2)
AssertionError: 231 not greater than 241
```

Please also do a performance check (if you haven't already) that sorting a 5-year survey zcat by TARGETID doesn't take more than a few seconds (or at least not a significant fraction of the quickcat runtime).

forero:
@sbailey Tests are failing in `test_multiobs()`. It turns out that the numbers returned by that test are random (!), and the test fails randomly.

sbailey:
> @sbailey Tests are failing in `test_multiobs()`. It turns out that the numbers returned by that test are random (!), and the test fails randomly.

I'm surprised that we haven't tripped on this before.  The test logic was not obvious to me (maybe incorrect, or maybe accidentally correct by relying on some underlying assumption that I was missing).  Regardless, I replaced the test with clearer (to me) logic.

That highlighted another issue though: it is testing that targets with multiple observations have a higher efficiency than those with single observations.  Although the *probability* is higher, once random numbers are thrown to realize those probabilities, sometimes fluctuations cause them to be lower.  I increased the number of targets simulated in the test and now I don't see failures in 20 runs.

forero:
Thanks @sbailey! Is this ready to merge?

PR #505, opened 2019-10-01
Pin numpy version

weaverba137:
After creating tag 0.33.0, tests were running with Numpy/1.17 and failing.  This sets the version to 1.16.

@sbailey, there are some other versions in the .travis.yml file that could use some freshening up.  Do you want to go ahead with that?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ env:
         # The following versions are the 'default' for tests, unless
         # overidden underneath. They are defined here in order to save having
         # to repeat them for all configurations.
-        # - NUMPY_VERSION=1.10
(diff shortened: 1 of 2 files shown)


sbailey:
Thanks.  Regarding updating other versions:

For external packages, there are advantages and disadvantages to testing against "oldest known to work" vs. "latest available" so I don't think it is worth trying to keep up in either direction, except for cases like this where we specifically need to pin the version.

For our internal packages, we specifically want to keep the Travis tests for those fairly old to establish the oldest version that is known to work, while the nightly cronjob tests at NERSC are testing the latest version of everything against each other, providing a nice compliment for the Travis vs. NERSC tests.

PR #508, opened 2019-10-05
support mocks in dark/bright subdirs

sbailey:
This PR updates desisim newexp-mock to support the updates in desihub/desitarget#543 to split the mocks into separate dark/bright directories, just like the target and MTL files for real data are split.

In particular it uses the PROGRAM of a tile to propagate `obscon` to `get_mock_spectra` which passes it to `desitarget.mock.io.findfile`.  The mapping of PROGRAM to obscon is a bit fragile.

This has been tested with the minitest notebook (along with desihub/desitarget#543 and desihub/fiberassign#223).

### DIFF
--- a/py/desisim/scripts/newexp_mock.py
+++ b/py/desisim/scripts/newexp_mock.py
@@ -77,9 +77,18 @@ def main(args=None):
             args.nspec, len(fiberassign)))
         sys.exit(1)
 
-    log.info('Simulating night {} expid {} tile {}'.format(night, args.expid, tileid))
+    log.info('Simulating night {} expid {} tile {} {}'.format(
+        night, args.expid, tileid, program))
+
+    if program.lower() in ('bright', 'sv_bgs', 'sv_mws'):
+        mock_obscon = 'bright'
+    else:
+        mock_obscon = 'dark'  #- includes gray for mocks
+
     try:
-        flux, wave, meta, objmeta = get_mock_spectra(fiberassign, mockdir=args.mockdir, nside=args.nside)
+        flux, wave, meta, objmeta = get_mock_spectra(
+                fiberassign, mockdir=args.mockdir,
+                nside=args.nside, obscon=mock_obscon)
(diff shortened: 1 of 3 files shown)


py/desisim/scripts/newexp_mock.py
```diff
@@ -77,9 +77,18 @@ def main(args=None):
             args.nspec, len(fiberassign)))
         sys.exit(1)
 
-    log.info('Simulating night {} expid {} tile {}'.format(night, args.expid, tileid))
+    log.info('Simulating night {} expid {} tile {} {}'.format(
+        night, args.expid, tileid, program))
+
+    if program.lower() in ('bright', 'sv_bgs', 'sv_mws'):
+        mock_obscon = 'bright'
```
moustakas: Will this list of programs ever change? In other words, can we glean this from some other master file or method in `desitarget`?

moustakas:
I only reviewed---but didn't test---the changes, and they look OK.

py/desisim/scripts/newexp_mock.py
sbailey: I'm sure this list will change, and this is a fragile bit of code to maintain as-is.  At the same time, I'm not aware of us currently having a canonical location to derive which programs go with which obsconditions and how those map to "dark" and "bright" and possibly others (e.g. "gray" obsconditions get combined with the "dark" target + MTL files...)

I think the right solution is to record this at the level of the original target file and then propagate through MTL and fiberassign, so at the point that this code is picking up the fiberassign file it has traceability back to which dark/bright/other subdir the original targets came from.

But for right now, I'm hesitant to embark on another 3-repo coordinated update, vs. a hack here to move on for now.  This is partially motivated / justified by this newexp_mock code only applying to sims, and being a relatively infrequently used piece of code (i.e. used during integration tests and major data challenges, but not day-to-day activities of the science working groups).

I'll leave this open for another day for discussion.  Thoughts?

PR #512, opened 2019-10-17
Fiberassign xy

sbailey:
This PR is a companion to desihub/desispec#821 to handle the renaming of DESIGN_X/Y -> FIBERASSIGN_X/Y.  I'm still testing at NERSC with the minitest notebook, but I'm opening this PR now to get Travis testing going too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ env:
         # - NUMPY_VERSION='<1.17'
         - NUMPY_VERSION=1.16.5
         # - SCIPY_VERSION=0.16
-        - ASTROPY_VERSION=2.0.15
+        - ASTROPY_VERSION=lts
         # - SPHINX_VERSION=1.6.6
         - DESIUTIL_VERSION=2.0.1
         - SPECLITE_VERSION=0.8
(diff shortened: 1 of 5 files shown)


weaverba137:
Interesting!  astropy/ci-helpers has recently added some code that tries to interpret version numbers, and it may be choking on the NUMPY_VERSION.

weaverba137:
See astropy/ci-helpers#422

weaverba137:
I think the problem is fixed, but I'm now seeing network errors during the install phase.  Stay tuned.

weaverba137:
O boy, what a mess, it looks like on Python 3.5, pip is uninstalling numpy, leaving no valid numpy version.  On Python 3.6, tests are failing due to genuine key errors, not install problems.

weaverba137:
I've found a numpy version that works for Python 3.5, but it's time to think about abandoning 3.5 altogether.  All test failures are now due to genuine key errors.  Over to you @sbailey.

sbailey:
Thanks for working on this.  The FIBERASSIGN_X key error when testing under 3.6 is because it needs the changes in desihub/desispec#821.  If you can get the python 3.5 tests back into a state of failing on genuine test errors and not on pip/numpy/etc installation problems, I'll run with the rest.

For the record, tests pass on my laptop using desisim/fiberassign_xy + desispec/fibermap_xy + master of everything else, and python 3.6.7 + numpy 1.16.4.  At NERSC tests pass with the same versions of desi code + python 3.6.6 + numpy 1.16.4.

sbailey:
Thanks!  The next time 3.5 becomes problematic, let's drop it, while still trying to retain support for python 3.6 through 3.8 (which just came out).

I tagged desispec/0.30.0 (long overdue) and am testing this branch against that now in prep for merging and a new desisim tag.

PR #514, opened 2019-11-07
use find_target_files not mockio.findfile

moustakas:
Companion PR to https://github.com/desihub/desitarget/pull/561.  

I *think* this is the only change needed to not break the integration tests, but I personally can't test the changes on my laptop because of the issue raised here: https://github.com/desihub/desisurvey/issues/104.


moustakas:
@sbailey I'm trying to test the changes in https://github.com/desihub/desitarget/pull/563 but I'm getting a much wider range of `desisim` unit test failures than I was expecting.  I don't want to update `.travis.yml` too much more before checking in with you.  Could you take a look at the failures if you get a chance?

Oh and I messed up and accidentally made two commits to master rather than to this branch. I can revert them but they should be mostly harmless and will be consistent with this branch once we merge this PR.  Please advise.

moustakas:
@sbailey not all tests are passing but many of the changes here are going to be needed for `19.12` integration tests.

PR #520, opened 2019-12-23
[WIP] Add freeze_iers

weaverba137:
This PR adds `freeze_iers()` calls where needed to prevent downloads of the IERS-A table.  At this point, I have updated pixsim.py, such that it can run to completion during integration tests, but the logging output from the integration test indicates that there are other modules or executables that are attempting the download, possibly `newexp-random`.

Do not merge until the additional download attempts are isolated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,7 @@ env:
         # - DESIMODEL_DATA=branches/test-0.9.6
         - DESISIM_TESTDATA_VERSION=master
         - SPECSIM_VERSION=v0.12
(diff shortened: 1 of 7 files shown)


weaverba137:
The rule seems to be: call `freeze_iers()` before *any* `import astropy.time` or `from astropy.time import ...`.

weaverba137:
Imports of `astropy.coordinates` *might* also trigger downloads of IERS.

weaverba137:
Merging to get this into a new tag.

Issue #532, opened 2020-04-27
Change default pixel width of DESI mocks to 0.8A


andreufont:
We recently found out that the Lyman alpha mocks had pixels of width 1A, instead of 0.8A.

We think this might be related to line 619-620 in simexp.py, that sets by default the output pixel width to 1A, when the user doesn't provide any value for dwave_out.

https://github.com/desihub/desisim/blob/3ae76e4c921f72b71ff7522462740e904136f428/py/desisim/simexp.py#L620

How confident are we that the final pixel width will be 0.8A? If we are confident, should we change this default value?

sbailey:
I wouldn't be surprised if 0.8 changed again, but we might as well be consistent in the meantime (including the exact phasing of the 0.8 grid).  I've added this to the 20.5 project to keep it on our radar for the next release.  Contributions welcome.  See [desispec/bin/desi_proc lines 680-685](https://github.com/desihub/desispec/blob/8e7ebe22ec6e1987860f1b66dd90953e11853984/bin/desi_proc#L680-L685) for current defaults (and yes, it is unfortunately hardcoded in a script; consider cleaning that up while implementing this so that desisim automatically uses whatever desispec is using as default).

alxogm:
coming back to this issue. We fixed this for quickquasars mocks in  desihub/desisim#533 . It seemed everything was right, but the balfinder is crashing for this mocks due to the following error 

```
ERROR:coaddition.py:230:coadd_cameras: Cannot directly coadd the camera spectra because wavelength are not aligned, use --lin-step or --log10-step to resample to a common grid
```
It works well for a pixel width of 1 and 0.6, but not 0.8... 
The error is rised in [#L225-L231](https://github.com/desihub/desispec/blob/1d78b9cf56ddb5dfc9c49cf0edd7f8d3f9804b8a/py/desispec/coaddition.py#L225-L231) in `desispec `, and particularly fails for the 'z' band.  I can,  indeed, fix it by resampling the spectra with a linear step before making the coaddition, but I wanted to be sure this isn't really an issue in `desispec` or` desisim`, at the time of creating the wavelength array... In the SV data we don't see this failure because the spectra is already coadded...
 @sbailey do you know if  this is an issue maybe related to the more general change that should be done to use the pixel width in mocks in general,  or is it somehow expected?

sbailey:
[...]

PR #533, opened 2020-05-04
Add dwave out to quickquasars

alxogm:
Their PR is to allow quickquasars to produce mocks with pixels width specified by the user. This is related to the more general issue  raised by Andreu https://github.com/desihub/desisim/issues/532. 

Change is very simple, I've tested it in a small run here /global/cfs/projectdirs/desi/mocks/lya_forest/develop/london/qq_desi/v9.0/v9.0.0/desi-1.2-1.tmp/ 

I was waiting for some feedback from the DLA team, but I think it will be fine if we just go ahead. 

**Other changes are only some missing spaces I added ...

### DIFF
--- a/py/desisim/scripts/quickquasars.py
+++ b/py/desisim/scripts/quickquasars.py
@@ -104,9 +104,7 @@ def parse(options=None):
 
     parser.add_argument('--mags', action = "store_true", help="DEPRECATED; use --bbflux")
 
-    parser.add_argument('--bbflux', action="store_true", dest='bbflux', default=True, help="compute and write the QSO broad-band fluxes in the fibermap ")
-
-    parser.add_argument('--no-bbflux', action="store_false", dest='bbflux', help="don't compute and write the QSO broad-band fluxes in the fibermap ")
+    parser.add_argument('--bbflux', action = "store_true", help="compute and write the QSO broad-band fluxes in the fibermap")
 
     parser.add_argument('--add-LYB', action='store_true', help = "Add LYB absorption from transmision file")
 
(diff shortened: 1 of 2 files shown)


py/desisim/scripts/quickquasars.py
```diff
@@ -299,6 +301,11 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
             log.info("Creating dir {}".format(pixdir))
             os.makedirs(pixdir)
 
+    if not eboss is None:
+        dwave_out=1.   #to keep the same value used for DR16
```
andreufont, line 305: @alxogm - I'm worried that eBOSS users might think that dwave_out will also affect them, and here we hide its effect. We could rename the adgument to dwave_DESI, or at least edit the help string to make it clear it only affects DESI. May be add also a print warning here if you were trying to run with eBOSS and also trying to change the value of dwave_out?

andreufont: Actually, the eBOSS version should really just ignore dwave_out, since the SDSS spectrograph doesn't have constant pixels in wavelength, but logarithmic. 

I'd suggest to set dwave_out=None when running with eBOSS flag.

py/desisim/scripts/quickquasars.py
```diff
@@ -102,7 +104,7 @@ def parse(options=None):
 
     parser.add_argument('--mags', action = "store_true", help="DEPRECATED; use --bbflux")
 
-    parser.add_argument('--bbflux', action = "store_true", help="compute and write the QSO broad-band fluxes in the fibermap")
+    parser.add_argument('--bbflux', type=bool, default=True, help="compute and write the QSO broad-band fluxes in the fibermap ")
```
alxogm, line 107: Ups! I think I accidentally pushed an unrelated change... should I included in this PR or left it aside?

andreufont: I'm ok putting it all together, it is a small fix.

py/desisim/scripts/quickquasars.py
alxogm: Hi @andreufont, right! 

dwave_out=None was my first option, but I changed to 1 because I was not considering the there is an eBOSS branch that is not yet merged, so I didn't see what you mention about eBOSS not having constant pixels. So I've set it back to dwave_out=None, and changed the argument name also, to avoid confusion...

alxogm:
@andreufont I made a few more changes to handle the --bblux as default, previous syntax was not correct... I don't see when we will need it to set it as False but I better left the option anyway.  I included an update of the warning message of a deprecated argument, just in case... I'll merge if there are no more changes. Thanks

alxogm:
We decided to do a clean up of arguments in a separate PR. Then I'll merge once tests passes.

weaverba137:
Can the branch `add_dwave_out` be deleted?

PR #538, opened 2020-07-09
new keyword in quickcat to indicate which HDU to read in fba files

forero:
This simple fix will allow reading either `fba-` or `fiberassign-` files to build a redshift catalog. The old hardcoded option only works for `fiberassign-` files.


weaverba137:
Can the branch `quickcat_fix` be deleted?

PR #545, opened 2020-11-26
Fix bug

alxogm:
It seems I didn't made the PR for this before. Its a simple change. The issue was that it was failing if there were no BALs in a specific pixel. Also I have checked now that it will work on select_mock_targets, I'll push a PR there after this one.


moustakas:
Not explicitly tested, but this looks OK to me.

alxogm:
> Not explicitly tested, but this looks OK to me.

Thanks @moustakas I'll merge now.

Issue #534, opened 2020-05-07
desisim sky level test failure


moustakas:
Opening a ticket for an issue sent by @sbailey on-list:

```
Something late last week changed that started breaking this test:

======================================================================
FAIL: test_newexp_sky (desisim.test.test_obs.TestObs)
Test different levels of sky brightness
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/global/common/software/desi/cori/desiconda/20190804-1.3.0-spec/code/desisim/master/py/desisim/test/test_obs.py", line 127, in test_newexp_sky
    self.assertTrue(np.all(sky_mws[nonzero] > sky_dark[nonzero]))
AssertionError: False is not true

Does anyone recognize what might have caused that?  Running "python py/desisim/test/test_obs.py" 
can confirm that without having to run all the other tests.

    it fails when using master of everything
    it fails when using master of everything except desisim/0.34.2 (used in the 19.12 software release)
    it passes when using the full suite of tags for 19.12

Adding to the mystery, there seems to be a clear bug in test_obs.py lines 119-122:

        sim_dark, fmap_dark, meta_dark, obscond_dark, objmeta_dark = obs.new_exposure('dark', nspec=10, night=night, expid=0, exptime=1000)
        dark = sim_dark.simulated.copy()
        sim_mws, fmap_mws, meta_mws, obscond_mws, objmeta_mws = obs.new_exposure('mws', nspec=10, night=night, expid=1, exptime=1000)
        mws = sim_dark.simulated.copy()

Note that "mws" is copying from sim_dark instead of sim_mws, so I don't see how the 
later test that the mws sky brightness > the dark sky brightness could have passed, 
unless they were all zeros which also seems unlikely.



```

moustakas:
The failure is coming from the fact that the simulated sky spectra are *negative* (in units of electrons).

<img width="571" alt="Screen Shot 2020-05-07 at 8 28 27 AM" src="https://user-images.githubusercontent.com/1431820/81294433-e5704880-903c-11ea-9282-1866bb4c6f31.png">

moustakas:
I tried to track this issue down into `specsim` but wasn't successful. As far as I can tell, the sky spectra are negative when they leave `simexp.simulate_spectra` but I can't quite track down where in `specsim.simulator.Simulator.simulate` this is occurring. 

[...]

Issue #507, opened 2019-10-01
Suspiciously low QSO flux


weaverba137:
Independently of the Numpy 1.17 issue, we are also seeing this error on some but not all tests:

```
======================================================================

FAIL: test_newexp (desisim.test.test_obs.TestObs)

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/build/desihub/desisim/py/desisim/test/test_obs.py", line 103, in test_newexp

    self.assertTrue(maxflux > 0.01, 'suspiciously low {} flux ({}) using seed {}; wrong units?'.format(objtype, maxflux, seed))

AssertionError: False is not true : suspiciously low QSO flux (0.0) using seed 800364; wrong units?

----------------------------------------------------------------------
```

It appears that some seeds are just "bad luck" and trigger this bug.

moustakas:
Using the seed from [this test failure](https://travis-ci.org/github/desihub/desisim/jobs/685028042), I'm able to reproduce the problem:
```
import numpy as np
from desisim.templates import QSO
qq = QSO()
ff, ww, mm, oo = qq.make_templates(1, seed=458870270)
WARNING:templates.py:2258:make_templates: Maximum number of iterations reached on QSO 0, z=1.47216.
WARNING:templates.py:2264:make_templates: 1 spectra could not be computed given the input priors!
np.all(ff==0)
  True
```

The reason is clearly stated by the code: with that seed, there is no non-negative combination of the PCA templates (at the randomly-generated redshift, which is why this failure is intermittent) which produce a spectrum which passes the QSO color-cuts. Increasing the maximum number of iterations to a number as high as 500 (the default is 20), which takes considerably longer, still returns an all-zero spectrum. So that's not a viable solution. One option would be that if this error condition is reached, the code could turn off color-cuts and throw up a non-catastrophic warning. And indeed, this works:
```
ff, ww, mm, oo = qq.make_templates(1, seed=458870270, nocolorcuts=True)
```

OK, I found the problem. The issue is that QSOs are required to have `17.5<r<22.7`--
  https://github.com/desihub/desitarget/blob/master/py/desitarget/cuts.py#L1225-L1226

However, the default magnitude prior range in `templates.QSO.make_templates` is `17<r<22.7`, and with this seed the randomly generated magnitude is `r=17.15`, hence no viable template can be generated because the template normalized by this magnitude.

[...]

PR #546, opened 2021-02-02
Migrate desisim to GitHub Actions

weaverba137:
This PR migrates desisim to use GitHub Actions instead of Travis CI for unit tests.

While testing, a version of #507 reappeared, although this time it is complaining of suspiciously low *ELG* or *LRG* flux.

See [this run](https://github.com/desihub/desisim/actions/runs/529377843) for full details.

Oddly though, #534 did not reappear.  Perhaps that was a red-herring and desimodel/branches/test-0.12 does have the fix installed.


sbailey:
FYI #547 documents the pixsim test failure and likely cause (a change in specter, but requiring a non-negative input flux cut in desisim).  We could proceed with this PR even if that test is still failing.

The "suspiciously low XYZ flux" problem appears occasionally in the nightly tests and previously travis tests; I usually ignore and rerun.

weaverba137:
OK, the tests were successful after a restart, but I still think the low flux problem should actually be fixed, not ignored.  In particular, experts know to restart the tests, but newcomers would have to be aware that this is a known issue.

weaverba137:
PS, I still have a few minor changes to make to this PR, like removing the `.travis.yml` file, so please don't merge immediately.

sbailey (approved):
This PR looks good, thanks.  Merge when ready.  I agree that the low flux bug should be fixed, but it doesn't need to be a blocking factor for this travis -> github actions transition.

weaverba137:
@sbailey, thank you merging very soon.

weaverba137:
Test failures are related to #507, ignoring.

PR #549, opened 2021-02-04
suspiciously low ELG/LRG/QSO flux

sbailey:
This is a PR to try to study the suspiciously low ELG/LRG/QSO flux problem (#507 et al), which suspiciously seems to be happening more frequently nowadays.


weaverba137:
@sbailey, the failing tests do record a random seed. What if you deliberately set that seed, so you can at least see what values are being generated by that seed?

sbailey:
@moustakas could you take a look at this "suspiciously low ... flux" false alarm again, using this PR as the sandbox if needed?  The original point of this test was to catch units interpretation errors for whether to include a 1e-17 term or not (we tripped on that multiple times years ago), but it is raising a false alarm so often that it is problematic for unrelated pull requests.

Critically: I tried hardcoding the random seed reported by the test and rerunning at NERSC and it *didn't* fail, but I haven't dug deeper for whether we have a bug in using the seed or whether we have a portability issue (e.g. different versions of numpy resulting in different sequences for the same seed).

moustakas:
Superseded by #559; closing without merging.

PR #551, opened 2021-02-15
add missing header keywords

sbailey:
This PR adds some header keywords that were missing from simulated output.  New template S/N ratio (TSNR) code in the spectro pipeline now requires these keywords, which was breaking the integration testing.  Hardcoding the CCDSIZE is admittedly yucky, but I didn't see how to access that information from the params available to fastframe at that point in the code.  I'm going to proceed with this so that I can proceed with integration testing, but if someone cares deeply enough they can separately contribute a cleaner solution later.


sbailey:
Failing test is suspiciously low LRG flux issue; proceeding without rerunning.

PR #566, opened 2022-02-22
Smoothing source contribution to noise in quickquasars DESI mocks

p-slash:
There is now an option in quickquasars to smooth out the source contribution to noise for DESI mocks in order to de-bias the weights in Lya forest analyses. The value passed will the Gaussian sigma for the smoothing in Angstrom. This smoothing is implemented with FFTs on a padded array in quickspectra.py.

### DIFF
--- a/py/desisim/scripts/quickspectra.py
+++ b/py/desisim/scripts/quickspectra.py
@@ -22,9 +22,47 @@ import desitarget
 from desispec.spectra import Spectra
 from desispec.resolution import Resolution
 
+def _fft_gaussian_smooth(array, sigmapix, pad_size=10):
+    iwavesize, nspec = array.shape
+    # Pad the input array to get rid of annoying edge effects
+    # Pad values are set to the edge value
+    arrsize    = iwavesize+2*pad_size
+    padded_arr = np.empty((arrsize, nspec))
+    padded_arr[:pad_size, :] = array[0, :]
+    padded_arr[iwavesize+pad_size:, :] = array[-1, :]
+    padded_arr[pad_size:iwavesize+pad_size, :] = array
+
+    kvals        = np.fft.rfftfreq(arrsize)
+    kernel_k     = np.exp(-(kvals*sigmapix)**2/2.)
+    snumsource_k = np.fft.rfft(padded_arr, axis=0)*kernel_k[:, None]
+
+    return np.fft.irfft(snumsource_k, n=arrsize, axis=0)[pad_size:-pad_size]
+
+# camera_output seems to change without assignment
+# Assignment yields attribute error
+# assumes dwave_out is not None
+def _smooth_source_variance(camera_output, sigma_A, dwave_out):
+    # arm_output shape is (wave.size, nspec)
+    for i in range(3):
+        arm_output = camera_output[i]
+
+        # num_source_electrons goes into poisson noise
+        # Remove it from the variance first
+        arm_output['variance_electrons'] -= arm_output['num_source_electrons']
+
+        sigmapix = sigma_A/dwave_out
+        arm_output['num_source_electrons'] = _fft_gaussian_smooth(arm_output['num_source_electrons'], sigmapix)
+
+        # add smoothed source electrons back to variance
+        arm_output['variance_electrons'] += arm_output['num_source_electrons']
+
+        arm_output['flux_inverse_variance'] = (
+            arm_output['flux_calibration'] ** -2 *
+            arm_output['variance_electrons'] ** -1)
+
 def sim_spectra(wave, flux, program, spectra_filename, obsconditions=None,
                 sourcetype=None, targetid=None, redshift=None, expid=0, seed=0, skyerr=0.0, ra=None,
-                dec=None, meta=None, fibermap_columns=None, fullsim=False, use_poisson=True, specsim_config_file="desi", dwave_out=None, save_resolution=True):
+                dec=None, meta=None, fibermap_columns=None, fullsim=False, use_poisson=True, specsim_config_file="desi", dwave_out=None, save_resolution=True, source_contribution_smoothing=0):
     """
     Simulate spectra from an input set of wavelength and flux and writes a FITS file in the Spectra format that can
     be used as input to the redshift fitter.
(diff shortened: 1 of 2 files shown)


py/desisim/scripts/quickspectra.py
```diff
@@ -186,6 +226,10 @@ def sim_spectra(wave, flux, program, spectra_filename, obsconditions=None,
         obsconditions=obsconditions, redshift=redshift, seed=seed,
         psfconvolve=True, specsim_config_file=specsim_config_file, dwave_out=dwave_out)
 
+    # Smoothing source electron numbers only works for DESI mocks
+    if specsim_config_file != "eboss" and source_contribution_smoothing > 0:
+        _smooth_source_variance(sim.camera_output, source_contribution_smoothing, dwave_out)
+
     random_state = np.random.RandomState(seed)
     sim.generate_random_noise(random_state,use_poisson=use_poisson)
```
andreufont, line 234: @julienguy , @p-slash , @alxogm - shouldn't we hack the estimate of the variance *after* we have generated the random noise?

py/desisim/scripts/quickspectra.py
```diff
@@ -22,9 +22,47 @@
...
+    padded_arr = np.empty((arrsize, nspec))
+    padded_arr[:pad_size, :] = array[0, :]
+    padded_arr[iwavesize+pad_size:, :] = array[-1, :]
+    padded_arr[pad_size:iwavesize+pad_size, :] = array
+
+    kvals        = np.fft.rfftfreq(arrsize)
+    kernel_k     = np.exp(-(kvals*sigmapix)**2/2.)
+    snumsource_k = np.fft.rfft(padded_arr, axis=0)*kernel_k[:, None]
+
+    return np.fft.irfft(snumsource_k, n=arrsize, axis=0)[pad_size:-pad_size]
```
andreufont, line 39: I haven't checked the pythonology of arrays and FFT conventions here, but I guess you have checked this.

andreufont (commented):
A part from my comment on when to smooth the noise variance, I think it looks good. I would like to see a comparison of reported values of IVAR from different runs of the same spectrum, with different values of smoothing, just to check that indeed the reported values are smoothed out. 

As I mentioned, I think the *flux* values should be exactly the same, including the noise realisation, and that only the reported values of noise variance should have changed.

p-slash:
Sorry, this is the first time I'm seeing this. This is a subtle point. If we were to have the exact flux and noise, but different reported variance, that would be wrong. That is just the wrong noise estimate for that realization. 

The point is to decouple the generated noise from the source flux, so they will be different for different smoothing scales. In this implementation, we smooth the variance first (only the source contribution) and generate noise from this smoothed variance. This way, the generated noise is decoupled from the source flux, _and_ we still know the exact noise that is put in. Furthermore, using these reported variances as inverse weights will be unbiased (less biased) in statistics estimations.
If we were to smooth out the generated noise, then it is hard to estimate the underlying variance.

In terms of P1D and Xi3D estimation, I think we should still smooth the noise in data and mocks, but only in its _contribution to weights_. Noise power subtraction should use the exact reported variance to correctly remove the noise power. The difference is that here we are smoothing out all contributions to noise, whereas the implementation in quickquasars smoothes out only the source contribution.

andreufont:
@julienguy , could you comment on this? I thought that what I suggested above is what the real pipeline does.

andreufont:
@julienguy , @p-slash - we forgot to discuss this at the telecon yesterday. Should we have a quick call to make sure we all agree on this? Do you prefer to discuss this online?

andreufont:
Also, @p-slash - can you move the branch to the desisim repo, instead of your own fork? César Ramírez would look at it in the context of BAO analyses.

p-slash:
> 

Not sure how. I can try. Can't he pull from my branch to his local repo? 

I would be fine with a chat or over text on Slack. I don't check here as often:)

py/desisim/scripts/quickspectra.py
julienguy: Yes. We should change the variance model after generating the noise.

py/desisim/scripts/quickspectra.py
p-slash: I can switch to fftconvolve. Any tricks for edge effects?
Why is this implementation not normalized though? kernel_k[0]=1, so sum of kernel_x=1. That's what I thought at least.

julienguy: (my bad, I was thinking about a kernel in configuration space, as it is the case in the argument of fftconvolve)

andreufont:
@p-slash , @julienguy - any update on this discussion?

julienguy:
I commented we should change the variance model after generating the noise. Naim @p-slash I can make the change if you are not sure how to do this.

p-slash:
I moved the function after sim.generate_random_noise(random_state,use_poisson=use_poisson). That should do the trick. I HAVE NOT tested it though. Sorry I've been busy with other things.

andreufont:
@p-slash , @julienguy , @alxogm - what is the status of this PR?

p-slash:
Thanks Alma

alxogm:
Hi @p-slash 
I have tested the branch and works well it gives the same flux and a smoothed version of the same noise for a given spectra . I was planning to run the deltas and CF but given the presentation today by Andrei, I think you can go ahead and merge. 

I have only requested to add one line to run log to inform the smoothing was applied.

p-slash:
Given our discussion at the workshop, should we change the default for `--source-contr-smoothing` to `10 A`? This would mean behavior change for quickquasars moving forward.
Also, @alxogm not sure where to add the line for logging.

py/desisim/scripts/quickspectra.py
```diff
@@ -189,6 +229,10 @@ def sim_spectra(wave, flux, program, spectra_filename, obsconditions=None,
     random_state = np.random.RandomState(seed)
     sim.generate_random_noise(random_state,use_poisson=use_poisson)
 
+    # Smoothing source electron numbers only works for DESI mocks
+    if specsim_config_file != "eboss" and source_contribution_smoothing > 0:
```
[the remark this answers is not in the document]
alxogm: @p-slash I suggest here you can add a line for the log

alxogm (changes requested):

PR #569, opened 2022-10-05
Updates in quickquasars to reproduce actual redshift and magnitude distribution as obtained in SV for DESIY5 mocks.

HiramHerrera:
This PR adds an option to quickquasars `--dn_dzdm` which expects a 2D table with number of quasars by square degree by redshift and r-band magnitude bins.  A default file was added: `$DESISIM/py/desisim/data/dn_dzdM_EDR.fits` which contains the expected distribution for Y5 based on SV results.

The procedure of this option goes as follows:

1. Downsampling of quasars by each redshift bin to the number in the table given marginalized over magnitudes.
2. Random assignation of r-band magnitude following a PDF obtained from dividing the number of quasars for each magnitude bin for a given redshift bin divided by the total quasars in such redshift bin.

This PR also modifies the option `--extinction` by moving its position in the `quickquasars` code, and by adding an option that rescales flux in order to mantain the magnitude distribution from the previous file.


py/desisim/scripts/quickquasars.py
```diff
@@ -372,8 +374,39 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
...
+        selection_z=[] #Initialize array to select qsos
+        for z_bin, dndz_bin in zip(zcenters,dn_dz):
+            nqso_bin=np.ceil(pixarea*dndz_bin).astype(int)
+            w_z = (z>=z_bin-dz)&(z<z_bin+dz)
+            nqso_orig = len(z[w_z])
+            if nqso_orig==0:
+                continue # If no QSOs in that bin, skip
+            idx = np.where(w_z)[0]
+            downsampling_bin = nqso_bin/nqso_orig
+            if downsampling_bin<1:
```
andreufont, line 400: We might want to add a warning, or even crash, if the required downsampling is larger than 1.

py/desisim/scripts/quickquasars.py
```diff
@@ -618,6 +678,14 @@ def simulate_one_healpix(ifilename,args,model,obsconditions,decam_and_wise_filte
...
         tmp_qso_flux = apply_metals_transmission(tmp_qso_wave,tmp_qso_flux,
                             trans_wave,transmission,args.metals)
+    # Attenuate the spectra for extinction
+    if not sfdmap is None:
+        Rv=3.1   #set by default
+        indx=np.arange(metadata['RA'].size)
+        extinction =Rv*ext_odonnell(tmp_qso_wave)
+        EBV = sfdmap.ebv(metadata['RA'],metadata['DEC'], scaling=1.0)
+        tmp_qso_flux *=10**( -0.4 * EBV[indx, np.newaxis] * extinction)
+        log.info("Dust extinction added")
```
andreufont, line 688: I don't know the details about this extinsion corrections, but I guess you have cross-checked this works?

andreufont (changes requested):
@HiramHerrera - thanks for putting together this PR. Could you also add a plot or two of the luminosity function used? It might be useful to have this together with the PR. I only asked for one change, and it is to add a warning or an error when the downsampling is requested to be larger than 1, we do not want this to silently happen without us noticing.

andreufont:
There were a couple of code blocks that I didn't inspect carefully, since they use functions that I'm not familiar with. Have others reviewed this? Are these new mocks much different than the previous once, in terms of quasar luminosities?

andreufont (approved):

alxogm:
The test that is failing is unrelated with this PR and desicode, it seems to be related with Hiram making the PR from within a fork. @sbailey are you ok with us merging this branch with this one test failing?

sbailey:
@alxogm yes, that is fine.  I'll let you do the actual merge when you are ready.

PR #574, opened 2022-11-10
- Change defaults for some options 

alxogm:
Changed the defaults for options 
--zbest, --bbflux and --desi_footprint to be true as defult. Now in order to set them false we have to give --no-zbest, --no-bbflux, and no-desifootprint. 
@andreufont @HiramHerrera @andreicuceu I think those are the only arguments that make sense to change for now... let me know if there are other suggestions. 

I also updated a bit the notebook about reproducibility to work with the latest version of the mocks. I simplified the notebook and included some information about the truth files. This can be improved in a future PR is we want to...


andreufont:
Not sure why the tests are failing, but I agree with the changes in the defaults (although we could also delete the options from argparser).

alxogm:
It fails because the functionality is only available in python 3.9 and the tests are in 3.8. 

@sbailey is there any reason to keep the tests with python 3.8? I see other packages like desispec are already in 3.9...

andreufont:
@alxogm - is this still a relevant PR? Was it addressed elsewhere?

andreufont (commented):
This PR might be obsolete after the changes to QQ that Hiram and I are working on in his fork of desisim. 

We'll report on this soon.

PR #578, opened 2023-11-14
Producing DESI-Y1 Lya mocks with quickquasars.

HiramHerrera:
This PR introduces the necessary tools to produce DESI-Y1 Lya mocks.

Here are the Major changes listed: 

- Introduction of the [survey_release.py](py/desisim/survey_release.py) module and the SurveyRelease class instance with the necessary tools to create mock catalogs to be read by `quickquasars`. 
    - Included the [generate_survey_release_catalogs.ipynb](doc/nb/generate_survey_release_catalogs.ipynb) noteboook with a brief tutorial on making these catalogs.
- Introduced new flags to  `quickquasars`.
    -   `--raw-mock [saclay, london, ohio]` allows to generate Y5 mocks directly from quickquasars for the specified raw mock team.
    -  `--dn_dzdm` was reworked to boolean flag, used when redshift and magnitude distributions are to be downsampled by `quickquasars`directly.
    - `--year1-throughput` adds the Y1 throughput model to mock spectra including the 440 nm dip observed during Y1
    - `--from-catalog` allows to read from an input pre-processed mock catalog generated by the SurveyRelease instance.
    - `--metal-strengths` allows to change the default metal strengths used for the `--metals` flag. This flag should be the same length and in the same order as the `--metals` flag in order to work. The `lya_spectra.py` script was modified to make this flag work.

An example quickquasars command to run DESI-Y1 mocks would be:

```
quickquasars -i <input_path> --outdir <output_path> --zbest  --bbflux  --save-continuum --zmin 1.7 --dla file --balprob 0.16 --metals LYB LY3 LY4 LY5 SiII\(1260\) SiIII\(1207\) SiII\(1193\) SiII\(1190\) --metal-strengths 0.1901 0.0697 0.0335 0.0187 1.3e-03 3.5e-03 0.7e-03 1.4e-03 --from-catalog <input_preprocessed_y1catalog> --seed <seed>
```

**Important notes:**

1. The `--year1-throughput` flag works if a file called `desiY1.yaml` file exists within `specsim/data/config` pointing to the `specpsf/psf-quicksim.fits` and files `thru-[brz]_y1measured.fits` which include the dip feature at 440 nm. We can take advantage of  https://github.com/desihub/specsim/pull/125 to include this file. However, if I'm correct, that PR changes the defaults of the `desi.yaml` file to those of DESI-Y1. I think it is important to keep the `desi.yaml` pointing to a throughput model without the 440 nm dip (default DESIMODEL) as we might want mocks without this feature.
[...]

PR #580, opened 2024-03-22
Pending changes requested on previous PR, from_catalog fix for Ohio mocks

HiramHerrera:
This Pull request addresses various of the requested changes that were pending in #578. Most of the comments were addressed except for one:
> Ah this is the seed. I recommend creating a member random engine `rng_engine = np.random.default_rng(seed)` and calling it to generate the random numbers you need. `default_rng` is recommended since numpy 1.17

I tested doing this and the resulting catalogs are different than the obtained with the currently implemented method. For the sake of reproducibility of the KP6 mocks I will leave it as is. Changing to a random engine will be left for future implementations.

Some additional changes in this PR:
1. Fixed a bug causing `quickquasars` to assign the magnitudes and exposure times from the input catalog to the incorrect MOCKID. This happens if the sorting order of the MOCKIDs in the input catalog is different to the sorting order in the metadata of the raw transmission files. I checked that this does not affect neither the Y1 Saclay or LyaCoLoRe Mocks for two main reasons. 
     * In both cases the MOCKID are sorted the same way in the input catalog and metadata. Therefore, the magnitudes and exposure times were correctly assigned. 
     * For LyaCoLoRe and Saclay we don't care much about what target gets assigned which magnitude, since these are random and only depend on the target's redshift. This is more relevant in Ohio mocks where we want specific targets to have specific magnitudes and exposure times for direct comparison with observed data.
2. I made it possible to use the `gen_qso_catalog.py` script to generate an input catalog for Y5 mocks that will follow the redshift-magnitude distribution of the DESI-Y1 Lya Sample.

### DIFF
--- a/py/desisim/scripts/gen_qso_catalog.py
+++ b/py/desisim/scripts/gen_qso_catalog.py
@@ -26,6 +26,15 @@ def main():
     parser.add_argument("--invert", action='store_true', default=False,
                         help='Invert the selection of the mock catalog')
 
+    parser.add_argument("--exptime", type=float, default=None, required=False,
+                        help='Exposure time to assign to all targets in the mock catalog')
+                    
+    parser.add_argument("--release", type=str, default='iron', choices=['iron','Y5'], required=False,
+                        help='DESI survey release to reproduce')
+    
+    parser.add_argument("--overwrite", action='store_true', default=False,
+                        help='Overwrite output file if it exists')
+    
     args = parser.parse_args()
 
     survey = SurveyRelease(mastercatalog=args.input_master,seed=args.seed, qso_only=True, data_file=args.input_data,invert=args.invert)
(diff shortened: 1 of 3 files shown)


p-slash:
Looks good to me. Thanks!

p-slash (approved):

p-slash:
I generated P1D mocks using this branch. Here are my observations:
- The code correctly generates quasars at the given redshifts and FLUX_R values.
- However, for quasars z>3.7, the IGM absorption reduces the FLUX_R value. This recalculated FLUX_R is less than the input FLUX_R.
 
image
- This means the input FLUX_R values are amplified accordingly. Alternatively, we can address this within quickquasars by amplifying FLUX_R values.

p-slash:
I tested amplifying the FLUX_R values using the mean flux of the Ohio mocks. This is what I get after running quickquasars:
image
Not perfect, but better. One caveat is that Ohio mocks are not tuned for high redshift mean flux, so they have less than the literature values.

I also compared IVAR values in B arm for some random spectra. The agreement between observed spectra and mock spectra okay.
image

p-slash:
@HiramHerrera Please go ahead and merge it. Good work!

PR #582, opened 2025-02-01
Remove DesiTest

weaverba137:
This PR removes `desiutil.setup.DesiTest` from setup.py.

It looks like the format of `desimodel/data/focalplane/platescale.txt` has changed since the last time desisim was successfully tested, and that is causing the majority of the test errors.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,13 +1,19 @@
-# numpy<1.20
 pytz
 requests
+# Ensure scipy dependency does not bring in a recent version of numpy.
 scipy<1.9
+# The version of astropy will be reset by GitHub Actions tests.
 astropy
+# Ensure compatibility with numpy <1.23.
 numba<0.60
+# Ensure compatibility with numpy <1.23.
 matplotlib<3.9
 healpy
+# It is becoming difficult to install speclite and specsim directly from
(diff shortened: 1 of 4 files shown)


weaverba137:
> It looks like the format of `desimodel/data/focalplane/platescale.txt` has changed since the last time desisim was successfully tested, and that is causing the majority of the test errors.

Actually, `platescale.txt` hasn't changed since 2020. It's more likely that the reader code in desimodel might not be compatible with recent numpy versions.

requirements.txt
```diff
@@ -1,10 +1,13 @@
 # numpy<1.20
 pytz
 requests
-astropy==5.0
-numba
+scipy<1.9
+astropy
+numba<0.60
+matplotlib<3.9
```
sbailey, line 7: If you still have the notes from the failed runs, please document here in comments why each of these version constraints is needed.  e.g. is it directly due to specsim itself, or incompatibilities in the available 3rd party libraries that aren't fully specified in their own constraints?

weaverba137: I added some notes in the latest commit, please take a look.

sbailey (approved):

Issue #583, opened 2025-07-18
pyproject.toml file needed


weaverba137:
In order for desisim to remain pip-installable, at least a minimal `pyproject.toml` file is needed, although in reality, much more significant package infrastructure updates are needed.

weaverba137:
Note: the requirements.txt file specifies `pytz`, but that is not explicitly imported by the package. `specter` is only ever used to obtain `specter.__version__`.

PR #589, opened 2025-09-08
fix tests and support numpy 2.x

### LINKED ISSUE #584: NumPy 2.0 compatibility
Update the test matrix and make other package improvements to ensure NumPy 2.0 compatibility.

### LINKED ISSUE #585: Tests failing due to missing 'gray' program
At least two desisim test failures are due to tests attempting to access a 'gray' program, which may have existed during the design phases of DESI but does not exist now.

### LINKED ISSUE #586: Tests failing due to missing OBSCONDITIONS
At least two desisim tests are failing because an input tile file is missing the column `OBSCONDITIONS`.

sbailey:
This PR
  * fixes code+unittests under the current desiconda environment at NERSC (fixes #585, fixes #586)
  * Adds support for numpy/2.x and scipy/1.16.x (both at NERSC and on GitHub actions) (fixes #584)
  * Removes "from __future__ import ..." leftover from python 2.x days (changing many files)
  * Converts use of deprecated `pkg_resources.resource_filename` to `importlib.resources.files`

As you probably noticed, it took me many iterations to get the GitHub Actions workflow passing even after the tests were completely passing in an environment at NERSC.  The result is a somewhat fragile state and will need more attention, e.g. in how dependencies are specified in requirements.txt vs. `.github/workflows/python-package.yml`. I suggest that we save that for a separate PR that is adding pyproject.toml / setup.cfg (see #583).  Additionally, some of the fragility is due to specsim still using astropy_helpers which is incompatible with the latest setuptools.  Addressing desihub/specsim#139 will help with that.

For the record: this is not yet compatible with astropy 7.x, though in the chaos of getting tests to pass I don't recall if that is due to desisim itself or one of its dependencies.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -76,12 +76,18 @@ jobs:
               run: |
                 mkdir -p ${GITHUB_WORKSPACE}/desimodel
                 svn export https://desi.lbl.gov/svn/code/desimodel/${DESIMODEL_DATA}/data ${GITHUB_WORKSPACE}/desimodel/data
+            - name: Install surveyops snapshot
+              env:
+                SURVEYOPS_VERSION: '2.0'
+              run: |
+                wget -nv https://data.desi.lbl.gov/public/epo/example_files/surveyops_${SURVEYOPS_VERSION}_ops.tar.gz
+                tar xzf surveyops_${SURVEYOPS_VERSION}_ops.tar.gz
             - name: Install desisim Data
               run: |
                 wget https://github.com/desihub/desisim-testdata/archive/${DESISIM_TESTDATA_VERSION}.zip
                 unzip ${DESISIM_TESTDATA_VERSION}.zip
             - name: Run the test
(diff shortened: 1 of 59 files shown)


.github/workflows/python-package.yml
```diff
@@ -37,33 +56,42 @@ jobs:
               uses: actions/checkout@v4
               with:
```
weaverba137, line 57: Just a suggestion:
```
with:
    fetch-depth: 0
```
It turns out this doesn't do what you might think. `fetch-depth = 0` means "fetch all commits". When in fact we want "fetch the latest commit", which is the *default* behavior of `actions/checkout`. So the `with:` clause in that step isn't needed at all. If you want to remove these in this PR, go ahead, but it's also something that could be done later.

doc/rtd-requirements.txt
```diff
@@ -1,4 +1,3 @@
 setuptools>60
```
weaverba137, line 1: I suggest simplifying this further and just have one line:
```
sphinx-rtd-theme>1
```

sbailey:
Thanks. Both of those suggestions refer to leftover pieces and/or copy-and-pasted from elsewhere, so I'm happy to further simplify by removing the unneeded lines.

The coverage tests are "failing" presumably due to low coverage, so I'm going to ignore those for now.

PR #598, opened 2026-05-15
parser argument to use desi-2.yaml in specsim

p-slash:
What the title says. A simple modification


sbailey:
Noticing this old PR while making new tags for Nevis/DR3.  @p-slash @HiramHerrera where is the desi-2.yaml file that this references? Is that a file that would be in desimodel but isn't committed, or elsewhere?

PR #603, opened 2026-09-18
add pyproject.toml and support desiutil 4.x

### LINKED ISSUE #583: pyproject.toml file needed

sbailey:
This PR fixes #583 by adding a pyproject.toml and doing various setup.py cleanup to avoid setup.py having `from desiutil.setup import DesiVersion, get_version` which no longer exist in desiutil 4.x.

I'm not entirely sure I like the split between how dependencies are tracked in 3 places:
* setup.cfg (pip)
* requirements.txt (things from github that can't be specified in setup.cfg), and
* .github/workflows/python-package.yml (packages that we specifically want to specify versions for testing).

there might be a cleaner way to do this, but I'm pausing for now for review so this doesn't take more time for a package that isn't strictly needed for Nevis.  But something along these lines is needed for the 26.9 release because currently tags of desisim can't be installed by desiutil 4.x due to the removed DesiVersion, get_version.

Co-authored with Claude


.github/workflows/python-package.yml
```diff
@@ -61,9 +61,14 @@ jobs:
               # setuptools version restriction due to specsim ah_helpers needing setuptools.package_index
```
weaverba137, line 61: I believe that the specsim version in PyPI no longer needs `ah_helpers`, and therefore the restriction on setuptools isn't necessary.

pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[build-system]
+requires = ["setuptools",
+            "wheel"]
```
weaverba137, line 3: Optional: `wheel` isn't necessary here, even though it appears in other files.

requirements.txt
```diff
@@ -1,30 +1,47 @@
...
-matplotlib
-healpy
-# It is becoming difficult to install speclite and specsim directly from
-# git+https, so installing from PyPI is definitely preferred here.
-# Update 2025-09-08: get specsim/main from github pending new tag
-speclite
-fitsio
-# Install desiutil separately since it is needed for the other installs.
-# git+https://github.com/desihub/desiutil.git@3.1.0#egg=desiutil
+# desisim's abstract dependencies are declared in setup.cfg (install_requires),
```
weaverba137, line 1: Do we need to keep the requirements.txt file at all? If it is only used for testing, the content could be moved into the `python-package.yml` file.

weaverba137 (changes requested):
I think this looks fine overall; just a few suggestions.

sbailey:
@weaverba137 thanks for the comment, which I believe I have addressed:
* removed version restriction on setuptools
* removed wheel from pyproject.toml requires 
* removed misleading/confusing requirements.txt and moved logic into python-package.yml

I also updated python-package.yml to unify the coverage vs. non-coverage paths, similar to desiutil.  Tests pass though coveralls is complaining perhaps due to a tiny change in coverage even though no code files changed. I'd prefer to ignore that "failure" here. Does the rest look ok?

doc/changes.rst
```diff
@@ -5,7 +5,19 @@ desisim change log
 0.39.2 (unreleased)
 -------------------
 
-* No chnages yet.
+* Move packaging metadata into ``pyproject.toml`` and ``setup.cfg``, following
+  the pattern used by desispec.
+* Trim ``requirements.txt`` to only the dependencies that ``install_requires``
```
weaverba137: This file is removed by this PR, so please change this item.

weaverba137 (approved):
Looks good, but one comment on the change log.

weaverba137:
PS, it is safe to ignore the coverage change warning.

